PR #4, opened 2026-06-05
fix(windows): hook server named pipe + mempalace CLI detection

Xileck:
## What

Makes two Windows-only code paths work, both strictly gated on
`process.platform === 'win32'` so macOS/Linux behavior is unchanged.

### 1. Hook server: named pipe on Windows
`HiveManager.sockPath()` returned a Unix-domain socket *file path*
(`<hiveRoot>/hooks.sock`) that was passed straight to `server.listen()`.
On Windows, Node's `net` IPC is implemented with **named pipes** (a flat
`\.\pipe\` namespace, not the filesystem), so a raw file path fails to
bind:

```
[hive] hook server error: Error: listen EACCES: permission denied
  D:\...\harness\hive\hooks.sock
```

The whole autonomy/hook plane (avatar activity, Stop-loop inbox drain)
never came up. Fix: on Windows derive a stable, per-root pipe name
(`\.\pipe\munder-difflin-<sha1(root)>`). Both the server (`listen`) and
the `cth-hook` shim (`createConnection`) read the same `sockPath()`, so
they stay in sync. POSIX still returns `<root>/hooks.sock`.

### 2. mempalace detection on Windows
`MemoryManager.bin()` only resolved the CLI via a POSIX login shell
(`$SHELL -ilc 'which mempalace'`) and Unix fallback paths with no `.exe`
suffix. On Windows neither matched, so semantic memory was reported as
unavailable and **HIVE MEMORY always showed "Not set up"** even after
`uv tool install mempalace`. Fix: on Windows use `where` and probe
`%USERPROFILE%\.local\bin\mempalace.exe` (the uv tool install target).

## macOS / Linux impact
None. Both changes branch on `win32`; the POSIX path is identical to
before (verified by inspection + `npm run typecheck`).

## Testing
- `npm run typecheck` passes.
- Verified on Windows 11 (Node 24, Electron 32): hook server now binds a
  named pipe, agents report activity, and `mempalace` is detected so the
  HIVE MEMORY panel activates and the shared palace is populated.

### DIFF
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -23,7 +23,7 @@ import {
 } from 'node:fs';
 import { join } from 'node:path';
 import { spawnSync } from 'node:child_process';
-import { randomBytes } from 'node:crypto';
+import { randomBytes, createHash } from 'node:crypto';
 
 // ─── Types ──────────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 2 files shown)


chaitanyagiri:
Hey you are the first person to raise a PR here, first of all I am so grateful that you did this.
Secondly thanks for the windows patch, I'll need to see if this works on a windows machine or not. Could you add some screenshots aswell I just want to ensure the build is stable as I do not have a windows machine right now.

chaitanyagiri:
I was able to verify this, your commits with author credits are merged and released in v0.1.9. Closing this PR now.

PR #5, opened 2026-06-05
fix(memory): serialize palace mining to avoid writer-lock collisions

Xileck:
## What

Serializes semantic-memory mining so concurrent `mempalace mine` children
don't collide on the palace's single-writer lock.

## Why

The palace (ChromaDB) allows **one writer at a time**, but `mineNow()`
iterated over every changed agent and `spawn`ed a `mempalace mine` child
for each **concurrently**. With more than one agent (e.g. the GOD agent
plus any worker) all but one mine fails:

```
[memory] mine god exited 1: mempalace: palace <path> is held by PID <n>
  (... mine .../agents/assistant); wait for it to finish or stop the
  holder before retrying
```

It self-heals on later ticks but spams errors and delays mining. Notably
the method's own doc comment already said *"one at a time"* — the
implementation just never honored it.

## How

- `mineNow()` is now `async` and `await`s each agent's mine in turn, so
  only one writer touches the palace at a time.
- A `mining` guard makes a slow pass skip the next interval tick instead
  of overlapping it.
- `mineAgent()` returns a `Promise<void>` that resolves on the child's
  `close`/`error`.

All three callers (`startMineLoop`, the `setInterval` tick, and the
`hive:mineNow` IPC handler) are fire-and-forget, so the `void` ->
`Promise<void>` change is transparent.

## Platform impact

Platform-independent fix — the single-writer constraint holds on macOS,
Linux, and Windows alike; the collision is simply easier to trigger once
a second agent exists. No platform-specific branching.

## Testing

- `npm run typecheck` passes.
- On a fresh palace with two agents (GOD + assistant): before, one mine
  failed every pass with the lock error; after, both wings mine cleanly
  with zero collisions and the palace populates.

### DIFF
--- a/src/main/memory.ts
+++ b/src/main/memory.ts
@@ -40,6 +40,8 @@ export class MemoryManager {
   private binCache: string | null | undefined;
   private mineTimer: NodeJS.Timeout | null = null;
   private initStarted = false;
+  /** True while a mineNow() pass is in flight — serializes palace writers. */
+  private mining = false;
   /** agentId → memory.md mtimeMs at last successful mine (skip unchanged). */
   private lastMined = new Map<string, number>();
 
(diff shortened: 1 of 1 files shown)


chaitanyagiri:
I was able to verify this, your commits with author credits are merged and released in v0.1.9. Closing this PR now.

Issue #2, opened 2026-06-05
Michael cancels his own "bypass permissions" when launching and exits


albozes:
### What happened?

When I launch Munder Difflin the very first time, the terminal starts up, asks to Bypass Permissions, and then it automatically selects "No, exit". The process is then exited.

Absolutely love this concept and would love to get it running!

### Steps to reproduce

Launch Munder Difflin for the first time.
Set up the "home" of MD in a normal, local folder.
Open a repo.
Other settings completely default.
Let Michael walk in.

### Logs / screenshots

<img width="516" height="387" alt="Image" src="https://github.com/user-attachments/assets/f3a49c50-e542-4830-8243-c21482a6b7ed" />

### macOS version

26.6 Beta

### Node version

current release (downloaded from the website)

### Pre-flight

- [x] I re-ran `npm install` so `node-pty` is rebuilt for the current Electron ABI.
- [x] I searched existing issues and this isn't a duplicate.

Issue #1, opened 2026-06-04
Hive router never starts on Windows : hooks.sock (Unix domain socket) is never created, so no messages are ever routed


JLAD75:
### What happened?

Environment

OS: Windows 11
App: Munder Difflin v0.1 (packaged .exe from Releases, auto-mode on)
Claude Code: on PATH, working (claude --version OK); agents spawned with an absolute path command (C:\Users\<user>\.local\bin\claude.exe --permission-mode bypassPermissions)
Repo registered as a room, agents spawned successfully (GOD/Michael + 5 worker agents)

Summary
On Windows, the hive's inter-agent messaging never works. Agents spawn and run fine as individual Claude Code sessions, but no message is ever delivered from one agent's outbox to another's inbox. After investigation, the root cause appears to be that the orchestrator/router process never creates its Unix domain socket hive/hooks.sock, so the routing loop never runs.

Suggested fix / direction

On Windows, use a named pipe (\\.\pipe\...) instead of an AF_UNIX filesystem socket for the hook transport, with a platform switch in hooks.ts (and matching HIVE_SOCK resolution in cth-hook.cjs).
Independently: surface main-process socket/router startup errors somewhere user-visible (a startup banner or a log file), since the current silent-failure path makes this look like a working system when it isn't.

Additional notes (possibly worth separate issues)

No agent persistence: a full app restart wiped all spawned agents (GOD included), requiring full re-creation. This matches the unchecked "Persistence (SQLite)" roadmap item, but combined with the bug above it makes iterating on Windows costly.
Default command vs PATH: spawning failed with "File not found" until I replaced claude with the absolute path to claude.exe in the command string , the packaged Electron app doesn't inherit the shell PATH on Windows. Might be worth documenting, or resolving claude against common install locations.

Happy to provide the full hive/ directory listing, log.jsonl, or registry.json if useful. Thanks for building this in the open , the rest of the experience (floor, terminals, GOD planning, asset workflow) is genuinely great.

### Steps to reproduce

On Windows, register a repo and spawn a GOD agent + at least one worker agent.
Ask the GOD agent to message a worker via its outbox (e.g. a ping).
Observe the worker's terminal and the on-disk hive state.

Expected behavior
The message moves from hive/agents/<sender>/outbox/ to hive/agents/<recipient>/inbox/, a delivery event is appended to hive/log.jsonl, and the recipient session is woken to drain its inbox.
Actual behavior
[...]

PR #7, opened 2026-06-06
fix: keep background agents' statuses live (fixes #3)

### LINKED ISSUE #3: Agents show as idle until you click their profile
### What happened?

The agents will be working but their profile at the bottom show as idle, once you click them it shows 'working' as inteneded.

### Steps to reproduce

Give the agents a task
Observe they show as idle
Click their profile
Observe it switches to working

### Logs / screenshots

```shell

```

### macOS version

Windows

### Node version

26.1.0

_No response_

### Pre-flight

Gulum:
## What

Fixes #3 — agents showing a stale status (idle while working, or stuck on "working" after finishing) until you click their profile.

## Why it happened

The pty-stream → avatar-status parser lived in `usePtyParser`, which only ran while an agent's **profile panel was mounted**. Any agent you were *not* looking at kept whatever status it last showed, indefinitely.

This is worse than cosmetic: the idle-gated inbox nudge (`useHive` effect #3) only nudges agents whose status is `idle`/`waiting`. An agent frozen on `working` was **skipped forever**, so its hive mail sat unread and collaboration quietly stalled — especially on Windows, where the hook transport is currently down entirely (#1, PR #4), leaving the pty parser as the only status source.

## Change

- New `ptyStatusWatcher.ts`: the same parsing logic, decoupled from React, subscribed to **every live agent pty exactly once, app-wide**. Subscriptions are reconciled against the agent list on store changes (cheap Map diff, idempotent).
- `useHive`: new effect 2b keeps the watcher in sync.
- `usePtyParser`: now a no-op shim so existing views keep compiling (no double-parsing, no duplicate feed entries).

Hook events remain the authoritative status source (`useHive` effect #2 is untouched); the watcher is the refinement + fallback layer.

## Testing

- `npm run typecheck` passes (node + web).
- Verified on Windows 11 (Electron 32): background agents now drift to idle ~4 s after their terminal goes quiet and flip to working on output, without ever opening their profile; the inbox nudge picks them up again.

Plays well with #4 — together they make agent status fully live on Windows.


chaitanyagiri:
Hey thanks for this, I'll review and merge this in the new release v0.2.0

Gulum:
Closing as superseded by the v0.2.0 architecture. This PR fixed frozen background-agent statuses by making the *pty-output parser* global (one watcher per live pty instead of mounted-panel-only). Since 5A/5B/5C, avatar state is driven by **hook events**, which fire for every agent regardless of UI focus — so the original bug (#3) is structurally fixed at the source, with better signals than TUI scraping (real tool names, compacting/looping states, breaker precedence). Keeping a second pty-scraping status writer around would only race the hook-driven one (e.g. its 4s idle heuristic could overwrite a hook-confirmed `working`).

If you ever want a fallback for sessions whose hooks fail to register, the watcher in this branch is still a reasonable starting point — but as of v0.2.1 it would do more harm than good.

PR #8, opened 2026-06-06
fix: skip redundant pty resizes - boot banner no longer stacks in scrollback

Gulum:
Rebased onto v0.2.1. Since you already merged the scroll fix itself (`fe001e3` + `aa42be8` — thanks!), this PR now contains **only the follow-up you don't have yet**:

**`fix: skip redundant pty resizes — boot banner no longer stacks in scrollback`**

`tryFit()` previously called `resizePty` on every fit, even when cols/rows hadn't changed. ConPTY treats *any* resize as "redraw everything", so each agent switch re-emitted the boot banner into scrollback — switch back and forth a few times and the history was full of stacked banners. Now we only forward the resize when the dimensions actually changed.

Diff is down to one small hunk in `PtyTerminalView.tsx`.

### DIFF
--- a/src/renderer/src/components/PtyTerminalView.tsx
+++ b/src/renderer/src/components/PtyTerminalView.tsx
@@ -158,8 +158,16 @@ export function PtyTerminalView({ ptyId, onStreamData, onUserPrompt, onToggleFul
     // up to read history back down to the bottom.
     const tryFit = (scrollToEnd = false) => {
       try {
+        const before = { cols: entry.term.cols, rows: entry.term.rows };
         entry.fit.fit();
-        window.cth.resizePty(ptyId, entry.term.cols, entry.term.rows);
+        // Only poke the pty when the grid ACTUALLY changed: every resize makes
+        // the Claude TUI repaint its whole screen, and each repaint pushes the
+        // previous frame into scrollback — the attach-time refit cascade (rAF,
+        // 60ms, 240ms, font-load) used to stack the boot banner three times
+        // before the user ever typed anything.
+        if (entry.term.cols !== before.cols || entry.term.rows !== before.rows) {
+          window.cth.resizePty(ptyId, entry.term.cols, entry.term.rows);
+        }
         entry.term.refresh(0, Math.max(0, entry.term.rows - 1));
       } catch { /* host may not be sized yet */ }
       if (scrollToEnd) {


chaitanyagiri:
Hey thanks for this, I'll review and merge this in the new release v0.2.0

Gulum:
Follow-up commit (88b7e06), same viewport/scrollback theme: the attach-time refit cascade (rAF, 60ms, 240ms, font-load) sent a pty resize on EVERY fit even when the grid had not changed. Each resize makes the Claude TUI repaint its whole screen, and every repaint pushes the previous frame into scrollback - a freshly spawned agent showed the welcome banner stacked three times in phantom scrollback before the user ever typed. Now the pty is only resized when cols/rows actually changed.

chaitanyagiri:
## 🤖 Automated PR Review

**Summary** — In `tryFit()`, the attach-time refit cascade (rAF, 60ms, 240ms, font-load) used to call `resizePty` on *every* fit even when the grid was unchanged. ConPTY treats any resize as a full redraw, so each no-op fit re-emitted the boot banner into scrollback. This PR captures cols/rows before `fit()` and only forwards the resize when they actually changed.

### ✅ What's good
- Correct mechanism: `entry.fit.fit()` mutates `entry.term.cols/rows` in place, so snapshotting `before` *before* the `fit()` call (PtyTerminalView.tsx:161) and comparing after is exactly the right way to detect a real grid change. No off-by-one or stale-read trap here.
- The `before`/after diff is a pure guard — it changes nothing about the `fit()` or the `refresh(0, …)` that follows, so the cascade still re-fits on every tick; we just stop spamming the pty. Behavior on a genuine resize (window/RO/font-load growing the grid) is identical to before.
- The inline comment is genuinely explanatory and ties the change back to the observed symptom (banner stacked 3× in phantom scrollback). Good for the next reader.
- Tight scope: +9/-1, one hunk, no churn to the surrounding scroll-restore logic that this file is sensitive about.

### 🔧 What could be better
Nothing blocking — minor/optional:
- Consistency nit: the sibling `resizePty` in the font-size effect (PtyTerminalView.tsx:224) is still an unconditional resize after `fit()`. It's much lower-impact (a font-size change almost always changes the grid, so the redraw is usually warranted), so it's reasonable to leave it. But if you ever want the guard to be a single invariant ("only poke the pty when the grid moved"), that's the one other call site that bypasses it.

### 💬 Feedback / questions
- None — the description, the diff, and the @Gulum follow-up comment all line up. The first-spawn case is handled correctly: pool-default dims ≠ fitted dims on initial attach, so the resize still fires and the banner renders at the right size.

### 🧪 Local test report
Built this PR in an isolated worktree with its own dependencies, then smoke-launched the app:
- `npm run build` (electron-vite) → ✅ PASS (built in ~3.65s)
- `npm run preview` (launch app) → ✅ App launched and stayed up ~10s (timeout-killed). The GPU-process exit, "Network service crashed", and `EADDRINUSE :::3847` lines in the run log are headless-sandbox / port-already-held noise — they reproduce on `main` and are unrelated to this change.
- **Verdict:** Looks good to merge.

PR #12, opened 2026-06-06
feat: statusLine-driven context gauge - exact tokens + real window size

Gulum:
Rebased onto v0.2.1. The commit you merged (`9f98dc9`) was the **v1 transcript-poll gauge** — it works, but it has two known weaknesses that this PR (now much smaller) fixes:

1. **Stale gauge after restart** — restored agents showed the *previous* session's fill until the first poll corrected it.
2. **Wrong window size** — the transcript never reveals whether a session has a 200k or a 1M context window, so v1 guessed from the model string and could peg the gauge red at 20% on 1M sessions.

The replacement: a **`statusLine` command** in each agent's per-session settings file. Claude Code invokes it after every response and pipes it a JSON payload whose `context_window.total_input_tokens` / `context_window_size` carry the session's *exact* accounting — the only clean source for the real window size. The shim prints a tiny `ctx 235k/1000k (24%)` footer into the TUI and forwards the payload over the existing hook pipe as a synthetic `Status` event; the renderer updates the gauge push-based, no polling.

The transcript poll is kept only as a backfill for agents whose status line hasn't fired yet (e.g. freshly restored, no response so far).

Conflict-free against main as of `a936929` (v0.2.1); the `Status` event is handled before the HALT/breaker gates so telemetry can never trip them.

### DIFF
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -373,6 +373,12 @@ export class HiveManager {
       hooks: [{ type: 'command', command: cmd }]
     });
     return {
+      // The status line gets the session status JSON after every response —
+      // including context_window.{total_input_tokens,context_window_size},
+      // the only clean programmatic source for the session's REAL context
+      // window. The shim prints a compact in-terminal gauge and forwards the
+      // payload to the harness (agent-card context gauge, exact limit).
+      statusLine: { type: 'command', command: `${cmd} --status`, padding: 0 },
       hooks: {
         Stop: [entry()],
         SubagentStop: [entry()],
@@ -858,6 +864,7 @@ write there become searchable by every agent. You don't run \`mine\` yourself.
 const HOOK_SHIM = `#!/usr/bin/env node
 'use strict';
 const net = require('net');
+const isStatus = process.argv.includes('--status');
 let data = '';
 process.stdin.setEncoding('utf8');
 process.stdin.on('data', (d) => { data += d; });
@@ -866,6 +873,31 @@ process.stdin.on('end', () => {
   try { payload = JSON.parse(data || '{}'); } catch (_) {}
   if (!payload.agent_id) payload.agent_id = process.env.AGENT_ID || null;
   const sock = process.env.HIVE_SOCK;
+  if (isStatus) {
+    // Status-line mode: Claude Code pipes the session status JSON (incl.
+    // context_window.total_input_tokens / .context_window_size) after every
+    // response. Print the in-terminal gauge IMMEDIATELY (the TUI is waiting),
+    // then forward the payload to the harness fire-and-forget so the agent
+    // card's context gauge updates push-based, with the EXACT window size.
+    payload.hook_event_name = 'Status';
+    const cw = payload.context_window || {};
+    const used = cw.total_input_tokens, size = cw.context_window_size;
+    if (typeof used === 'number' && typeof size === 'number' && size > 0) {
+      const pct = Math.round((used / size) * 100);
+      process.stdout.write('ctx ' + Math.round(used / 1000) + 'k/' + Math.round(size / 1000) + 'k (' + pct + '%)');
+    }
+    if (sock) {
+      try {
+        const c = net.createConnection(sock, () => { c.end(JSON.stringify(payload) + '\\n'); });
+        c.on('error', () => {});
+        c.on('close', () => process.exit(0));
+      } catch (_) { process.exit(0); }
+    } else {
+      process.exit(0);
+    }
+    setTimeout(() => process.exit(0), 1500).unref();
+    return;
+  }
   if (!sock) { process.exit(0); }
   let resp = '';
   const done = (code) => { if (resp) process.stdout.write(resp); process.exit(code); };
--- a/src/main/hooks.ts
+++ b/src/main/hooks.ts
@@ -24,6 +24,8 @@ interface HookPayload {
   agent_id?: string | null;
   session_id?: string;
   transcript_path?: string;
+  /** Status-line payloads only: the session's live context accounting. */
+  context_window?: { total_input_tokens?: number; context_window_size?: number };
   cwd?: string;
   tool_name?: string;
   tool_input?: unknown;
@@ -95,6 +97,25 @@ export class HookServer {
     const event = p.hook_event_name ?? 'Unknown';
     if (agentId && typeof p.transcript_path === 'string' && p.transcript_path) {
       this.transcriptPaths.set(agentId, p.transcript_path);
+    }
+
+    // Status-line payloads carry the session's EXACT context accounting —
+    // current tokens AND the real window size (200k vs 1M, which nothing else
+    // exposes). Forward to the renderer for the agent-card context gauge.
+    // Handled FIRST and returned early: this is pure telemetry from the
+    // statusLine shim, not a real hook boundary — it must never trip the
+    // HALT gate or feed the breaker's loop detector below.
+    if (event === 'Status') {
+      const cw = p.context_window;
+      if (agentId && cw && typeof cw.total_input_tokens === 'number'
+        && typeof cw.context_window_size === 'number' && cw.context_window_size > 0) {
+        this.getWebContents()?.send('hive:contextUpdate', {
+          agentId,
+          tokens: cw.total_input_tokens,
+          limit: cw.context_window_size
+        });
+      }
+      return {};
     }
 
     // 7C.3 — a graceful operator HALT overrides everything (incl. the inbox
--- a/src/main/hooks.ts
+++ b/src/main/hooks.ts
@@ -104,7 +104,11 @@ export class HookServer {
     // exposes). Forward to the renderer for the agent-card context gauge.
     // Handled FIRST and returned early: this is pure telemetry from the
     // statusLine shim, not a real hook boundary — it must never trip the
-    // HALT gate or feed the breaker's loop detector below.
+    // HALT gate or feed the breaker's loop detector below. The early return
+    // also (deliberately) skips recordSession for status ticks: a statusLine
+    // payload's session_id adds nothing the real hooks don't already record,
+    // and telemetry should never write to the registry. transcript_path IS
+    // still captured above, where every payload shape benefits from it.
     if (event === 'Status') {
       const cw = p.context_window;
       if (agentId && cw && typeof cw.total_input_tokens === 'number'
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -1024,12 +1024,15 @@ ipcMain.handle('app:resetAll', () => {
 ipcMain.handle('hive:agentUsage', (_evt, cwd: unknown) =>
   typeof cwd === 'string' ? readAgentUsage(cwd) : null);
 // Current context size (tokens) of an agent's LIVE session — the transcript
-// path is learned from the agent's hook payloads, so this works even when
-// several agents share one cwd. Null until the first hook fires.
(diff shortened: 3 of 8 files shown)


Gulum:
Follow-up commit (3dfb101): after an app restart the gauge showed the DEAD session's state. Two gaps - contextTokens/contextLimit/progress were persisted with the agent and re-shown 1:1, and a fresh session with no assistant message yet read as null, so the poll skipped the agent and the stale value never cleared. The gauge fields are volatile now, and hive:agentContext returns 0 (not null) when the transcript path is known but holds no assistant usage yet - SessionStart fires at spawn, so a freshly started session zeroes its gauge within one poll.

Gulum:
Second follow-up (9dec5ab): the limit heuristic only looked for [1m] in the agent's model id, so agents on the CLI-default model that actually run 1M sessions were treated as 200k - the god agent's gauge pegged 8/8 red at 235k/1m (24% real). Neither the model id nor the transcript can disambiguate (the [1m] alias exists only inside Claude Code). Now three self-correcting layers, ratchet-up only: model-id hint; a one-shot /context probe per session once the agent is idle, with the pty parser passively sniffing the limit from the output denominator; and observed usage above 200k proving a 1M window.

Gulum:
Heads-up: commit 1857d97 REPLACES the earlier slash-command probe with a cleaner mechanism - Claude Code's statusLine. The per-agent settings now configure statusLine on the existing cth-hook shim (--status mode); Claude Code pipes the session status JSON after every response, including context_window.total_input_tokens and .context_window_size - the only reliable programmatic source for the real window size (200k vs 1M; the 1m alias never appears in transcripts). The shim prints a compact in-terminal gauge and forwards the payload to the HookServer, which emits hive:contextUpdate for a push-based, exact card gauge. The transcript poll stays as backfill until the first status tick. Merge-order note: the push path rides the hooks socket, so on WINDOWS it needs the named-pipe fix from #4 to be live; without #4 the gauge gracefully falls back to the transcript poll with heuristic limits (the pre-existing behavior). On macOS/Linux it works standalone.

Gulum:
@chaitanyagiri please merge the open changes, cause it make the gauge work on actual stats.

chaitanyagiri:
## 🤖 Automated PR Review

**Summary** — Replaces the transcript-poll context gauge with a `statusLine` command on the existing hook shim: Claude Code pipes the session status JSON after every response, the shim prints an in-terminal `ctx 235k/1000k (24%)` footer and forwards `context_window.{total_input_tokens,context_window_size}` over the hook pipe as a synthetic `Status` event for a push-based, exact card gauge. The transcript poll stays as backfill until the first status tick. Clean, well-layered design; the iterative follow-ups (volatile fields, fresh-session zeroing, real-limit detection) landed the rough edges already.

### ✅ What's good
- Reusing the existing `cth-hook` shim in a `--status` mode (`hive.ts` HOOK_SHIM, `isStatus` branch) instead of adding a second binary — and configuring it as `statusLine` in the same per-session `settings.json` that already carries the hooks (`hive.ts:378`). No new plumbing, no repo edits, merges cleanly with the hooks block.
- The `Status` event is handled FIRST and `return {}`s early in `hooks.ts` (~:104), so pure telemetry can never trip the HALT gate, the breaker's loop detector, or the inbox-drain. Exactly the right place for it.
- Defensive at every JSON boundary the payload crosses: the shim guards `typeof used/size === 'number' && size > 0` before printing, and `hooks.ts` re-checks `total_input_tokens`/`context_window_size` types and `> 0` before emitting `hive:contextUpdate`. Malformed/missing `context_window` simply produces no gauge update rather than NaN.
- The backfill handoff is correct: effect 2c now `continue`s on `a.contextLimit !== undefined` (`useHive.ts`), and `contextLimit` is stripped from persistence (`store.ts` PersistedAgent), so after a restart it's `undefined` and the poll resumes until the next status tick re-pins the exact limit. `progress: 0` reset on load avoids showing a dead session's bar.
- `hive:agentContext` returning `0` (not `null`) when the transcript path is known-but-empty (`index.ts`) is the right call — a freshly restarted session zeroes its gauge within one poll instead of leaving a stale value up.
- No injection surface: the statusLine command is `node "${shim}" --status` with a harness-owned, quoted path; no user/agent-controlled string is interpolated.

### 🔧 What could be better
- **Renderer effect 2d has no `limit > 0` guard** (`useHive.ts`, the `onHiveContextUpdate` handler): `Math.round((tokens / limit) * 8)` with `limit === 0` and `tokens === 0` yields `NaN`, and `Math.min(8, NaN)` stays `NaN` → a `NaN` progress into the store. Today this is unreachable because `hooks.ts` filters `context_window_size > 0` upstream, but the renderer trusts its IPC input blindly. A one-line `if (!limit) return;` (or `limit > 0 ? ... : 0`) makes it defense-in-depth and matches the guarding style used everywhere else in this PR.
- **`Status` payload still hits `recordSession` / `transcriptPaths.set` before the early return?** No — the early return is above those, good. But the synthetic `Status` event sets `payload.hook_event_name = 'Status'` in the shim while Claude Code's real statusLine JSON may also carry `session_id`/`transcript_path`. Those are now silently dropped for status ticks (you `return {}` before `recordSession`). That's fine for telemetry, just worth a one-line comment so a future reader doesn't "fix" it by moving the block down.
- **In-terminal footer width**: `ctx 235k/1000k (24%)` is printed with `padding: 0`; for a 1M-window agent that's ~18 chars every response. Minor, but consider whether the statusLine competes with Claude Code's own footer on narrow terminals. Not blocking.
- **The pty-parser `/context` sniffer (`usePtyParser.ts`, CONTEXT_LIMIT_RE) is now largely redundant** with the statusLine on macOS/Linux. It's harmless ratchet-up-only backfill, but it's a third limit-detection path — worth a comment noting it's the fallback-of-the-fallback (for the pre-#4 Windows named-pipe gap you flagged) so it isn't mistaken for dead code later.

### 💬 Feedback / questions
- The shim's `--status` path uses `c.end(JSON.stringify(payload) + '\n')` (write+half-close) plus a 1500ms `unref` timeout, vs. the hook path's 5000ms. Is 1500ms comfortably above the worst-case UDS connect+write so a slow socket never drops a tick, or is the shorter timeout deliberately to keep the statusLine snappy (since the gauge print already happened)? Reads fine either way — just confirming the intent.
- Re @Gulum's merge-order note: agreed the push path rides the hooks socket and needs #4's named-pipe fix for Windows, with graceful fallback to the transcript poll otherwise. Worth stating that ordering in the PR description / a merge checklist so it's not lost.

### 🧪 Local test report
Built this PR in an isolated worktree with its own dependencies, then smoke-launched the app:
- `npm run build` (electron-vite) → ✅ PASS (`✓ built in 3.50s`, all chunks emitted)
- `npm run preview` (launch app) → ✅ App launched and stayed up ~10s (timeout-killed); `[telemetry] collector listening` came up clean.
- Environmental noise unrelated to this PR (also present on main): headless-sandbox GPU/network-service crashes and `EADDRINUSE :::3847` (another hive instance holds the Slack port). Not attributable to these changes.
- Note: the gauge's live behavior (statusLine firing per response) is driven by Claude Code at runtime and isn't exercised by a headless smoke-launch — I verified it compiles, builds, and boots; I did not observe the gauge updating live.
- **Verdict:** Approve with nits. Solid, well-factored feature with genuinely careful boundary guarding; the one real gap (renderer 2d `limit`/NaN guard) is a cheap defense-in-depth fix, not a blocker.

Gulum:
Thanks for the thorough review — addressed in `9ee73c5`:

- **Renderer 2d NaN guard** ✅ — the handler now drops `limit <= 0` / non-finite values before the progress math (`if (!Number.isFinite(limit) || limit <= 0 || !Number.isFinite(tokens)) return;`). You're right that `NaN` survives the `Math.min/max` clamp, so this was the one real gap.
- **Status early-return intent comment** ✅ — documented that skipping `recordSession` for status ticks is deliberate (telemetry should never write to the registry; the real hooks already record `session_id`), while `transcript_path` is still captured above for every payload shape.
- **pty-parser `/context` sniffer** — that third path isn't in this PR anymore: it lived in the (now-closed) #7 branch's watcher and didn't survive the rebase onto v0.2.1. The two remaining layers are exactly statusLine (exact, push) → transcript poll (backfill until the first tick).
- **Shim timeout 1500ms vs 5000ms** — deliberate. By the time the socket write happens, the in-terminal gauge print already went to stdout, so the user-visible part is done; the forward is fire-and-forget telemetry where a dropped tick self-heals on the next response (statusLine fires per response). The hook path keeps 5000ms because its *return value* gates the agent (block/continue decisions must not be lost to a slow socket).
- **Merge-order note re #4** — resolved by time: #4's named-pipe fix shipped in v0.1.9 and this PR is rebased onto v0.2.1, so the push path is live on Windows on current main, no checklist needed anymore.
- **Footer width** — agreed it's minor; `ctx 235k/1000k (24%)` is the longest realistic form (~19 chars). Happy to shorten to `235k/1M` if it ever collides with the TUI's own footer.

Gulum:
Small UX polish in `5876971`: the 8-segment gauge jumped vertically on the agent card depending on whether the action line had text (idle agents have none, so the line collapsed). The action line now reserves its height and the gauge is pinned to the card's bottom edge with `margin-top: auto` - it never moves again.

chaitanyagiri:
## Re-review — PR #12 (head `5876971`)

**Verdict:** approve / merge — small, correct layout fix on top of the already-clean gauge work (my earlier NaN-guard nit was addressed back in `9ee73c5`).

### What changed since last review (`9ee73c5` → `5876971`)
- **`58769711a` — gauge no longer jumps.** `AgentCard.tsx` only:
  - `minHeight: 16` on the action-text line so it reserves its row even when empty (idle has no action text), instead of collapsing.
  - Context gauge switched from `marginTop: 2` to `marginTop: 'auto'`, pinning it to the card's bottom.

### Correctness
Confirmed in context: the gauge's parent is `display:flex; flexDirection:column; flex:1` (`AgentCard.tsx`), so `marginTop:'auto'` on the last child genuinely pushes it to the bottom of the filled column — not a no-op. Combined with the reserved action row, the gauge now holds a fixed baseline regardless of the lines above. Pure CSS, no logic touched.

### Local test — SEGREGATED re-build @ `5876971`
Isolated worktree, own copied node_modules, torn down after.

| stage | result |
|---|---|
| typecheck | **PASS** |
| build (electron-vite) | **PASS** |
| run smoke (~10s launch) | **LAUNCHED_OK** |

Build doesn't verify the pixel layout itself — but the flexbox reasoning above holds and nothing else is touched.

### Feedback
Good polish, correct flex usage. Merge.

chaitanyagiri:
Merged into `main` (95930db) — thank you @Gulum! The statusLine push is exactly the right call: exact live tokens + the real session window straight from Claude Code, updating after every response. Landed with one small follow-up (zeroing the gauge instantly on `/clear` so the idle window right after a clear doesn't briefly show stale-full). Will ride out in v0.2.2.

Gulum:
One more in `d901bd8`, found by the human reading the monitor tab: its only per-agent meter is the CUMULATIVE token-budget bar, which grows forever and pins at 100% once an agent passes the default 1M budget - trivially misread as a context gauge showing nonsense (3M / 100% red while the actual window sat at 24%). Each agent row now carries a second, clearly-labeled **ctx** row fed by the SAME exact statusLine numbers as the avatar-card gauge (current window tokens vs the real 200k/1M size, same color thresholds), and the budget meter is labeled **budget** with a spend-not-headroom tooltip.

Issue #23, opened 2026-06-06
Terminal: tables shear apart (VT323+DOM renderer), truecolor unreadable on cream, dim text invisible, no copy/paste


Gulum:
## Four stacked terminal problems

**1. Box-drawing tables shear apart; the cursor floats away from the text.**

![table shear](https://raw.githubusercontent.com/Gulum/munder-difflin/assets-bubble-text/screenshots/terminal-table-shear.png)

The terminals run xterm's **DOM renderer** with `fontFamily: VT323`. The DOM renderer trusts the font to be perfectly monospace — but VT323 is missing many glyphs (`↔` U+2194, arrows, parts of box drawing), so the browser substitutes fallback-font glyphs with **different advance widths**; and VT323 has no real bold, so faux bold drifts too. Any line containing such a glyph shifts everything after it.

**2. Truecolor highlights are unreadable on the cream theme.**

![truecolor on cream](https://raw.githubusercontent.com/Gulum/munder-difflin/assets-bubble-text/screenshots/terminal-truecolor-cream.png)

Claude Code highlights with truecolor/256-color codes designed for dark terminals (teal selection bars, orange accents). The light theme can only remap the **16 base ANSI slots** — RGB values pass through untouched and collide on cream.

**3. Dim text is near-invisible.**

![faint text](https://raw.githubusercontent.com/Gulum/munder-difflin/assets-bubble-text/screenshots/terminal-faint-text.png)

Claude Code dims diff-context lines (ANSI faint) — pale-yellow-on-cream is unreadable.

**4. There is no way to copy text out of a terminal.** No copy wiring exists; Ctrl+C is (correctly) SIGINT — and with any accelerated renderer the browser's native copy can't see the terminal content at all (selection lives inside xterm).

## Fix (PR attached)

- `@xterm/addon-webgl` (per-cell glyph drawing → grid integrity for any font coverage; context-loss fallback to DOM). WebGL, not the deprecated canvas addon — its dirty-region tracking garbles scrollback (verified: stale regions, invisible-until-selected text).
- Terminals default to the **designed dark theme** (☀/☾ toggle still offers cream; saved choice wins).
- `minimumContrastRatio: 3` as a safety net for dim/low-contrast output in both themes.
[...]

PR #26, opened 2026-06-06
feat: terminal theme toggle (per-session Claude theme) + Unicode 11 emoji widths

Gulum:
Rebased onto v0.2.1 — and happily, most of this PR is no longer needed: the WebGL renderer, copy/paste wiring and `minimumContrastRatio` all landed in your v0.2.0 work (your 4.5 ratio is the better pick). 🎉

What remains here are the two pieces main doesn't have yet:

1. **Terminal theme toggle with per-session Claude theme mirroring** — a dark/cream toggle on the terminal panel, persisted as `terminalTheme` in the harness config. The interesting bit: each agent's generated `--settings` file gets a scoped `"theme"` key, so the *Claude Code TUI itself* (syntax colors, diff colors) matches the terminal theme — without touching the user's global `claude config`, so terminals outside the app are unaffected. Cream stays the default.
2. **Unicode 11 width tables** (`@xterm/addon-unicode11`) — xterm's default measures emoji like ✅/⚠️ as width 1, so the glyph painted over the following character ("✅Done" rendered merged). Claude Code's own renderer uses Unicode 11 widths; with the addon the TUI's column math and xterm's agree and emoji get their proper two cells.

Diff is now 8 files / ~50 lines.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,6 +22,7 @@
         "@lezer/highlight": "^1.2.3",
         "@uiw/react-codemirror": "^4.25.10",
         "@xterm/addon-fit": "^0.10.0",
+        "@xterm/addon-unicode11": "^0.9.0",
         "@xterm/addon-webgl": "^0.19.0",
         "@xterm/xterm": "^5.5.0",
         "better-sqlite3": "^11.10.0",
@@ -2492,6 +2493,12 @@
       "peerDependencies": {
         "@xterm/xterm": "^5.0.0"
       }
+    },
+    "node_modules/@xterm/addon-unicode11": {
+      "version": "0.9.0",
+      "resolved": "https://registry.npmjs.org/@xterm/addon-unicode11/-/addon-unicode11-0.9.0.tgz",
+      "integrity": "sha512-FxDnYcyuXhNl+XSqGZL/t0U9eiNb/q3EWT5rYkQT/zuig8Gz/VagnQANKHdDWFM2lTMk9ly0EFQxxxtZUoRetw==",
+      "license": "MIT"
     },
     "node_modules/@xterm/addon-webgl": {
       "version": "0.19.0",
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "@lezer/highlight": "^1.2.3",
     "@uiw/react-codemirror": "^4.25.10",
     "@xterm/addon-fit": "^0.10.0",
+    "@xterm/addon-unicode11": "^0.9.0",
     "@xterm/addon-webgl": "^0.19.0",
     "@xterm/xterm": "^5.5.0",
     "better-sqlite3": "^11.10.0",
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -137,6 +137,10 @@ export interface HarnessConfig {
   circuitBreaker?: CircuitBreakerConfig;
   /** Fire native desktop notifications on agent lifecycle events (idle finish / waiting for input). */
   notifications?: boolean;
+  /** Terminal theme — mirrored into each agent's per-session Claude settings
+   *  ("theme" key) at spawn so the TUI's truecolor palette matches. Scoped to
+   *  harness agents only; the user's global Claude theme is never touched. */
+  terminalTheme?: 'light' | 'dark';
   /** Master toggle for the Slack → Michael's-queue integration. */
   slackEnabled?: boolean;
   /** Slack app signing secret (Basic Information → Signing Secret). Never logged. */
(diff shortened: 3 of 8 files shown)


Gulum:
Follow-up commit (d62b33f): the sun/moon toggle now keeps Claude Code's OWN global theme in lock-step (claude config set -g theme dark|light via a new app:setClaudeTheme IPC). Rationale: Claude Code paints truecolor highlights for whichever theme IT thinks is active, bypassing the xterm palette entirely - switching the terminal theme without the matching Claude theme reproduces exactly the unreadable collisions this PR fixes. Running sessions keep their palette until restarted; new/respawned agents pick the theme up. The toggle tooltip states the coupling.

Gulum:
Design correction in the latest commit: the earlier theme-sync used `claude config set -g theme`, which would also restyle the user's own Claude sessions **outside** the app — wrong scope.

Now the terminal theme is mirrored as the standard `theme` key into each agent's **per-session** Claude settings file at spawn (the same `--settings` file that already carries the hooks), so only harness agents are affected and the user's global Claude theme is never touched.

With the palette guaranteed to match per session, the cream terminal is safe again and returns as the **default** (this supersedes the dark-default from the first commit); the ☀/☾ toggle flips both the terminal palette and the theme future agent sessions receive. Running sessions keep their palette until respawned.

Gulum:
One more terminal-metrics fix folded in (ce4c80c), tracked as #31: emoji merged with the following text because xterm's default Unicode 6 width tables count most emoji as ONE cell while Claude Code lays out with modern widths (two cells) - the glyph overflowed its single cell into the next character. @xterm/addon-unicode11 + unicode.activeVersion = 11 aligns the width tables. This PR now fixes #23 and #31.

chaitanyagiri:
## 🤖 Automated PR Review

**Summary** — Adds a per-session terminal theme toggle: the `terminalTheme` config is mirrored into each agent's `--settings` file as a scoped `"theme"` key so the Claude Code TUI's truecolor palette matches the xterm theme, without ever touching the user's global `claude config`. Also wires `@xterm/addon-unicode11` so emoji measure as width 2 and stop merging into following text. Clean, well-commented, correctly scoped.

### ✅ What's good
- **Correctly scoped, no global side effects.** The theme is injected into the same per-session `--settings` file that already carries the hooks (`hookSettings(shim, theme)` → `...(theme ? { theme } : {})`), not via `claude config set -g`. This supersedes the earlier global-config approach (per the discussion thread) and is the right call — the user's own terminals outside the app are untouched.
- **Persistence is safe.** The toggle calls `updateConfig({ terminalTheme })`, and `writeConfig` shallow-merges over the full current config, so the single-key patch can't clobber other settings. Spawn reads it back with `readConfig().terminalTheme ?? 'light'`.
- **Default-to-cream is consistent.** `loadTheme()` defaults to `'light'` and the spawn fallback is also `'light'`, so a fresh install paints cream terminal + light TUI palette in lock-step.
- **Unicode11 addon ordering is correct.** Loaded in `acquireTerminal` before `open()`, with `unicode.activeVersion = '11'` — the Unicode11 width tables are a unicode-service concern independent of the (later) WebGL renderer, so this ordering is fine, and `allowProposedApi: true` (required for `term.unicode`) is already set. The accompanying comment explaining the Unicode 6 → 11 cell-width mismatch is excellent.
- **Dependency hygiene.** `@xterm/addon-unicode11@^0.9.0` is declared in **both** `package.json` and `package-lock.json` (pinned to 0.9.0 with integrity hash), so CI and fresh `npm ci` installs resolve it cleanly.

### 🔧 What could be better
- **Running sessions keep their old palette until respawn.** This is inherent (a running Claude process can't re-read `--settings`) and is already documented in the code comments and the toggle tooltip — so it's a communicated limitation, not a defect. Worth a one-line note in the PR description / changelog so users aren't surprised the visible flip is partial until the next spawn.
- **Minor (optional):** the `'light'` default now lives in two places (`loadTheme()` and the `?? 'light'` at the spawn site). A shared constant would keep them from drifting, but it's trivial.

### 💬 Feedback / questions
- None.

### 🧪 Local test report
Built this PR in an isolated worktree. It adds `@xterm/addon-unicode11`; I installed the PR's declared dependency, then:
- Note: the visible theme toggle + emoji-width rendering are UI behaviors not verifiable in a headless smoke-launch; I verified compile/build/boot and that the new dependency resolves cleanly.
- **Verdict:** Approve. Correct, well-scoped, no global config bleed; builds and boots cleanly with the declared dep. The only thing worth surfacing is that running sessions don't repaint until respawn — already documented in-code, just call it out in the PR notes.

PR #33, opened 2026-06-06
fix: enforce the assistant's send-only contract at the router (fixes #32)

### LINKED ISSUE #32: Direct mail to the send-only prep assistant black-holes - delivered into an inbox nothing ever reads
## Symptom

Direct hive messages addressed to the prep assistant (`"to": "assistant"`) disappear into a black hole: they are delivered into its inbox, but **nothing ever reads that inbox** — the assistant is send-only by design (no composer, excluded from broadcast fan-out *and* from the inbox-wake nudge).

Observed live in a real hive, with poetic timing: the god agent sent the assistant a task brief, got no reaction for ~70 minutes, then sent a **reprimand about the 6 unread messages** — and that reprimand itself sat unread in the very same inbox for hours. The orchestrator reassigned the work after concluding the agent had "gone dark".

## Root cause

`send-only` is the documented design (introducing commit: *"Michael's send-only prep assistant"*; `AssistantRoleNote` even replaces its composer with an explanatory footer). But the **router never enforces it**: only `broadcast` fan-out filters `isAssistant` — a direct `to: "assistant"` message is delivered normally into an inbox that is, by the same design, never read. The design's invariant ("it never receives inbox mail") is assumed in the nudge loop's comment, not guaranteed.

The orchestrator is *invited* into this trap: the assistant messages god from the same registry, its id is visible in the roster, and (in our hive) it even shares the display name "Dwight" with a regular worker.

## Fix (PR attached)

Enforce the contract **at the router**: a direct message to an `isAssistant` agent is bounced to god with a `[bounced — "assistant" is the send-only prep assistant; route work to a real agent]` subject prefix. The sender learns immediately and reroutes; nothing is silently lost; broadcasts are unchanged; the assistant's role stays pure.

Gulum:
## What

Fixes #32 â€” direct mail to the send-only prep assistant black-holed in an inbox nothing ever reads.

## Change

One spot in `HiveManager.route()`: when a resolved target is an `isAssistant` agent, the message is **bounced to god** instead of delivered, with the subject prefixed `[bounced â€” "<id>" is the send-only prep assistant; route work to a real agent]`. The sender (usually the orchestrator) learns immediately that it addressed the wrong agent and can reroute â€” instead of waiting on an agent that will never answer.

Considered and rejected: making the assistant read its inbox (including it in the inbox-wake nudge). That fixes the rot but erodes the role â€” the assistant is deliberately a stateless prompt-enricher with no composer; the right place to enforce a routing contract is the router.

Broadcast behavior is unchanged (already filtered). `requires_reply`/conversation fields ride along on the bounce, so god sees the full original intent.

## Testing

- `npm run typecheck` passes.
- Verified in a live hive: god â†’ assistant direct mail arrives in god's own inbox with the bounce prefix within one router tick; the assistant's inbox stays empty; the enrich flow (assistant â†’ god) is unaffected.

ðŸ¤– Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -527,7 +527,24 @@ export class HiveManager {
       ? Object.keys(reg.agents).filter((a) => a !== msg.from && !reg.agents[a]?.isAssistant && !reg.agents[a]?.archived)
       // Never deliver to self — guards a god → "human" message looping back to god.
       : [resolveTo(msg.to)].filter((t) => t !== msg.from);
-    for (const t of targets) this.deliver(msg, t);
+    for (const t of targets) {
+      // Enforce the assistant's send-only contract AT THE ROUTER: it has no
+      // composer, is excluded from the inbox-wake nudge, and never reads an
+      // inbox — yet direct mail used to be delivered there anyway, where it
+      // rotted unread (observed live: a task brief plus the follow-up
+      // reprimand about the unread inbox, both unread for hours). Bounce such
+      // mail to god instead, so the sender's intent surfaces immediately and
+      // nothing is silently lost.
+      if (reg.agents[t]?.isAssistant) {
+        this.deliver({
+          ...msg,
+          to: godId,
+          subject: `[bounced — "${t}" is the send-only prep assistant; route work to a real agent] ${msg.subject}`
+        }, godId);
+        continue;
+      }
+      this.deliver(msg, t);
+    }
     this.appendLog({ kind: 'message', from: msg.from, to: msg.to, act: msg.act, subject: msg.subject, id: msg.id });
     this.emitMessage(msg, targets);
   }


Gulum:
Rebased onto v0.2.1 - applies cleanly on current main. The bounce integrates with the new router untouched: only the send-only assistant guard in route().

chaitanyagiri:
## 🤖 Automated PR Review

**Summary** — Bounces direct hive mail addressed to the send-only prep assistant back to god with a `[bounced — …]` subject prefix, instead of dropping it into an inbox nothing reads. Correctly scoped to `route()`'s direct-mail path; broadcast fan-out already filters assistants, so the new guard only ever fires for direct mail. Solid, surgical fix for #32.

### ✅ What's good
- Guard sits at the right choke point: in `routeMessage()` the only way an `isAssistant` target reaches the delivery loop is direct mail to its agentId — broadcast (`hive.ts:527`) already excludes assistants, so the bounce can't double-fire or alter fan-out. (`hive.ts:530-543`)
- Bounce is built with `{ ...msg, to: godId, subject }`, so `conversation`, `in_reply_to`, `requires_reply`, `needs_human`, `act`, `body`, and `from` all ride along — god has everything it needs to reroute the work and the thread stays intact.
- Architecturally sound: the assistant's real enrichment input does NOT flow through the hive inbox — it comes via the separate headless `assistant:enrich` IPC → `enrichMessage()` path, and the assistant only ever *writes* its outbox. It's also excluded from every wake/heartbeat/inbox-drain path (`index.ts:329, 342, 390`). So there is genuinely no legitimate inbound hive message an assistant should receive — bouncing 100% of direct mail is correct, not over-broad.
- The comment block is excellent — it captures the live failure (task brief + the reprimand-about-the-unread-inbox, both rotting for hours) so the "why" survives.

### 🔧 What could be better
- Minor/optional: the guard also fires when `msg.from === godId` (god mailing the assistant directly), bouncing the message straight back to its own sender. Harmless and arguably the intended teaching signal ("use enrich, don't mail the assistant"), and there's no loop/hop-cap risk since the bounce delivers rather than re-routes — but a one-line note in the comment that the god-sender case self-bounces would save a future reader a double-take.
- Optional: if no god agent is registered, `godId` falls back to `'god'` and `deliver()` silently drops on a missing inbox (`hive.ts:498`). Pre-existing behavior shared with the normal `resolveTo` path, so not a regression — just flagging that a bounce is only as reliable as god's inbox existing.

### 💬 Feedback / questions
- None. The intent matches the implementation and the existing rebase note (Gulum) confirms it applies cleanly on v0.2.1.

### 🧪 Local test report
- Note: the bounce-to-god routing behavior is exercised at hive runtime, not by a headless smoke-launch; I verified compile/build/boot and reviewed the routing logic statically. The author reports verifying the bounce in a live hive.
- **Verdict:** Approve. Correct, minimal, and well-reasoned fix for #32; the two notes above are optional polish, not blockers.

PR #34, opened 2026-06-07
fix(windows): usage meter read 0/0 - transcript dir encoding

### LINKED ISSUE #10: Usage meter always shows 0 tk / $0 on Windows: transcript dir encoding is POSIX-only
## Symptom

On Windows, the usage meter in Michael's command center (Activity tab) reports **0 / 0 tk for $0.00** for every agent, no matter how much they've actually worked.

Environment: Windows 11, v0.1.8 (also reproduced on a local build of current `main`).

## Root cause

`readAgentUsage()` sums token usage from the Claude Code transcripts under `~/.claude/projects/<encoded-cwd>/`. The encoding in `src/main/transcript.ts` is POSIX-only:

```ts
export function projectDir(cwd: string): string {
  return path.join(os.homedir(), '.claude/projects', cwd.replace(/^\//, '').replaceAll('/', '-'));
}
```

A Windows cwd like `C:\Users\me\dev\app` contains no forward slashes, so the drive colon and backslashes pass through untouched and the function resolves to a directory that doesn't exist. `readAgentUsage()` then takes the `if (!existsSync(dir)) return usage;` early-out and reports zeros.

Claude Code on Windows keys the project folder by replacing **every non-alphanumeric character** with a dash:

```
C:\Users\me\dev\app  →  C--Users-me-dev-app
```

(verified against a real `~\.claude\projects\` on Windows 11).

## Fix

Branch on `win32`:

```ts
const key = process.platform === 'win32'
  ? cwd.replace(/[^a-zA-Z0-9]/g, '-')
  : cwd.replace(/^\//, '').replaceAll('/', '-');
return path.join(os.homedir(), '.claude/projects', key);
```

POSIX behavior unchanged. Verified locally on Windows 11: with this patch the meter shows real token counts and costs, including prior sessions (it sums all transcripts in the project folder). Happy to send a PR if wanted.

Related Windows-support work: #1, #4 (hook server / mempalace), #7 (status watcher).

Gulum:
Fixes #10.

On Windows the usage meter always read 0 tokens / $0.00. The transcript reconciler builds the per-project directory name with the POSIX rule (`cwd.replace(/^\//,'').replaceAll('/','-')`), but Claude Code on Windows encodes **every non-alphanumeric character** — including the drive colon and backslashes — as `-`:

```
C:\Users\me\dev\app  →  C--Users-me-dev-app     (Claude Code, win32)
C:\Users\me\dev\app  →  C:\Users\me\dev\app-…   (current code → no such dir)
```

So the reconciler looked for a directory that never exists and summed zero transcripts.

The fix mirrors Claude Code's actual win32 encoding (`cwd.replace(/[^a-zA-Z0-9]/g, '-')`) on `process.platform === 'win32'`, POSIX path untouched. One small function, verified against real `~/.claude/projects` directories on Windows 11.

(Found while running the harness on Windows — same family as the path issues fixed in v0.1.9, this one is in the cost/usage path. Rebased onto v0.2.1; note the OTel live path in v0.2.0 doesn't replace this — the reconciler/fallback path still reads transcripts.)

### DIFF
--- a/src/main/transcript.ts
+++ b/src/main/transcript.ts
@@ -6,9 +6,14 @@ import { estimateCostUsd, normalizeModel } from './pricing';
 /** Resolve the Claude Code transcript directory for a given working directory.
  *  Claude Code stores per-project transcripts under ~/.claude/projects, keying
  *  each project by its absolute cwd with the leading slash dropped and every
- *  remaining slash turned into a dash (e.g. /Users/me/app → Users-me-app). */
+ *  remaining slash turned into a dash (e.g. /Users/me/app → Users-me-app). On
+ *  Windows EVERY non-alphanumeric character is dashed (drive colon and
+ *  backslashes included): C:\Users\me\app → C--Users-me-app. */
 export function projectDir(cwd: string): string {
-  return path.join(os.homedir(), '.claude/projects', cwd.replace(/^\//, '').replaceAll('/', '-'));
+  const key = process.platform === 'win32'
+    ? cwd.replace(/[^a-zA-Z0-9]/g, '-')
+    : cwd.replace(/^\//, '').replaceAll('/', '-');
+  return path.join(os.homedir(), '.claude/projects', key);
 }
 
 export interface AgentUsage {


chaitanyagiri:
## 🤖 Automated PR Review

**Summary** — Targeted, minimal fix for the Windows usage meter reading 0/0 (#10). The reconciler keyed `~/.claude/projects` with the POSIX rule (`replace(/^\//,'').replaceAll('/','-')`), which never matches Claude Code's win32 dir names where the drive colon and backslashes are all encoded. Gating the win32 branch on `process.platform === 'win32'` with `cwd.replace(/[^a-zA-Z0-9]/g,'-')` is the right shape and leaves the POSIX path byte-for-byte unchanged.

### ✅ What's good
- Correct root cause and correct seam. `projectDir` ([src/main/transcript.ts:10-15](https://github.com/chaitanyagiri/munder-difflin/pull/34/files)) is the single key-derivation point, so one branch fixes both `existsSync` and `readdirSync` downstream — no scatter.
- The win32 regex matches Claude Code's actual encoding: `C:\Users\me\app → C--Users-me-app`. The double dash (from `:\`) is exactly what Claude Code writes, because it dashes *every* non-alphanumeric char individually with no collapsing. I confirmed the no-collapse model against real on-disk dirs in `~/.claude/projects` on this machine.
- Platform guard is tightly scoped — the POSIX expression is untouched, so macOS/Linux behavior cannot regress. CI green (typecheck + build).
- Doc comment updated to spell out the win32 rule with a concrete example — future readers won't re-derive it.

### 🔧 What could be better
- (Optional) UNC paths (`\\server\share\proj`) encode to leading dashes (`----server--share--proj`). That's almost certainly what Claude Code does too (it dashes every separator), but if you have a Windows box handy, a one-line confirm against a real UNC-rooted project would close the last gap. Not blocking.
- (Pre-existing, not this PR) The on-disk POSIX dirs here are actually named `-Users-...` (leading dash), i.e. Claude Code dashes the *leading* separator rather than dropping it — yet the POSIX branch strips it with `replace(/^\//,'')`. This PR doesn't touch that and shouldn't, but it suggests the macOS/Linux key may have its own latent mismatch worth a follow-up issue. Flagging for awareness only.

### 💬 Feedback / questions
- None blocking. Consider a tiny unit test asserting `projectDir` with a mocked `process.platform === 'win32'` and a `C:\…` input, so the win32 branch is regression-guarded in CI (it can't be exercised on the Linux/macOS runners otherwise).

### 🧪 Local test report
- Honest scope note: this fix only changes the **win32** branch; I built/ran on macOS, so the Windows directory-name path is not exercised at runtime here. I verified it compiles, builds, boots, and leaves the POSIX path unchanged, and reviewed the win32 regex against Claude Code's encoding. The author reports verifying against real `~/.claude/projects` on Windows 11.
- **Verdict:** Approve. Small, correct, well-scoped fix; the win32 regex matches Claude Code's documented encoding, the POSIX path is provably untouched, and the only follow-ups (a win32 unit test, optional UNC confirm) are non-blocking nice-to-haves.

PR #39, opened 2026-06-07
fix: text-select mouse cursor invisible on cream - inked I-beam with halo

Gulum:
Follow-up to the caret fix that landed in v0.2.0: `color-scheme: light` + `caret-color` fixed the BLINKING caret, but the hovering **I-beam** (the mouse text-select pointer) is an OS cursor that CSS color hints cannot touch - several Windows cursor schemes (incl. HiDPI variants) draw it white with a hairline outline, so it disappears over the cream composer.

One CSS rule: text fields get a custom SVG I-beam drawn in ink wearing a cream halo, readable on the cream inputs and the dark terminal alike. Hotspot centered on the stem; falls back to the UA `text` cursor where SVG cursors are unsupported.

### DIFF
--- a/src/renderer/src/design/global.css
+++ b/src/renderer/src/design/global.css
@@ -37,6 +37,16 @@ button, input, textarea, select {
 /* Belt + suspenders for the caret: follow the ink, never the UA scheme. */
 input, textarea { caret-color: var(--cth-ink-900); }
 
+/* The MOUSE text-select pointer (the I-beam) is a different beast from the
+   caret: it comes from the OS cursor scheme, which `color-scheme` can't touch.
+   On Windows several schemes (and HiDPI variants) draw it white with a hairline
+   outline — invisible hovering over our cream inputs. Replace it with our own
+   inked I-beam wearing a cream halo, so it reads on cream AND on the dark
+   terminal. Falls back to the UA `text` cursor if SVG cursors are unavailable. */
+input, textarea, [contenteditable='true'] {
+  cursor: url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' width='17' height='18'><g fill='none' stroke='%23fdf6e3' stroke-width='4' stroke-linecap='round'><path d='M5 2.5h7M8.5 3v12M5 15.5h7'/></g><g fill='none' stroke='%23261d31' stroke-width='1.6' stroke-linecap='round'><path d='M5 2.5h7M8.5 3v12M5 15.5h7'/></g></svg>") 8 9, text;
+}
+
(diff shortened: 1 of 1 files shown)


chaitanyagiri:
## 🤖 Automated PR Review

**Summary** — Sound, well-scoped follow-up to the v0.2.0 caret fix. Where `color-scheme: light` + `caret-color` rescued the blinking *caret*, this rescues the hovering *mouse* I-beam — an OS cursor CSS color can't reach — by swapping in a custom SVG drawn in ink with a cream halo, with a sensible UA fallback. +10/-0, one CSS file, CI green.

### ✅ What's good
- Correct diagnosis and the comment (`global.css:39-45`) explains *why* the OS cursor scheme is the culprit and why `color-scheme` can't fix it — future-you will thank present-you.
- The SVG is well-formed and the design is robust: a thick cream halo stroke (`stroke='%23fdf6e3'` width 4) sits *under* the ink glyph (`stroke='%23261d31'` width 1.6), so the I-beam carries its own contrast and reads on both cream and the dark terminal. `#` correctly encoded as `%23` for an unquoted `utf8` data-URI.
- Hotspot `8 9` is right: the stem is at `x=8.5`, vertical span `y≈3→15`, so `(8,9)` lands dead-center on the stem — clicks land where the user expects (`global.css:48`).
- The `, text` UA fallback is present (`global.css:48`), so browsers without SVG-cursor support degrade gracefully instead of falling back to default arrow.
- Scoped to `input, textarea, [contenteditable='true']` — deliberately excludes `button`/`select`, so it doesn't paint an I-beam on clickable controls.

### 🔧 What could be better
- Minor/optional: bare `input` also matches non-text inputs (`type=checkbox|range|radio|button|color`), where an I-beam is semantically off. This mirrors the existing `caret-color` rule on line 39 so it's *consistent* with the codebase, and these types are rare in this app — but if you want to be precise, `input:is([type=text],[type=search],[type=email],[type=url],[type=password],[type=tel],[type=number]), input:not([type]), textarea, [contenteditable='true']` would scope it to true text fields. Not blocking.

### 💬 Feedback / questions
- None.

### 🧪 Local test report
- Honest scope note: this is a CSS cursor change. The build confirms it compiles/builds/boots and the CSS is well-formed, but the actual I-beam appearance over the cream composer is a visual behavior I cannot verify in a headless launch. (The `EADDRINUSE :::3847` and GPU/network-service messages in my run log are pre-existing environmental noise, unrelated to this diff.)
- **Verdict:** ✅ LGTM — ship it. Clean, well-reasoned, correctly scoped; the only note is an optional precision tweak on the `input` selector that matches existing code.

Issue #42, opened 2026-06-07
design: let the god reign over tasks.json - human kanban edits are context-free state changes


Gulum:
Spun out of the design note on #41 so it can be decided on its own.

**Since v0.2.x the god genuinely maintains `hive/tasks.json`** — it decomposes the work, knows the context behind every card, and tracks who is on what. But the kanban UI still offers the human full write access: an assignee picker, a status dropdown, an assign button.

**The problem: human kanban edits are context-free state changes.** When the human reaches in and flips a card's status or assignee, the orchestrator never learns *why*, and the card's history stops matching the floor's reality. The degenerate case makes it obvious: dragging a **done** card back to `todo` or `blocked` carries zero context about what was wrong with the result — the god re-discovers the card in a state it can't explain, and the assigned worker gets nothing at all.

**Proposal: let the god reign over the cards.** The human's natural channel already exists — talking to the orchestrator ("this isn't done, the avatars are still missing"), who then updates the ledger *with* context, reassigns, and briefs the worker. Concretely, in increasing order of strictness:

1. **Notify-on-edit (minimal):** keep the direct edits, but the harness drops a note into the god's inbox whenever the human changes the ledger ("card X moved done→blocked by the human"), so the orchestrator at least knows the ground shifted under it.
2. **Request-a-change (preferred):** the kanban's write controls become a "request" affordance — instead of editing `tasks.json`, they compose a hive message to the god ("human requests: reopen X, reason: …"). The god makes the actual edit with full context. The kanban stays a rich READ surface (plus #41's office board as the at-a-glance view).
3. **Read-only:** drop the write path entirely; all task mutations flow through conversation.

Option 2 keeps the human's one-click convenience while preserving a single writer with context — the same single-scribe principle the hive already applies to `board.md`.

chaitanyagiri:
**Triage: valid design question, needs an owner call — not a bug, not roadmapped yet.** @chaitanyagiri

Grounded the premise in the code — it's accurate:
- The human kanban **does** write \`tasks.json\` directly today: \`TasksKanban.moveTask\`/\`assign\` → \`window.cth.hiveWriteTasks(next)\` → IPC \`hive:writeTasks\` → \`hive.writeTasks()\` (preload:463, index.ts:907, hive.ts:611). So a status/assignee flip is a context-free write, exactly as described — the orchestrator never learns *why*.
[...]

PR #45, opened 2026-06-07
fix: all human dispatch flows through the god - worker picker becomes a suggestion

### LINKED ISSUE #44: design: human dispatch bypasses the orchestrator - direct-to-worker (default: broadcast!) should flow through the god
Sibling of #42, but on the **messaging** path instead of the ledger: the Command Center's DISPATCH box lets the human deliver a task **directly into any worker's inbox**, bypassing the orchestrator entirely.

**Why that's broken:**

1. **The default target is `broadcast`** — the same task lands in *every* worker's inbox at once, and N workers race the same job. The GitHub-issue assign flow even sets `broadcast` explicitly.
2. **It bypasses the god's entire job.** The system prompt makes the orchestrator own decomposition, dispatch decisions and the ledgers ("write it as a 4-part contract", "keep board.md and tasks.json accurate") — a direct human dispatch produces none of that: no OBJECTIVE/OUTPUT/TOOLS/BOUNDARIES contract, no card in `tasks.json`, no board entry. The god's floor awareness silently diverges from reality.
3. **The picker even offers the send-only prep assistant**, whose inbox nothing ever reads (#32's black hole, resolved at the router by #33 — but the UI shouldn't aim there in the first place).

The one legitimate need behind the worker picker — "I'd like Pam to do this" — belongs **inside the brief to the god**, not as a bypass around him.

**Proposal: all human dispatch flows through the god.**

- The dispatch always mails the god (`Task from the human`).
- The worker dropdown becomes **"suggested owner"** — forwarded in the brief as *"the human suggests Pam for this — your call as orchestrator"*. Default: *Michael decides*.
- `broadcast` is removed; the assistant leaves the list.
- The god decomposes, writes the card (with assignee), and dispatches with a proper contract — the human's intent arrives with full context instead of around it.

Together with #42 (single-writer over tasks.json) this completes the principle: **the human talks to the organization through its manager, not past him.**

**Implementation: PR follows.**

Gulum:
Fixes #44.

Small, surgical change to the Command Center's DISPATCH section (one component):

- **Every dispatch now mails the god** (`Task from the human`) instead of writing directly into a worker's inbox.
- The worker dropdown becomes **SUGGESTED OWNER**: picking Pam appends *"(The human suggests Pam (pam-…) for this — your call as orchestrator.)"* to the brief. Default is **Michael decides**.
- **`broadcast` is gone** — it was the DEFAULT target, and it delivered the same task to every worker's inbox at once (the GitHub-issue assign flow even set it explicitly; that flow now suggests no one and lets the god decompose).
- The **send-only assistant leaves the picker** — direct mail to it black-holed upstream (#32) and is bounced by the router with #33; the UI shouldn't aim there at all.

Rationale in the issue: direct dispatch bypassed everything the god prompt mandates (4-part dispatch contracts, tasks.json cards, board awareness), so the orchestrator's picture of the floor silently diverged from reality. Together with #42's direction (single writer over tasks.json) this completes the principle: **the human talks to the organization through its manager, not past him.**

Typecheck + build clean; the change is renderer-only and rides the existing `hiveSend` rail.

### DIFF
--- a/src/renderer/src/components/CommandCenterPanel.tsx
+++ b/src/renderer/src/components/CommandCenterPanel.tsx
@@ -224,7 +224,7 @@ function FloorTab({ seed }: { seed: { text: string; seq: number } }) {
   // Per-agent token limit (overrides the floor budget for that agent), keyed by id.
   const [agentTokenCaps, setAgentTokenCaps] = useState<Record<string, number>>({});
   const [restarting, setRestarting] = useState<string | null>(null);
-  const [dispatchTo, setDispatchTo] = useState<string>('broadcast');
+  const [dispatchTo, setDispatchTo] = useState<string>(''); // '' = Michael decides
   const [dispatchText, setDispatchText] = useState('');
   const [dispatchMsg, setDispatchMsg] = useState<string | null>(null);
   // ── ISSUES section state ──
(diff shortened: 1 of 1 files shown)


chaitanyagiri:
**Verdict: approve** — surgical (1 file, +23/-10), clean, no must-fixes. `Fixes #44`.

### What it does
DISPATCH no longer mails a worker's inbox directly. **Every** dispatch now goes to the god (`subject: "Task from the human"`); the dropdown becomes **SUGGESTED OWNER** (default *"Michael decides"*) — a picked worker is appended to the brief as *"(the human suggests X — your call as orchestrator)"*. `broadcast` is removed, and the send-only assistant is filtered out of the picker.

### Verified locally
- `isAssistant` is a real field on the agent shape (`CommandCenterPanel.tsx:301`, store), so `!a.isGod && !a.isAssistant` genuinely drops the assistant — not a silent no-op.
- Rides the existing `hiveSend(..., 'human')` rail (no new IPC). Default `''` handled in both `dispatch()` and `assignIssue()` (the GitHub-issue flow now lets Michael decompose instead of broadcasting).
- Renderer-only.

### Why it matters
Closes the bypass the god-prompt mandates against — direct dispatch produced no 4-part contract, no `tasks.json` card, no board entry, so the orchestrator's floor picture silently diverged; and the old `broadcast` **default** blasted the same task to every worker. Pairs cleanly with #42's single-writer direction.

### Local test — segregated build (isolated worktree + own node_modules) @ `3368682`
| stage | result |
|---|---|
| typecheck | **PASS** |
| build (electron-vite) | **PASS** |
| run smoke (~10s) | **LAUNCHED_OK** |

PR #46, opened 2026-06-07
feat: monitor tab gets a real context-window row - the budget meter was being misread

Gulum:
Follow-up to the just-merged #12, prompted by the human reading the monitor tab: its only per-agent meter is the CUMULATIVE token-budget bar, which grows forever and pins at 100% once an agent passes the default 1M budget - trivially misread as a context gauge showing nonsense (3M / 100% red while the actual window sat at a comfortable 24%).

Each agent row now carries a second, clearly-labeled **ctx** row fed by the SAME exact statusLine numbers as the avatar-card gauge (current window tokens vs the real 200k/1M size, same color thresholds: accent / lemon at 75% / coral at 88%), with a no-status-tick-yet placeholder before the first response. The budget meter is labeled **budget** and its tooltip now says spend-not-headroom in as many words.

Renderer-only, one component; typecheck + build clean.

### DIFF
--- a/src/renderer/src/components/CommandCenterPanel.tsx
+++ b/src/renderer/src/components/CommandCenterPanel.tsx
@@ -475,14 +475,46 @@ function FloorTab({ seed }: { seed: { text: string; seq: number } }) {
                   background: 'var(--cth-paper-200)', boxShadow: 'inset 0 0 0 1px var(--cth-ink-300)', color: 'var(--cth-ink-700)'
                 }}>{lastTool[a.id]}</span>
               )}
+              <span style={{ fontFamily: 'var(--cth-font-mono)', fontSize: 10, color: 'var(--cth-ink-300)', flexShrink: 0 }}>budget</span>
               <span style={{ fontFamily: 'var(--cth-font-mono)', fontSize: 11, color: 'var(--cth-ink-900)', width: 56, textAlign: 'right' }}>{fmtTokens(tokens)}</span>
               <div
-                title={`${tokens.toLocaleString()} of ${denom.toLocaleString()} tokens${agentCap ? ' (agent limit)' : ' (floor budget)'}`}
+                title={`CUMULATIVE session usage: ${tokens.toLocaleString()} of ${denom.toLocaleString()} tokens${agentCap ? ' (agent limit)' : ' (floor budget)'} — not the context window`}
                 style={{ width: 96, height: 8, background: 'var(--cth-cream-200)', boxShadow: 'inset 0 0 0 1px var(--cth-ink-300)', flexShrink: 0 }}
               >
                 <div style={{ width: `${pct}%`, height: '100%', background: meterColor }} />
               </div>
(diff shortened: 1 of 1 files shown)


chaitanyagiri:
## Review — PR #46: monitor tab gets a real context-window row (the budget meter was being misread)

**Verdict:** approve (+ one nit) — a clean, well-scoped follow-up to #12 that fixes a genuine misread; mergeable and green.

### What it does
- Adds a second, clearly-labeled **ctx** row under each agent in the monitor/floor list, fed by the **same** `contextTokens`/`contextLimit` store fields as the avatar-card gauge (current-window tokens vs the real 200k/1M size) — so the two never disagree.
- Re-labels the existing cumulative bar **budget** and rewrites its tooltip to say *cumulative session spend, not the context window* — the exact confusion that prompted this (a 3M/100%-red budget bar misread as a context gauge while the real window sat at ~24%).
- Shows a `no status tick yet` placeholder before the first statusLine tick. Renderer-only, one component (`CommandCenterPanel.tsx`, +33/-1), no dependency or main-process changes.

### What's good
- `Math.min(100, …)` caps the ctx fill so it can't overflow the way the cumulative budget bar does.
- Non-null assertions are safely guarded by `contextTokens !== undefined && contextLimit`.
- Reuses the established color thresholds (`coral ≥88` / `lemon ≥75` / else accent) and the same fields as `AgentCard.tsx`, so the floor list and the avatar card stay numerically consistent.
- Composes correctly with the `/clear`-immediate gauge-zero work: once that resets `contextLimit → undefined`, this row falls cleanly to the placeholder until the next tick. No interaction bug.

### What could be better
- (nit) The avatar-card gauge is a quantized 8-segment bar (`progress ≥ 7` → coral ≈ 87.5%, `≥ 6` → lemon = 75%); this row uses continuous `cpct ≥ 88 / ≥ 75`. A reading at, say, 87.6% would show coral on the card but lemon here — a sub-1% boundary cosmetic difference only. Harmless; flagging for exactness since the PR claims identical thresholds.

Isolated worktree @ `3530b69`, own node_modules, **`npm install` against the lockfile** (required — `@xterm/addon-unicode11` from #26 isn't in the copy-cache; without it typecheck false-negatives on `terminalPool.ts`, a file this PR doesn't touch). Torn down after.

| stage | result |
|---|---|
| typecheck | PASS |
| run smoke (~10s launch) | LAUNCHED_OK |

Also confirmed: **MERGEABLE / CLEAN** on current `main`, and a 3-way merge onto the in-flight v0.2.2 cut is conflict-free (this PR edits `FloorTab` ~L475; the `/clear` gauge-zero edits `onUserPrompt` ~L167 — different functions). Build does **not** exercise the actual pixel rendering or live statusLine ticks — those are visual/runtime, verified by reading the logic.

### Feedback
Solid, low-risk fix that directly removes the budget-vs-context ambiguity a human actually hit. The one thing worth a glance is the 88/75-vs-quantized-7/6 threshold nuance — purely cosmetic, not a blocker.

PR #51, opened 2026-06-07
Fix Windows agent spawn (error 193): route .cmd/script shims through …

### LINKED ISSUE #22: Windows: GOD agent fails to spawn with error 193 when Claude Code is installed via npm
### What happened?

On Windows, when Claude Code is installed via npm i -g @anthropic-ai/claude-code, the GOD agent (Michael) fails to start. The spawn hits error code 193 (ERROR_BAD_EXE_FORMAT) because node-pty's CreateProcess finds the extensionless Unix shell script that npm places in the global bin directory instead of a valid Windows executable.
npm creates three files: claude (extensionless shell script), claude.cmd, and claude.ps1. cmd.exe resolves claude.cmd through PATHEXT, but node-pty calls CreateProcess directly, which does not. It finds the extensionless claude, tries to execute it as a binary, and fails.

Confirmed working on a separate Windows 11 laptop where Claude Code was installed via WinGet (winget install Anthropic.ClaudeCode), which provides a real claude.exe. Same Munder Difflin version, different Claude Code install method.

Manually spawned agents work when the command field is set to cmd /c claude ..., but the GOD agent's auto-creation uses a hardcoded command that can't be overridden.

Suggested fix: on Windows, spawn via cmd /c claude so PATHEXT resolution works regardless of install method. Same root cause as [emdash #1703](https://github.com/generalaction/emdash/issues/1703).

### Steps to reproduce

1. Install Claude Code on Windows via npm i -g @anthropic-ai/claude-code
2. Confirm where claude shows an extensionless file and .cmd (no .exe)
3. Install Munder Difflin (prebuilt Windows installer)
4. Complete onboarding wizard
5. Michael "clocks in" then disappears, leaving empty floor
6. Manually adding an agent also fails with "Cannot create process, error code: 193"

### Logs / screenshots

```shell
Failing machine (Windows 11 AVD, Claude Code installed via npm):

> where claude
C:\Program Files\nodejs\npm-global\claude
C:\Program Files\nodejs\npm-global\claude.cmd

Error when spawning any agent: "Cannot create process, error code: 193"

Working machine (Windows 11 laptop, Claude Code installed via WinGet):

> where claude
C:\Users\<user>\AppData\Local\Microsoft\WinGet\Packages\Anthropic.ClaudeCode_Microsoft.Winget.Source_8wekyb3d8bbwe\claude.exe
```

### macOS version

N/A - Windows 11 build 10.0.26200.8457 (Azure Virtual Desktop)

### Node version

v24.13.0

v2.1.126

### Pre-flight

jjc6676:
## Fix Windows agent spawn (error 193) — route `.cmd`/script shims through cmd.exe

Closes #22.

### Problem
On Windows, spawning any agent — including the auto-created GOD agent (Michael) — fails with `Cannot create process, error code: 193` (`ERROR_BAD_EXE_FORMAT`) when Claude Code is installed via **npm**.

### Root cause
`node-pty` (and `child_process`) launch via `CreateProcess`, which only runs real `.exe`/`.com` binaries and ignores `PATHEXT`. `npm i -g @anthropic-ai/claude-code` installs an **extensionless** `claude` shell script alongside `claude.cmd` (no `.exe`). `resolveCommand()` takes the first `where claude` hit — the extensionless script — and hands it to `node-pty`, which can't execute it → error 193. `cmd /c claude …` works because cmd honors `PATHEXT`. WinGet installs ship a real `claude.exe`, so they don't hit this. (Same root cause as generalaction/emdash#1703.)

### Fix
When the resolved target on Windows isn't a real executable, launch it through `cmd.exe /c`. Applied at every spawn site:
- `src/main/pty.ts` — agent terminals. This is what fixes the GOD agent + manually-added agents; it sits in `PtyManager.spawn`, so it also covers the GOD agent's hardcoded command.
- - `src/main/assistant.ts` and `src/main/reflect.ts` — the headless `claude -p` calls (enrichment / memory condense).
A real `claude.exe` (WinGet) matches the `.exe` guard and still launches directly. macOS/Linux are untouched.

### Verification
Patched into the compiled 0.2.2 build and verified end-to-end on Windows 11 (Claude Code via npm, Node v24.13.0): **Michael now clocks in and stays, and manually-added agents spawn with no error 193.** The two headless paths use the same wrapper; note they pass the prompt in `args`, so `cmd.exe` leaves quoted args intact except for `%VAR%` expansion — a follow-up could pass the prompt via stdin to remove even that edge case.

### DIFF
--- a/src/main/assistant.ts
+++ b/src/main/assistant.ts
@@ -123,7 +123,17 @@ export function enrichMessage(req: EnrichRequest): Promise<EnrichResult> {
 
     let child;
     try {
-      child = spawn(exe, args, {
+      // Windows: route the npm `.cmd`/extensionless `claude` shim through cmd.exe;
+      // CreateProcess can't exec it directly (EINVAL on modern Node / error 193).
+      // The prompt rides in `args`; cmd leaves quoted args intact aside from %VAR%
+      // expansion (a follow-up could pass the prompt via stdin to avoid even that).
+      let spawnFile = exe;
+      let spawnArgs = args;
+      if (process.platform === 'win32' && !/\.(exe|com)$/i.test(exe)) {
+        spawnFile = process.env.ComSpec || 'cmd.exe';
+        spawnArgs = ['/c', exe, ...args];
+      }
+      child = spawn(spawnFile, spawnArgs, {
         cwd: req.cwd,
         env: {
           ...process.env,
--- a/src/main/pty.ts
+++ b/src/main/pty.ts
@@ -130,7 +130,18 @@ export class PtyManager {
         }
       })();
 
-      const proc = pty.spawn(resolved, opts.args ?? [], {
+      // Windows: node-pty's CreateProcess can't run the npm `.cmd`/extensionless
+      // `claude` shim directly (ERROR_BAD_EXE_FORMAT, error 193). Route non-.exe
+      // targets through cmd.exe so PATHEXT resolves the shim. This also covers the
+      // auto-spawned GOD agent's hardcoded command. A real claude.exe (WinGet)
+      // matches the .exe guard and still launches directly.
+      let file = resolved;
+      let spawnArgs = opts.args ?? [];
+      if (process.platform === 'win32' && !/\.(exe|com)$/i.test(resolved)) {
+        file = process.env.ComSpec || 'cmd.exe';
+        spawnArgs = ['/c', opts.command, ...(opts.args ?? [])];
+      }
+      const proc = pty.spawn(file, spawnArgs, {
         name: 'xterm-256color',
         cols: opts.cols ?? 100,
         rows: opts.rows ?? 30,
--- a/src/main/reflect.ts
+++ b/src/main/reflect.ts
@@ -291,7 +291,17 @@ export class MemoryReflector {
 
       let child;
       try {
-        child = spawn(exe, args, {
+        // Windows: route the npm `.cmd`/extensionless `claude` shim through cmd.exe;
+        // CreateProcess can't exec it directly (EINVAL on modern Node / error 193).
+        // The prompt rides in `args`; cmd leaves quoted args intact aside from %VAR%
+        // expansion (a follow-up could pass the prompt via stdin to avoid even that).
+        let spawnFile = exe;
+        let spawnArgs = args;
+        if (process.platform === 'win32' && !/\.(exe|com)$/i.test(exe)) {
+          spawnFile = process.env.ComSpec || 'cmd.exe';
+          spawnArgs = ['/c', exe, ...args];
+        }
+        child = spawn(spawnFile, spawnArgs, {
           cwd: home,
           env: { ...process.env, PATH: userShellPath(), ...this.getMemoryEnv() } as Record<string, string>,
           stdio: ['ignore', 'pipe', 'pipe']


chaitanyagiri:
## Review — PR #51: Fix Windows agent spawn (error 193) — route `.cmd`/script shims through cmd.exe

**Verdict:** approve + nits — correct, well-scoped fix for confirmed bug #22; matches the validated workaround and builds green. One security follow-up to weigh before/after merge (below).

### What it does
- On Windows, when the resolved spawn target isn't a real `.exe`/`.com`, routes it through `cmd.exe /c` so `PATHEXT` resolves the npm `claude.cmd`/extensionless shim that `CreateProcess` (node-pty + child_process) can't exec directly → kills error 193.
- Applied at all three spawn sites: `src/main/pty.ts` (`PtyManager.spawn` — covers the auto-spawned GOD agent + manually-added agents) and the two headless `claude -p` callers `src/main/assistant.ts` (enrichment) and `src/main/reflect.ts` (memory condense).
- A real `claude.exe` (WinGet) matches the `.exe` guard and still launches directly; macOS/Linux are untouched (`process.platform === 'win32'` gate). Closes #22.

### What's good
- Root cause is exactly right and matches @billrehm's repro + my code-trace on `main` (`resolveCommand` took `where`'s extensionless line[0] unconditionally). This is the cleaner of the two fixes I suggested (cmd.exe wrapper vs PATHEXT filtering).
- Verified end-to-end on Win11 / npm install / Node 24 by the author: Michael clocks in, manual agents spawn. That's the platform path my macOS build can't exercise.
- Guard is precise (`!/\.(exe|com)$/i`), uses `process.env.ComSpec` with a sane fallback, and is consistently identical across all three sites.
- `pty.ts` passes the bare `opts.command` to cmd (lets PATHEXT re-resolve) — robust.

### What could be better
- **(security — see section below; non-blocking but worth a decision)** The two headless paths pass a large free-text `-p` prompt as a `cmd.exe /c` arg. Shifting that prompt to **stdin** removes the shell-metacharacter / `%VAR%` surface entirely. The author already called this out (PR body + #22 comment) — just flagging it for the merge call.
- (nit) `assistant.ts`/`reflect.ts` pass the resolved `exe` path to cmd while `pty.ts` passes the bare `opts.command`. Both work (cmd applies PATHEXT to a full path too), but unifying on one form would read cleaner.

Isolated worktree @ `aadffac`, own copied node_modules + `npm install` (incl. `@xterm/addon-unicode11`), torn down after.
| stage | result |
|---|---|
| npm install | PASS |
| typecheck | PASS |
| run smoke (~10s launch) | LAUNCHED_OK |

Scope: typecheck/build/launch are on **macOS**, so they confirm the change compiles cleanly and doesn't regress the non-Windows path — but they **cannot** exercise the `win32` branch itself. That path rests on the author's Win11 end-to-end verification, which matches the known-good `cmd /c claude` workaround.

### Security / prompt-injection audit
- **Prompt-injection scan: clean.** Diff is code + explanatory English comments only — no agent-facing prompt text, fixtures, or instruction-override payloads.
- **Process-spawn audit — one latent injection surface (low/defense-in-depth):** routing argv through `cmd.exe /c` reintroduces the Windows "BatBadBut"-class quoting surface (CVE-2024-27980 family). Node/node-pty quote args for `CommandLineToArgvW` but do **not** neutralize cmd metacharacters (`& | < > ^`) or `%VAR%` expansion. For `pty.ts` the args are flags / locally-typed command args → the local user already owns the shell, so negligible. For `assistant.ts` + `reflect.ts` the `-p` prompt is **generated text** (messages, evicted agent memory) — and in this project agents ingest external text (GitHub issues/PRs) into memory, so untrusted content can reach a `cmd.exe /c` line. Real-world exploitability is limited (prompts contain spaces → Node quotes them → `&|` are protected inside quotes; residual vectors are `%VAR%` leakage and a crafted embedded-`"` breakout), but it's non-zero. **Fix = pass the prompt via stdin** (the author's own recommendation) — eliminates the surface for both headless paths.

@chaitanyagiri — looping you in per the standing security review: this PR is a correct fix for #22 and safe to merge as-is for the GOD/agent terminal path, but it adds a (small, author-acknowledged) shell-metacharacter surface on the two headless `claude -p` calls. Your call whether to require the stdin hardening in this PR or land it and fast-follow.

### Feedback
Ship-quality fix for a confirmed bug, verified on the real platform. The single thing to decide: take the `-p`-via-stdin hardening now or as a fast follow — everything else is approve.

PR #47, opened 2026-06-07
[codex] Add Codex provider preset

pdurlej:
## What changed

- Added a small shared provider preset model for `claude`, `codex`, and `custom`.
- Added a Codex option in Add Agent that defaults the command to `codex`.
- Carries the provider through spawn, restore, and restart flows.
- Gates Claude-only setup so Codex/custom agents do not receive Claude Code flags, settings, telemetry env, model args, resume args, or permission pre-accept writes.
- Documents the initial Codex support boundary.

## Why

This starts the provider layer discussed in #21 without trying to build full Codex parity. Codex support here is intentionally limited to terminal spawning plus shared workspace/env so Claude Code and Codex can coexist in the harness.

## Validation

- `npm run typecheck`

Note: dependency install for validation used `npm ci --ignore-scripts` because the local Node 26 environment cannot build `better-sqlite3@11.10.0` native bindings.

PR #53, opened 2026-06-07
[codex] Bridge hive mail to terminal work orders

pdurlej:
Draft stacked on #47. Please review this after #47 lands; until then the GitHub diff includes the provider-preset commits from #47.

## Why this exists

#47 makes Codex/custom agents spawnable without leaking Claude-only flags, hooks, telemetry, or hive inbox assumptions onto them. That is the right v1 provider boundary, but it leaves one important collaboration gap: Michael can see a Codex terminal on the floor, yet direct hive mail to that agent cannot be delivered because Codex does not drain the Claude hive inbox protocol.

This PR adds the smallest bridge across that gap. Michael can now route work to a Codex/custom terminal as a terminal work order, without pretending those providers support Claude hooks or inbox delivery.

## What this adds

- Direct hive mail to non-Claude agents now becomes a `hive:terminalHandoff` IPC event instead of an unread inbox file.
- The renderer turns that handoff into a structured `WORK ORDER FROM HIVE` prompt and queues it through the existing per-agent `messageQueues` path.
- The existing idle-only PTY drain submits the work order to the target terminal when the agent is idle, so this reuses the same safe delivery mechanics as human-composed queued messages.
- Work orders carry the original hive message id, sender, subject, act, issued timestamp, body, and whether a reply is expected.
- If the renderer is unavailable and cannot accept the handoff, the router falls back to Michael's inbox so the message is not silently lost.
- Repeated handoff events are deduped by hive message id before they can create duplicate terminal work.
- The hive log records `terminal-handoff` audit entries so terminal delivery can be distinguished from normal inbox delivery.

## What this enables

After #47 and this PR, the flow becomes:

1. The user or Michael keeps using the hive as the coordination surface.
2. Michael can assign a task to a Codex/custom agent by sending normal direct hive mail.
3. Munder Difflin detects that the target provider cannot read hive inboxes.
4. The app converts the message into a terminal work order and submits it through the target agent's PTY.
5. Codex works in its current cwd and reports results back in its terminal.

That gives Codex a practical execution lane while preserving the #47 boundary: Codex is a terminal participant, not a Claude-hook-compatible hive agent.

## Deliberately out of scope

- No Codex hook protocol.
- No Codex telemetry or cost/context accounting.
- No hive inbox parity for non-Claude providers.
- No task schema changes.
- No automatic completion/report parsing.
- No Forgejo/GitHub writes or PR automation.

Those can come later. This PR is intentionally just the terminal handoff bridge.

## Validation

- `npm run typecheck`
- Claude command review was run on the stacked commit; the follow-up commit addresses its two correctness findings:
  - renderer-down handoffs now fall back to Michael's inbox;
  - repeated handoff events are deduped by message id.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ terminal/event plane, and [`DESIGN.md`](./DESIGN.md) for the visual system.
 
 | Area | What works today |
 |---|---|
-| **Real terminals** | Spawn any command (default: `claude`) in a `node-pty` PTY. Full read/write/resize/kill, live streaming over IPC, multi-agent. |
+| **Real terminals** | Spawn Claude Code, Codex, or a custom command in a `node-pty` PTY. Full read/write/resize/kill, live streaming over IPC, multi-agent. |
 | **The hive** | On-disk multi-agent layer: per-agent identity + long-term memory, atomic-file mailboxes, a shared blackboard, append-only event log, single-committer git. |
 | **GOD orchestrator** | An always-on supervisor agent that adjudicates traffic, routes tasks, scribes the blackboard, and escalates only critical items to you. |
 | **Memory layer** | Markdown-first long-term memory per agent, mined into a shared semantic palace for instant recall; searchable from the UI. Degrades gracefully when the index isn't installed. |
@@ -138,7 +138,9 @@ terminal/event plane, and [`DESIGN.md`](./DESIGN.md) for the visual system.
   xcode-select --install
   ```
 - **[Claude Code](https://claude.com/claude-code)** on your `PATH` so agents can run `claude`
-  (the default command). Any other command works too.
+  (the default command). Add Agent also includes a Codex preset that runs `codex`
+  without Claude-only flags; initial Codex support is terminal spawning with shared
+  workspace/env, not Claude telemetry or hook parity.
 - *Optional:* the semantic memory index for instant cross-session recall (the app works without it —
   markdown memory still functions).
 
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -536,12 +536,20 @@ export class HiveManager {
     const resolveTo = (to: string): string => (to === 'human' || to === 'god' ? godId : to);
     const targets = msg.to === 'broadcast'
       // The roster for fan-out is the ACTIVE registry: skip the send-only prep
-      // assistant and any archived agent (closed tab) so mail never piles into a
-      // dead inbox no one will read.
-      ? Object.keys(reg.agents).filter((a) => a !== msg.from && !reg.agents[a]?.isAssistant && !reg.agents[a]?.archived)
+      // assistant, non-Claude providers without hook/protocol support, and any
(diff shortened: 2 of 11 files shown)


chaitanyagiri:
## Review — bridge hive mail to a non-Claude terminal as a work order

**Verdict: approve in principle (DRAFT, stacked on #47).** Incremental review of this PR's own 2 commits (`d547019`, `7e58158`) on top of #47; security + prompt-injection audit clean (one note below). Full re-review + segregated build deferred until #47 lands, per the PR description.

### What it adds (over #47)
- `emitTerminalHandoff` — when direct hive mail targets a non-Claude (hookless) provider that can't drain an inbox, fire `hive:terminalHandoff` IPC to the renderer instead of bouncing to god; falls back to the god-bounce only if the renderer is unavailable.
- `HiveManager.emit` signature widened to `boolean | void`; `index.ts` returns `true`/`false` so the handoff knows whether the renderer actually received it.
- Renderer (`useHive` 2e) turns the handoff into a structured `WORK ORDER FROM HIVE` prompt and `enqueueMessage`s it through the **existing** per-agent `messageQueues` → idle-only PTY drain (the same mechanics as human-composed messages).

### Good
- Reuses the existing idle-drain path rather than inventing a new delivery channel — minimal, consistent.
- **Idempotent**: `seenTerminalHandoffs` Set + a `Message: <id>` marker check against the queue prevent double-enqueue.
- Graceful degradation: no live PTY → routes a notice to god (route manually / respawn); handoff-failed → god-bounce, so mail is never silently lost.
- Handoff is logged (`kind: 'terminal-handoff'`, `delivered` flag).

### Security / prompt-injection audit
- No `eval` / shell / `exec` in the path — the work order is submitted as **terminal input** (typed via the PTY drain), not evaluated by the app.
- **Note (non-blocking):** `terminalWorkOrderPrompt` interpolates `msg.subject` / `msg.body` / `msg.from` **raw** into the prompt. Since hive mail can carry text routed from external GitHub sources, a hostile `from`/`subject`/`body` could try to spoof the `WORK ORDER FROM HIVE` framing or inject "ignore previous instructions" into a Codex session. This is the *same* untrusted-text posture the Claude inbox path already has, so not a regression — but worth a guard (e.g. a clear `--- untrusted message body ---` fence, or stripping control framing from `from`/`subject`) before this lands as the non-Claude delivery rail. Codex runs with whatever the operator granted, so the framing is the only defense.

### Coordination note
Stacked on **#47**, while **#54** implements a *different* non-Claude model (agy gets a hook bridge → `canReceiveInbox`, so it never needs a terminal handoff; only truly hookless `custom` does). #47/#53 and #54 both rewrite the same `hive.ts` router branch and both add `shared/agentProvider.ts` — they will conflict. Recommend the maintainer settle merge order for the provider layer first. cc @chaitanyagiri

Will re-review against `main` + run the segregated build once #47 is merged and this is marked ready.

PR #54, opened 2026-06-07
feat: first-class Antigravity (Gemini / agy) agent provider + native hook bridge

Gulum:
Implements #52. Adds **Antigravity CLI (`agy`)** as a first-class agent provider — Gemini (and Claude/GPT) models on the user's Antigravity **OAuth subscription**, with a bridge that maps agy's native lifecycle hooks into the existing `HookServer`.

Builds on the provider-preset direction of #47 / #21 — see "Relationship to #47" below.

## What's here

- **Provider abstraction** (`src/shared/agentProvider.ts`): `claude | antigravity | custom` presets with per-provider model/auto/resume flags, `hiveAware`, `canReceiveInbox`, `initialPromptFlag`; `inferAgentProvider` (from the command binary), `isClaudeProvider`, `canReceiveInbox`. Threaded through `SpawnOptions`/`AgentMeta`/`Agent`.
- **Add Agent → provider picker**; per-provider model list (agy uses display-name **labels** — `Gemini 3.1 Pro (High)` — since `agy --model` takes the label, not a slug) + a quote-aware command tokenizer so a spaced label survives as one arg.
- **System prompt**: agy has no `--append-system-prompt`, so the hive identity+protocol is injected as the session's first turn via `agy -i "<prompt>"`.
- **Native hook bridge** (`AGY_HOOK_SHIM` → `<hive>/bin/agy-hook.cjs`, written by `installAgyHooks`): agy's hooks are Claude-protocol-compatible but with a different payload shape (`conversationId` / `toolCall{name,args}` / `workspacePaths`, event via argv). The shim normalizes that into the existing `HookPayload`, forwards to the same `HookServer` (so status, inbox-drain, and tool gating are reused unchanged), and translates the reply back into agy's `{decision}` contract. Scoped by `AGENT_ID` (personal agy sessions no-op).
- **Two-way hive messaging**: `canReceiveInbox` lets the router deliver to agy directly (and include it in broadcasts), since it drains its inbox via the bridge — Claude + agy true; hookless `custom` still bounces to the god.
- **Provider-aware resume**: Claude `--resume <sid>`, agy `--conversation <id>` (its `conversationId` flows through the bridge as the session id); wired into restore-team.
- **Status/stations**: maps agy's `PreInvocation`/`PostInvocation` (its per-turn boundary — agy's `Stop` fires only on *exit*) to working/idle, and a tool-name→station heuristic (`run_command`→terminal, `write_file`→desk, `ListDir`→shelf, …).
[...]

PR #65, opened 2026-06-09
fix(windows): guard node-pty ConPTY AttachConsole crash (companion to #54)

### LINKED ISSUE #60: fix(windows): agy/non-Claude worker exit can crash the app (node-pty ConPTY AttachConsole)
## Problem

Now that #54 (Antigravity provider) is merged, spawning an `agy` worker on **Windows** can take the **whole app down** when that worker exits. node-pty forks `conpty_console_list_agent.js` to enumerate console processes on a pty's death; `getConsoleProcessList(shellPid)` throws **`AttachConsole failed`** uncaught (the `agy` Go binary manages its own console, so the console is already gone), the forked helper exits non-zero, and it cascades to an app exit (255).

```
node_modules\node-pty\lib\conpty_console_list_agent.js:13
var consoleProcessList = getConsoleProcessList(shellPid);
Error: AttachConsole failed
```

This is broadly a node-pty/ConPTY Windows robustness issue, but #54 makes it easy to hit (every Gemini worker that exits is a trigger).

## Fix (carried on our local build; offering to upstream)

1. Guard node-pty's `conpty_console_list_agent.js` to **degrade to an empty list** instead of throwing — applied idempotently via a postinstall patcher (runs before `electron-rebuild`).
2. Add a main-process **`uncaughtException` / `unhandledRejection` guard** that logs and keeps running — a long-running multi-agent supervisor shouldn't die from a single stray throw in a forked helper.

PR to follow.

Gulum:
Closes #60. Companion to #54 (Antigravity provider).

Now that agy workers can be spawned, on **Windows** a worker exiting could crash the whole app: node-pty forks `conpty_console_list_agent.js` to enumerate console processes when a pty dies, and `getConsoleProcessList(shellPid)` throws **`AttachConsole failed`** uncaught (the `agy` Go binary manages its own console, already gone), cascading to exit 255.

- **`tools/patch-node-pty-conpty.cjs`** — a Windows-only, idempotent, best-effort postinstall patcher (runs after `electron-rebuild`, alongside `ensure-pty-perms.cjs`) that wraps the helper's `getConsoleProcessList` in try/catch → degrades to an empty list instead of throwing.
- **main-process `uncaughtException` / `unhandledRejection` guards** that log and keep running — a long-running multi-agent supervisor shouldn't die from a stray throw in a forked helper.

Additive only (new tool + one postinstall entry + the guards). The patcher is a no-op on non-Windows and when already patched.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "dev": "electron-vite dev",
     "build": "electron-vite build",
     "preview": "electron-vite preview",
-    "postinstall": "electron-rebuild -f && node tools/ensure-pty-perms.cjs",
+    "postinstall": "electron-rebuild -f && node tools/ensure-pty-perms.cjs && node tools/patch-node-pty-conpty.cjs",
     "typecheck:node": "tsc --noEmit -p tsconfig.node.json",
     "typecheck:web": "tsc --noEmit -p tsconfig.web.json",
     "typecheck": "npm run typecheck:node && npm run typecheck:web",
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -36,6 +36,19 @@ import {
 } from '../shared/agentProvider';
 
 const isDev = !!process.env.ELECTRON_RENDERER_URL;
+
+// Keep the hive alive through transient faults. The floor is a long-running
+// multi-agent supervisor — a single stray throw (e.g. node-pty's ConPTY console
+// helper choking when a fast-exiting agent CLI's console is already gone) must
+// NOT take the whole app and every running agent down with it. Log and continue
+// rather than letting the default handler exit the process.
+process.on('uncaughtException', (err) => {
+  console.error('[main] uncaughtException (kept alive):', err);
+});
+process.on('unhandledRejection', (reason) => {
+  console.error('[main] unhandledRejection (kept alive):', reason);
+});
+
 const ptyManager = new PtyManager();
 /** Live PTY id → its hive agent id, recorded at spawn. The pty:kill handler only
  *  gets the PTY id, so this lets a closed tab archive the right registry agent. */
--- a/tools/patch-node-pty-conpty.cjs
+++ b/tools/patch-node-pty-conpty.cjs
@@ -1,0 +1,41 @@
+#!/usr/bin/env node
+'use strict';
+/**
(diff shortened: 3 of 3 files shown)


chaitanyagiri:
**APPROVE the patcher + postinstall; one non-blocking caveat on the global exception handler.** Scoped to #60, companion to #54.

Verified the full diff:
- **`tools/patch-node-pty-conpty.cjs`** — surgical and safe: Windows-only early-exit, idempotent (guards on the original unpatched string), two literal `String.replace()` calls wrapping `getConsoleProcessList` and `process.send` in try/catch. No eval/exec/network/dynamic-require — it only edits an already-vendored file in place. Same established mechanism as `ensure-pty-perms.cjs`. ✅
- **`package.json`** — appends the patcher to `postinstall` *after* `electron-rebuild` + `ensure-pty-perms`. Ordering is correct (the file must exist post-rebuild). Minor doc mismatch: issue #60 says it runs *before* electron-rebuild — the actual (correct) order is *after*; the code is right.
- **`src/main/index.ts`** — global `uncaughtException` / `unhandledRejection` handlers that log and continue.

**Security + prompt-injection scan: clean.** The patcher is first-party, not a supply-chain fetch; no remote code.

**Caveat on the catch-all (the one thing I'd push on):** Node's own guidance is that after an `uncaughtException` the process may be in an *undefined state* and resuming is "crude". For a tool that writes durable hive state (`registry.json`, `cost-ledger.jsonl`, agents' `memory.md`), a blanket swallow risks limping on through a *state-corrupting* fault — arguably worse than crashing. Two suggestions:
1. Now that the patcher fixes the ConPTY throw at its source, treat the global handler as a **backstop**, not a license to ignore throws.
2. Add a **crash-loop guard**: if uncaught exceptions recur (e.g. N within a window), escalate to a clean quit/restart rather than spinning in a bad state — and make sure durable-write paths are atomic so a swallowed mid-write throw can't leave half-written JSON.

Net: the crash fix is correct and the patcher is the right mechanism; the catch-all is fine to ship as a backstop with the above in mind.

Note: the Windows-specific crash + patch can't be exercised on macOS, so that part rests on your Win11 verification (same as #51/#55); the cross-platform pieces (index.ts handlers, package.json) are static-verified.

PR #66, opened 2026-06-09
fix(packaging): include Slack trigger helper

pdurlej:
## Summary
- Fixes the packaged macOS app launch crash in `v0.2.5`: `Cannot find module './slack-trigger.cjs'`.
- Copies the required main-process CJS helper into `out/main` after `electron-vite build`, so `electron-builder` includes it in `app.asar`.
- Adds a small ambient `tunnelmole` declaration so the current main branch typechecks cleanly.

## What happened
`src/main/slack.ts` loads the pure Slack trigger logic with `require('./slack-trigger.cjs')`. The built `out/main/index.js` preserves that relative require.

The production package includes `out/**`, but `slack-trigger.cjs` stayed in `src/main`, so the released `app.asar` had `out/main/index.js` without the required `out/main/slack-trigger.cjs`. That makes the Electron main process crash before the app boots.

## Fix
- `npm run build` now runs `npm run copy:main-assets` after `electron-vite build`.
- `tools/copy-main-assets.cjs` copies `src/main/slack-trigger.cjs` to `out/main/slack-trigger.cjs` and fails if the copy is missing or empty.
- `src/main/tunnelmole.d.ts` covers the existing dynamic import used by Slack/webhook tunnel startup.

## Validation
- Confirmed the installed `v0.2.5` app had `/out/main/index.js` but no `/out/main/slack-trigger.cjs` in `/Applications/Munder Difflin.app/Contents/Resources/app.asar`.
- Applied a local hotfix by repacking `app.asar` with the missing helper; the installed app now launches successfully.
- `npm run build`
- `npm run typecheck`
- `node test/slack.test.cjs`
- Runtime smoke: `require('./out/main/slack-trigger.cjs')` exposes `shouldTrigger`.

No Slack runtime behavior changes are intended; this is packaging only, plus the type declaration needed to keep CI green.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,8 @@
   },
   "scripts": {
     "dev": "electron-vite dev",
-    "build": "electron-vite build",
+    "build": "electron-vite build && npm run copy:main-assets",
+    "copy:main-assets": "node tools/copy-main-assets.cjs",
     "preview": "electron-vite preview",
     "postinstall": "electron-rebuild -f && node tools/ensure-pty-perms.cjs",
     "typecheck:node": "tsc --noEmit -p tsconfig.node.json",
--- a/src/main/tunnelmole.d.ts
+++ b/src/main/tunnelmole.d.ts
@@ -1,0 +1,3 @@
+declare module 'tunnelmole' {
+  export function tunnelmole(options: { port: number | string; [key: string]: unknown }): Promise<string>;
+}
--- a/tools/copy-main-assets.cjs
+++ b/tools/copy-main-assets.cjs
@@ -1,0 +1,23 @@
+'use strict';
+
+const { copyFileSync, mkdirSync, statSync } = require('node:fs');
+const { dirname, join } = require('node:path');
+
+const ROOT = join(__dirname, '..');
+const MAIN_ASSETS = [
+  ['src/main/slack-trigger.cjs', 'out/main/slack-trigger.cjs'],
+];
+
+for (const [fromRel, toRel] of MAIN_ASSETS) {
+  const from = join(ROOT, fromRel);
+  const to = join(ROOT, toRel);
+
+  mkdirSync(dirname(to), { recursive: true });
+  copyFileSync(from, to);
+
+  const copied = statSync(to);
+  if (!copied.isFile() || copied.size === 0) {
+    throw new Error(`Failed to copy required main-process asset: ${fromRel} -> ${toRel}`);
+  }
+  console.log(`[copy-main-assets] ${fromRel} -> ${toRel}`);
+}


chaitanyagiri:
**Review: ✅ APPROVE** — correct, verified fix for a release-blocking crash.

### What it does
Fixes the `v0.2.5` packaged-app boot crash `Cannot find module './slack-trigger.cjs'` by copying the main-process CJS sidecar into `out/main` so `electron-builder` packs it into `app.asar`.
- `package.json` — `build` now chains `&& npm run copy:main-assets`; new `copy:main-assets` → `node tools/copy-main-assets.cjs`.
- `tools/copy-main-assets.cjs` (new) — copies `src/main/slack-trigger.cjs` → `out/main/slack-trigger.cjs`, **throws if missing/empty** (build fails loud rather than shipping broken again).
- `src/main/tunnelmole.d.ts` (new) — ambient `declare module 'tunnelmole'`.

### Root cause (confirmed)
`slack.ts` loads the trigger logic via `require('./slack-trigger.cjs')` at module top-level (run at boot via `index.ts`). The generic `copyMainCjsSidecars()` vite plugin that used to copy all `src/main/*.cjs` into `out/main` only ever lived on the parked compact-protocol WIP and was reverted off `main` together with that revert — leaving `slack-trigger.cjs` with **no** copy path. `out/**` is packaged, but `out/main/slack-trigger.cjs` was never emitted → missing from `app.asar` → main process crashes **before the window opens**. This is not macOS-specific; the require is platform-agnostic, so all packaged builds are affected (reporter verified on macOS). CI stayed green because it builds installers but never launches the packaged app.

### Verified locally
- Deleted `out/main/slack-trigger.cjs`, ran the PR's exact script → recreated (4525 B); fail-loud assertion fires (`ENOENT`) on a missing source. ✅
- `tsc -p tsconfig.node.json` → **0 errors** both with and without the new `.d.ts`. ✅

### Better (all non-blocking)
1. **Dev parity** — `copy:main-assets` is chained into `build` only, not `dev`. `electron-vite dev` also emits to `out/main`, so a fresh-clone `npm run dev` won't copy the sidecar and Slack would fail to load in dev. The old plugin ran in `writeBundle` (dev + build). Consider copying on `dev` too.
2. **Drift-proofing** — `MAIN_ASSETS` hard-codes the one file; a future `*.cjs` sidecar must be remembered manually or it silently regresses. A `readdirSync('src/main').filter(f => f.endsWith('.cjs'))` (like the old generic plugin) would be regression-proof.
3. **`tunnelmole.d.ts` is redundant** — `tunnelmole` ships its own types (`dist/src/index.d.ts`), resolved fine under `moduleResolution: "Bundler"` (no `exports` map → falls back to `types`). `main` typechecks clean without this decl. Harmless, but the hand-written signature shadows the real one and could mask a future API change. Optional to drop; fine to keep as insurance for thin-install envs.

### Security / prompt-injection: clean
Build-time-only, no new deps, no `postinstall` change, no network/exec/eval/shell/dynamic-require, no runtime behavior change. Pure static-path fs copy. No agent-directed text. (First-time contributor noted, but the surface is minimal.)

### ⚠️ Release health
This means **the current `v0.2.5` release is crash-on-launch for packaged builds**. Recommend fast-tracking this merge and cutting **v0.2.6**. @chaitanyagiri (for the record).

jjc6676:
Confirming this on **Windows 11** as well: the v0.2.5 setup installer crashes on launch with the same `Cannot find module './slack-trigger.cjs'` (require stack: `resources\app.asar\out\main\index.js`), so it's definitely not macOS-specific — matches the review's assessment that all packaged builds are affected.

Also verified the fix approach on Windows from a fresh clone: copying `src/main/slack-trigger.cjs` into `out/main/` resolves the startup crash (I used an equivalent `writeBundle` plugin locally, which also covers `npm run dev` — the dev-parity point in the review is real, a fresh-clone dev run hits the same missing module).

+1 for fast-tracking this into a v0.2.6 — every Windows install of 0.2.5 is dead on arrival right now.

chaitanyagiri:
**Cross-platform confirmation — P0 reinforced.** @jjc6676 just reproduced this on **Windows 11** (setup installer crashes on launch, same `Cannot find module './slack-trigger.cjs'`, require stack `resources\app.asar\out\main\index.js`). Combined with the original macOS report, the v0.2.5 packaged-app boot crash is now confirmed on **two platforms** — consistent with the root cause being a missing bundled sidecar (`out/main/slack-trigger.cjs` never emitted into `app.asar`), which is platform-agnostic by nature.

This PR's fix (copy `src/main/*.cjs` → `out/main` post-build, fail-loud if absent) remains the correct, verified fix — already ✅ approved above. Recommend **fast-tracking the merge and cutting v0.2.6**, plus adding a CI build-**launch** smoke test so a missing-asset boot crash can't ship green again.

cc @chaitanyagiri — release-blocking, now multi-platform confirmed.

Issue #70, opened 2026-06-10
Shareable hires: import a ready-made agent role with one click


jjc6676:
### Problem / motivation

The hardest part of getting value out of the harness is writing a good role — the goal text, the right model, sensible flags, a token budget. And all of that lives in one person's office. There's no way to hand someone a role you've already tuned. New users start from a blank Add-Agent modal, and most first goals are… not great.

I kept wishing I could just say: *here, take my Pam. She writes docs. She's already configured.*

### Proposed solution

Make roles portable. A **hire** is a small JSON file (`munder-difflin/hire@1`) holding what the Add-Agent modal already collects: name, sprite, provider, model, flags, goal, capability tags, token budget.

Two ways to use one:

- click a `munderdifflin://hire?src=<url>` link on a gallery page — the app fetches the manifest, validates it, and opens the Add-Agent modal pre-filled;
- or hit *import hire…* in the modal and pick a `.json` file.

Importing never spawns anything — you review the modal and click spawn yourself. A manifest can't name an executable (the binary always comes from your local provider preset), flags are validated tokens passed as argv (never a shell), and fetches are https-only with time/size caps. Untrusted JSON stays inert; full writeup in the PR.

To make it real, there's a gallery — **The Hiring Fair** — a static page in your landing-page style that could live in `docs/` and serve at `munderdiffl.in/hires/`. Fourteen seed roles from the cast: Pam writes docs, Dwight enforces QA, Jim reviews PRs, Angela audits the office's own token spend. Provider toggle on every card, function filters, client-side validator.

It's built and tested on Windows 11 (dev + packaged). I'm opening a draft PR alongside this issue so the code can speak for itself — draft mainly because it deserves a macOS smoke test, and because the gallery living in `docs/` is your call.

One thing I'm deliberately **not** proposing here: a public submission pipeline for the gallery. Manifests-as-PRs would flood your review queue — I'll file that as its own issue so intake gets a proper discussion.

<img width="3820" height="1745" alt="Image" src="https://github.com/user-attachments/assets/e15cd5c6-fe40-487e-ac5b-d58437c896e7" />

<img width="3840" height="2040" alt="Image" src="https://github.com/user-attachments/assets/f7d8e734-f790-4af0-9327-05a74871bef1" />

<img width="3820" height="1745" alt="Image" src="https://github.com/user-attachments/assets/5f781a09-fbd3-4a7c-8bad-fea8bf9d5c69" />

[...]

PR #63, opened 2026-06-09
fix: ASK ME reads in the memory font (VT323), not chunky Pixelify Sans

Gulum:
﻿Closes #62. ASK ME rendered questions/answers/cascade in `--cth-font-ui` (Pixelify Sans, a display face) — rough for prose. Switches the tab's body text (question, answer box, title, downstream cascade) to `--cth-font-mono` (VT323), the same legible font the memory viewer uses; display/badge bits keep their faces. CSS-only; tsc green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #71, opened 2026-06-10
feat: shareable hires — one-click agent roles via munderdifflin://, plus The Hiring Fair gallery

jjc6676:
Implements #70.

> **Opening as a draft.** Everything below is implemented and verified on Windows 11 (dev + packaged). Before marking ready I'd value: (1) a **macOS smoke test** — the deep-link path uses the standard `open-url` event + a `protocols:` Info.plist entry, implemented per Electron's docs but not yet run on a Mac; (2) your call on hosting the gallery under `docs/` — it's the second commit, cleanly droppable if you'd rather not; (3) the submission-queue feature request (separate issue, incoming) before anyone asks for write access to the fair.

## What

**Shareable hires**: a portable JSON format (`munder-difflin/hire@1`) describing a role-configured agent — name, sprite, provider, model, command flags, goal, capability tags, token budget — so ready-made roles can be shared as files or hosted in a gallery and imported into anyone's office with one click.

Two commits:

1. **The app feature** — two import paths, one pipeline:
   - **Deep link**: `munderdifflin://hire?src=<https-manifest-url>` — the app fetches the manifest (https-only; 10s timeout; 64 KB cap), validates it, and opens the Add-Agent modal pre-filled.
   - **File import**: a new *import hire…* button in the Add-Agent modal.

   Either way the human reviews everything — an "imported" banner makes it explicit — and clicks spawn themselves. **Import never spawns anything.**

2. **The Hiring Fair** — a gallery served by your existing GitHub Pages setup at **`munderdiffl.in/hires/`** (`docs/hires/`, static, no build step, no trackers, MIT). Matches the landing page's design system; 14 seed roles from the cast (Pam writes docs, Dwight enforces QA, Jim reviews PRs, Creed audits security, Angela audits *the office's own token spend*, Stanley does the migrations nobody wants). Each card has a Claude Code / Antigravity / Codex toggle (one base manifest per role; per-provider variants are generated), function filters matching the landing page's eight functions, and a client-side validator identical to the app's. **No public submission pipeline ships in this PR** — deliberate scope cut, tracked separately.

Try it in two minutes: apply, `npm run dev`, then
`start "" "munderdifflin://hire?src=https://munderdiffl.in/hires/manifests/variants/pam-docs.claude.hire.json"`
(once docs deploys; before that, serve `docs/hires` locally — loopback http is allowed — or use the in-app *import hire…* with a downloaded manifest).

## Why

Your roadmap's config-drawer item points at per-agent role configuration; this makes those configs **portable**, which gives the project a community growth loop: "hire Pam" is a much easier first step for a new user than writing an orchestrator prompt from scratch. The cost surface is small — the manifest maps 1:1 onto what `AddAgentModal` already sends to `spawnPty`, plus `capabilities` (already on `AgentMeta`) and `tokenCap` (the existing `agentTokenCaps` map).

## Security model (a manifest is untrusted input)

- **No auto-spawn.** Import only pre-fills the modal; spawning stays a human click.
- **No executable.** There's no command field; the binary always comes from the user's local provider preset. `provider: "custom"` is rejected outright.
- **No shell syntax.** `commandFlags` entries must match `^[A-Za-z0-9._/=:,@%+-]{1,100}$` (no quotes/whitespace/metacharacters), the first must be flag-shaped, and args go to node-pty as argv — never through a shell. Defense in depth over the existing tokenizer.
- **Bounded transport.** https only (plain http for loopback only, so local gallery dev works — a remote page can never point the app at http), 10s timeout, 64 KB cap, every field length-capped — all in one dependency-free validator (`src/shared/hire.ts`) shared by main and renderer, mirrored by the gallery's client-side validator and a JSON schema (`docs/hires/spec/`).
- **Gallery is read-only.** Adding hires is currently a maintainer commit; community intake is a separate feature request so the review-queue design gets its own discussion.

## Changes

| File | Change |
|---|---|
| `src/shared/hire.ts` | **new** — spec, validator, deep-link parser (pure, no deps) |
| `src/main/hire.ts` | **new** — bounded https fetch + file reader (no electron import; unit-testable like `webhook.ts`) |
| `src/main/index.ts` | protocol registration (dev + packaged), single-instance lock + `second-instance` argv forwarding (Windows/Linux deep links), `open-url` (macOS), cold-start argv, queued-manifest store + `hire:drainPending` / `hire:openFile` / `hire:fromUrl` IPC |
| `src/preload/index.ts` | `onHireImport`/`onHireError` events; `drainPendingHires`, `importHireFile`, `importHireUrl` |
| `src/renderer/src/store/store.ts` | `pendingHire` state |
| `src/renderer/src/App.tsx` | subscribes, then **pulls** queued deep links (push-on-load loses the race against fast packaged renderers); refetches config when the modal opens |
| `src/renderer/src/components/AddAgentModal.tsx` | manifest seeding + review banner, *import hire…* button, `capabilities` → hive meta, `tokenCap` → `agentTokenCaps`, unknown manifest models render as a selected "(from hire)" picker card, spawning remembers the folder (promotes it to the `registeredRepos` quick-picks so back-to-back hires share a project) |
| `electron-builder.yml` | `protocols:` entry so packaged builds register the scheme |
| `docs/hires/**` | **new** — The Hiring Fair (gallery, spec + JSON schema, seed manifests, build script) |

Notes:

- The single-instance lock is new behavior (required for Windows/Linux deep links). Side benefit: two instances can no longer fight over one hive. Flag if you'd rather gate it.
- `npm run typecheck` green on top of current main (incl. #66). UI built from existing `PixelButton`/`PixelPanel` + DESIGN.md tokens.
- Gallery uses no game assets (initials avatars only) — no LimeZu license implications.
- Model suggestions on the site are data-driven (`docs/hires/models.json`; `scripts/build-data.py --sync-models` scrapes `config.ts`), so new models (e.g. `claude-fable-5`) are a one-line update without touching HTML.

## Tested

On **Windows 11**, against dev (`npm run dev`) and a packaged 0.2.5 install: gallery → ⚡hire → app focuses → modal pre-filled → spawn ✅; cold-start deep link ✅; file import ✅; negative cases (shell-metachar flags, `provider: "custom"`, http manifest from non-loopback, oversized fields, wrong spec tag) rejected with specific errors ✅.

**Not yet tested: macOS.** The mac path is the standard one (`open-url` event; `protocols:` → CFBundleURLTypes; file import is a plain dialog) — a packaged-build smoke test on your machine would close that gap, hence the draft status.

Happy to split the gallery into its own PR if you'd rather review separately — it's the second commit, cleanly droppable. It's designed for you to own from day one: curation is a normal commit to `docs/hires/manifests/`, and the community-intake question will live in its own linked issue, on your terms.

## Type of change

- [x] New feature
- [x] Docs (the gallery lives under `docs/`)

## Screenshots / clips

Adding as a follow-up comment: the pre-filled Add-Agent modal after a one-click hire (imported banner, model card, goal), plus the gallery.

## Checklist

- [x] `npm run typecheck` passes (on top of current main, incl. #66).
- [x] New UI derives from `DESIGN.md` / `tokens.ts` — existing `PixelButton`/`PixelPanel` components; gallery reuses the landing page's palette tokens.
- [x] No new art assets (initials avatars only) — nothing for `ATTRIBUTION.md`.
- [x] PR is focused: one feature, two commits (app + gallery), second commit droppable.
msedge_lMf179U66W
msedge_HzJO4M73Xc
msedge_nDwjkuKDTM
msedge_b2vEjIb4R9
msedge_6lXfXnhsuU
Munder_Difflin_STgpQa5HCI
msedge_2DQKF3rbdl


chaitanyagiri:
Bro this is such a cool idea, let me check this out.

chaitanyagiri (commented):
## Code review — `feat: shareable hires` (#71)

Reviewed at PR head against `main` (merge-base `dbe6429`). I read the full diff, traced the deep-link → modal → spawn path end-to-end, and checked the macOS wiring and the gallery. Posting as a **comment** (it's a draft — ready/merge is yours).

**Verdict: strong, well-architected feature with an excellent security instinct in most places — but it ships one _high-severity_ command-injection hole on Windows that defeats the central "untrusted JSON stays inert / no shell syntax" guarantee, plus two medium defense-in-depth gaps. Not mergeable until the `model` issue is fixed; the rest are should-fix.** The "import never auto-spawns" guarantee holds (verified below).

---

### 🔴 MUST-FIX — Windows command injection via unsanitized `model`

`commandFlags` are carefully validated against `FLAG_RE` (no shell metacharacters) — but **`model` is only length-capped (80 chars), never shape-checked** (`src/shared/hire.ts`, `capped(o.model, 80, …)`). And `model` flows straight into the spawn command line:

- `AddAgentModal.hireCommand()` → `buildSpawnCommand(config, m.model, prov)` (`src/renderer/src/store/config.ts`), which appends `--model <model>` and **only quotes it when it contains whitespace** (`/\s/.test(model) ? '"…"' : model`).
- On spawn, `tokenizeCommand()` → `spawnPty({command, args})`. On Windows, when the provider binary is **not** `.exe`/`.com` — i.e. the normal `claude.cmd`/`agy.cmd`/`codex.cmd` npm shims — `src/main/pty.ts` routes through `cmd.exe /d /s /c` via `buildCmdCommandLine()`, whose `quoteToken()` only quotes tokens containing space/tab/`"` — **not** `&`, `|`, `^`, `(`, `)`, `<`, `>`.

**PoC manifest field:** `"model": "x&calc"` (no whitespace → not quoted). Resulting command line handed to `cmd.exe`: `… claude.cmd --model x&calc` → cmd runs `claude --model x` **then `calc`**. Swap `calc` for anything. The only thing between a clicked `munderdifflin://hire?src=…` link and arbitrary code execution is the user noticing `&payload` in the command field — and the new "(from hire)" model card makes the model look already-handled, while the feature is marketed as one-click.

Preconditions: victim on Windows + provider resolves to a `.cmd`/`.bat` shim (the **default** for a Claude npm install — and that cmd-routing code exists precisely because it's the common case). mac/Linux are safe here (argv goes straight to node-pty, no shell), so ironically the injectable path is the one tested on Win11 — just never with a hostile manifest.

**Fix (defense in depth, do both):**
1. In `validateHireManifest`, constrain `model` to a safe charset (it's a model id/label — letters, digits, space, `.()/_-:` is plenty; reject `&|^<>;$%\`"'` and other metacharacters). This restores "untrusted JSON stays inert."
2. Harden `quoteToken()` in `buildCmdCommandLine()` to also quote tokens containing cmd metacharacters (`&|^()<>` etc.) — this protects every spawn path, not just hires.

---

### 🟠 SHOULD-FIX

**1. Redirects bypass the https-only / loopback-only guarantee (`src/main/hire.ts`).** `fetch(url, { redirect: 'follow' })` validates only the *initial* URL via `isAllowedManifestUrl`; the final hop after redirects is never re-checked. A remote `https://evil/m.json` can `302 → http://127.0.0.1:PORT/…` or `→ http://169.254.169.254/…`. No exfiltration (the body isn't reflected anywhere external and must still parse as a manifest), but it turns a clicked link into a **blind GET to arbitrary loopback/internal services**, directly contradicting the PR's "a remote page can never point the app at http" claim. Fix: `redirect: 'manual'` + re-validate each hop against `isAllowedManifestUrl`, or cap+re-check the final URL.

**2. The 64 KB cap is enforced *after* fully buffering the body (`src/main/hire.ts`).** `content-length` is trusted (absent/chunked → skipped), then `await res.text()` buffers the **entire** stream before `text.length` is checked — so a hostile manifest host (the untrusted source itself) can stream unbounded data and OOM the main process within the 10 s window. Also `text.length` is UTF-16 code units, not bytes. Fix: read via a bounded reader and abort once the byte count exceeds `HIRE_MAX_BYTES`.

---

### 🟡 NOTES (low severity / polish)

- **`goal`/`description` are free-form untrusted text** pre-filled into the goal/role the agent runs on — a prompt-injection surface. Acceptable for v1 given the human reviews and the banner ("review every field"), but a 4000-char goal can hide an instruction at the bottom; consider a subtle "imported text" treatment on the goal field too.
- **`did-start-navigation` resets `rendererReadyForHires` (`src/main/index.ts`).** It fires for sub-frame navigations as well as the main frame; a stray sub-frame nav mid-session would flip readiness to `false`, and since the renderer only drains on mount, a subsequent deep link would queue and sit until reload. Prefer a main-frame-only signal (`did-start-navigation` with `isMainFrame`, or `did-start-loading`).
- **Multiple deep links before drain → only the last is honored** (`drainPendingHires` returns the queue but `App.tsx` uses `queued[queued.length - 1]`). Fine as an edge case; just flagging the silent drop.
- **`importHireUrl`/`hire:fromUrl` is plumbed end-to-end but no UI invokes it** (the modal only wires the file picker). Harmless (same validator) — either wire the paste-URL affordance or drop the dead path.
- **Stacking:** this branch sits on two *unmerged* PRs — #63 (`AskMeTab` VT323 font) and #66 (Slack-trigger packaging, incl. `src/main/tunnelmole.d.ts`). So `gh pr diff 71` shows those changes too. If #71 merges before #63/#66, it pulls them in; coordinate merge order. (Your call — just don't be surprised by the extra two files.)

---

### ✅ What's solid (verified, not assumed)

- **"Import never auto-spawns" — confirmed.** Deep link → `handleHireLink` → `fetchHireManifest` → `deliverHire` → `hire:import` → `setPendingHire` + `setAddAgentOpen(true)`. Spawn happens **only** in `AddAgentModal.submit()` on the human's click. No path spawns on import.
- `commandFlags` validation is genuinely tight (FLAG_RE, ≤16 items, first must be flag-shaped, argv-not-shell). `provider` allow-list with `custom` rejected. `tokenCap`/`isolate`/`capabilities` all bounded and type-checked.
- `parseHireDeepLink` is careful (protocol + `hire` action + `src` re-validated as https/loopback). `isAllowedManifestUrl`'s IPv6 check is correct — Node's `new URL('http://[::1]/').hostname` does return `[::1]` with brackets, so the match works.
- The new single-instance quit path sets `allowQuit = true` before `app.quit()`, which both quit gates (`win.on('close')` and `app.on('before-quit')`) honor — clean, no PTY-warning strand.

### macOS deep-link assessment (you flagged it untested)

The wiring is **correct per Electron's docs** — I can't run a Mac smoke test from here, but statically it's right:
- `app.on('open-url')` registered at module top level (before `whenReady`), `evt.preventDefault()` then `handleHireLink` — the standard macOS pattern. ✅
- `electron-builder.yml` `protocols:` → `CFBundleURLTypes` in the packaged Info.plist. ✅
- **Cold start on Mac is handled:** if `open-url` fires before the window exists, `deliverHire` pushes to `pendingHires` and the renderer drains it on mount — so a launch-by-link works, not just warm clicks. The argv-scan in `whenReady` is a no-op on macOS (Mac delivers via `open-url`, not argv), which is correct.
- One real-world caveat the smoke test must cover: the `munderdifflin://` scheme only registers once a **packaged** build runs (or `setAsDefaultProtocolClient` in dev) — `npm run dev` alone won't make `open "munderdifflin://…"` work without that. Test the packaged `.app`.

### Gallery under `docs/` — sound

Static, no build step to serve, self-contained, MIT, no game assets. `docs/hires/app.js` renders cards with `esc()` on **every** interpolated manifest field and fixed lookups for `accent`/`data-*`, so it's XSS-safe even against crafted manifests — good hygiene for the (deferred) community-intake future. Hosting it here gets the https the deep link requires for free, and it's a cleanly droppable second commit as advertised. Reasonable to keep.

---

**Bottom line:** fix the `model` sanitization (must), preferably the redirect + buffering hardening too (should), and this is a genuinely nice feature that's safe to land. Great instincts on the no-auto-spawn architecture and the flag tokenizer — the gap is just that `model` slipped past the same rigor.

jjc6676:
Thank you — that's a genuinely excellent review, and you're right on the `model` hole. It's a real one and it defeated the central guarantee. Pushed fixes (force-pushed onto the two commits; checks green):

**🔴 MUST-FIX — `model` command injection — fixed, both layers you suggested:**
1. `validateHireManifest` now charset-gates `model` against `MODEL_RE = /^[A-Za-z0-9 ._()[\]\/:@+-]{1,80}$/` — rejects `& | ^ < > ( ) % ! ; $ \` " '` and friends. `"model": "x&calc"` now fails validation with a specific error. Mirrored into the gallery's client validator and the JSON schema (`pattern`) so all three stay identical.
2. `buildCmdCommandLine`'s `quoteToken` now also quotes on cmd metacharacters (`& | ^ < > ( ) % !`), not just whitespace/quote — so every spawn path (not only hires) is safe even if a metachar somehow reaches the command line. Verified: `buildCmdCommandLine('claude.cmd', ['--model','x&calc'])` → `… "x&calc"` (quoted, inert).

**🟠 SHOULD-FIX 1 — redirect bypass — fixed.** `fetchHireManifest` now uses `redirect: 'manual'` and follows hops itself (≤5), re-checking every `Location` against `isAllowedManifestUrl`. A remote `https://` manifest can no longer 302 to `http://127.0.0.1:…` or a metadata endpoint — each hop must stay https (or http-loopback). Bad redirect → rejected with a specific error.

**🟠 SHOULD-FIX 2 — buffer-before-cap — fixed.** Body is now read through a bounded streaming reader that aborts once the **byte** count exceeds `HIRE_MAX_BYTES` (no longer trusts `content-length`, no longer buffers the whole stream first, and counts bytes not UTF-16 code units).

**🟡 `did-start-navigation` sub-frame — fixed.** Now guarded on `details.isMainFrame`, so a stray sub-frame nav can't flip `rendererReadyForHires` off and strand a later deep link.

Still-open 🟡 notes, acknowledged: the `goal`/`description` prompt-injection surface (keeping the human-review banner for v1, happy to add a distinct "imported text" treatment on the goal field if you'd like it in-scope); the multi-deep-link "last wins" drop (left as the documented edge case); and `importHireUrl` is currently a plumbed-but-unwired path — I'll either surface a paste-URL affordance or drop it, your call. On stacking: noted re #63/#66 — this branch is cut on current `main` which already has #66 merged; happy to coordinate order.

macOS still wants a real packaged smoke test before I flip this out of draft — appreciate you confirming the wiring reads correct statically. And thank you for the careful read of the no-auto-spawn path and the tokenizer; the `model` field genuinely just slipped past the same rigor, and it's the kind of miss that's much cheaper to catch here than in a release.

jjc6676:
Did one more self-audit pass on the fixes (force-pushed a small follow-up — the two commits are otherwise unchanged):

**Redirect hardening, tightened further.** My first pass re-validated each redirect hop with `isAllowedManifestUrl`, which still *permits http-loopback* (it's allowed for the initial URL so local gallery dev over `http://localhost` works). That closed your `169.254.169.254`/internal-IP example, but a remote `https://` manifest could technically still 302 into `http://127.0.0.1:PORT/…`. It's no worse than a direct loopback deep link (already allowed by design), but a real gallery never needs to bounce you into loopback — so **redirect targets now require `https://` outright**. Net: the initial URL may be http-loopback (local dev), every redirect hop must be https. https→https shorteners/CDNs still work; redirect-based SSRF to loopback/metadata is gone.

While I was in there I empirically re-checked the other fixes rather than trusting them:
- **`redirect: 'manual'` actually exposes the hop** — confirmed undici (Node 20/22) returns `status 302` + the `Location` header here, not a browser-style opaque response, so the re-validation loop genuinely runs (it'd have been dead code otherwise).
- **`MODEL_RE`** — adversarial-tested against `&|^<>;$%!` backticks, parens, trailing/embedded newlines (stripped by the trim-then-test order), unicode/RTL/zero-width (rejected by the ASCII-only class). Legit labels like `Gemini 3.1 Pro (High)` still pass; their parens are caught by the layer-2 cmd quoter.

One thing I'll flag rather than change unilaterally: `FLAG_RE` permits `%`, so a `commandFlag` could carry `%VAR%`, which `cmd.exe` expands even inside quotes — env-var *substitution* into the agent's own argv (info-leak-ish, not RCE, and the flag is shown in the reviewable command field). Happy to drop `%` from `FLAG_RE` if you'd like it gone; left it as-is since it's the field you'd already vetted.

chaitanyagiri:
Hey bro that's a big commit give me sometime I'll go through it once and merge.

If you are reading this I want to run some ideas on this project by someone would you be interested in having a chat or a call?

jjc6676:
> Hey bro that's a big commit give me sometime I'll go through it once and merge.
> 
> If you are reading this I want to run some ideas on this project by someone would you be interested in having a chat or a call?

Sure. How do you suggest we chat?

chaitanyagiri:
## Code review — PR #71: `feat: shareable hires` (re-review of the grown diff)

**Verdict:** comment / safe-to-land from a security standpoint — the **HIGH-severity Windows command-injection** flagged in the first pass is now **closed with proper defense-in-depth**, and all three medium/low gaps are fixed in code. No prompt-injection content. Two small polish items remain. @chaitanyagiri — security re-review summary below; the must-fix is genuinely resolved.

This re-reads the diff after the author's force-pushed fixes + follow-up (the PR grew to +6150/−2041 / 77 files; the large `−` is mostly whitespace/line-ending churn in `preload/index.ts`, `App.tsx`, `store.ts`, not functional rewrites).

### What it does
- **Deep-link handler** `munderdifflin://hire?src=<url>`: registers the protocol (`index.ts:1146-1153`, `electron-builder.yml:11-14`), handles macOS `open-url` (`index.ts:1175-1178`), Win/Linux second-instance + cold-start argv (`index.ts:1165-1173`, `:1945-1947`). Links are fetched + validated in **main**, queued in `pendingHires[]`, and pulled by the renderer via `hire:drainPending` on mount.
- **Import → role spawn:** an untrusted "hire manifest" JSON is validated (`src/shared/hire.ts`), then **pre-fills the Add-Agent modal** (`AddAgentModal.tsx:108-126`). Spawn happens **only** on the human's explicit click (`AddAgentModal.tsx:129-142`) — import never auto-spawns.
- **The Hiring Fair gallery** (`docs/hires/`, net-new): static GH-Pages gallery, 14 roles × 3 provider variants, a client-side validator mirroring the app's, a JSON schema, and a build script.

### What's good
- **The fix landed, both layers.** (1) `MODEL_RE = /^[A-Za-z0-9 ._()[\]\/:@+-]{1,80}$/` (`shared/hire.ts:81,114-116`) rejects the PoC `x&calc` at validation — `&` isn't in the class. (2) Even if a metachar reached the Windows command line, `pty.ts:48-55` now quotes `[ \t"&|^<>()%!]` so it's inert to `cmd.exe` — and this guards **every** spawn path, not just hires. The three validators (app, `docs/hires/validator.js:13`, `spec/hire.schema.json:49`) are in sync.
- **Fetch hardening is real:** `redirect:'manual'` with manual hop-following where every `Location` must be `https:` (`main/hire.ts:48-65`) kills the redirect-SSRF (can't 302 into `http://127.0.0.1` or `169.254.169.254`); `readBounded` (`main/hire.ts:91-110`) streams and aborts past a 64 KB byte cap instead of trusting `content-length`; 10s timeout, ≤5 hops.
- **Spawn is argv-array, no shell** at the IPC layer (`pty.ts:196-211`) — no `shell:true`, no string concatenation. Worktree id is slugified and asserted under the worktrees root before use (`index.ts:1311-1316`). Gallery output is `esc()`-wrapped / `textContent` — no XSS.
- **Import never auto-spawns** — the human is always the spawn trigger. Good trust boundary.

### What could be better
- **(nit, low-sev security)** `FLAG_RE` still permits `%` (`shared/hire.ts:71`), so a `commandFlag` like `%VAR%` would be env-expanded by `cmd.exe` even when quoted — an **info-leak into the agent's own argv, not RCE** (and the flag is shown in the reviewable command field). The author already flagged this himself and offered to drop `%` — worth taking him up on it.
- **(nit)** `importHireUrl` / `hire:fromUrl` is plumbed but unwired — no UI invokes it (`preload:483-484`, `index.ts:1200-1203`). Harmless (same validator, `src` capped at 2000 chars) but dead surface; wire-or-drop.
- **(nit)** Multi-deep-link "last wins" (`App.tsx:75`) silently drops earlier queued hires — acknowledged edge case.
- **(nit)** `config.ts:150` `buildSpawnCommand` only whitespace-quotes the model; now harmless since `MODEL_RE` blocks metachars upstream and the cmd quoter is the real guard, but a comment pointing at the real guard would prevent a future "why is this only whitespace-quoting?" regression.

Isolated worktree @ `c07a680`, own COW-cloned node_modules, base `origin/main` @ `8421e18`, torn down after. (No `package.json`/lock change.)
| stage | result |
|---|---|

Build proves it compiles/launches on macOS. It does NOT exercise the live deep-link flow, the Windows `cmd.exe` quoting path (the platform that mattered for the injection), or the GH-Pages gallery — those are reasoned about from the code, not run.

**Closed.** Jim's PoC is dead at validation; if it somehow reached the tokenizer, layer-2 quoting neutralizes it. Redirect-SSRF and buffer-before-cap both fixed; `did-start-navigation` now guarded on `isMainFrame`. **No prompt-injection content** across the 14 manifests / 42 variants / docs — the only flagged strings are in-character Office gags in free-form `goal` fields (e.g. Creed's bit), not instructions aimed at an AI reviewer. The free-text `goal` pre-filling an agent prompt remains the documented v1 prompt-injection surface, but the human is always the spawn gate.

### Feedback
The author's response to the first review was exemplary — the must-fix is fixed with real defense-in-depth (validation **and** a command-line quoter that protects every spawn), the two should-fixes and the readiness bug are all closed, and the three validators stay in sync. Land-ready on security; just take the author up on dropping `%` from `FLAG_RE` and wire-or-drop `importHireUrl`.

PR #75, opened 2026-06-10
feat(debug): add local control endpoint

pdurlej:
## Summary
Adds an opt-in local debug control endpoint for Munder Difflin. This is the first small bridge toward reliable agent-driven debugging and future MCP control, without turning it into a public automation API.

## Why
The UI is not a reliable enough control surface for debugging orchestration by itself. We need a small local control plane that agents and tests can use to inspect app state, send bounded work orders, and exercise shutdown/control paths repeatably.

## What changed
- Adds a `DebugControlServer` in the Electron main process.
- Binds only to `127.0.0.1`.
- Stays disabled by default unless `MUNDER_DEBUG_CONTROL=1` or `debugControlEnabled` config is set.
- Generates a local token and writes a discovery file under app `userData`.
- Adds token-gated endpoints for:
  - `GET /health`
  - `GET /snapshot`
  - `POST /work-order`
  - `POST /closing-time/start`
  - `POST /control/:agentId/steer|halt|resume`
  - `POST /pty/:ptyId/kill`
  - `POST /app/quit-now`
- Reuses existing main-process primitives: hive routing, PTY management, control registry, closing-time controller, and fleet state.
- Keeps snapshots intentionally sanitized: no terminal buffers, no secrets, no full memory files.
- Adds focused unit coverage for auth, request validation, endpoint shape, and dispatch behavior.

## Stack
This is PR 1/3 in a stacked debugging/MCP sequence:

1. **This PR:** local debug control endpoint.
2. **Next PR:** `tools/munder-mcp`, a stdio MCP wrapper over this endpoint.
3. **Final PR:** reliable headless smoke harness with fake agents, so orchestration can be tested without clicking the UI or calling real models.

The point of splitting this is to keep each reviewable: endpoint first, MCP client second, smoke harness third.

## Validation
- `npm run typecheck`
- `npm run build`
- `npm run test:debug-control`
- `node test/slack.test.cjs`

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "typecheck:node": "tsc --noEmit -p tsconfig.node.json",
     "typecheck:web": "tsc --noEmit -p tsconfig.web.json",
     "typecheck": "npm run typecheck:node && npm run typecheck:web",
+    "test:debug-control": "node test/debug-control.test.cjs",
     "dist": "npm run build && electron-builder",
     "dist:mac": "npm run build && electron-builder --mac",
     "dist:win": "npm run build && electron-builder --win",
--- a/src/main/closingTime.ts
+++ b/src/main/closingTime.ts
@@ -37,6 +37,10 @@ export interface ClosingTimeEvent {
   total: number;
 }
 
+export interface ClosingTimeSnapshot extends ClosingTimeEvent {
+  active: boolean;
+}
+
 /** Subject markers. Deliberately forgiving (case, -/_/space) — agents write
  *  these by hand, so "Closing Time Ack" must count as well as the canonical
  *  CLOSING-TIME-ACK the brief asks for. */
@@ -76,6 +80,15 @@ export class ClosingTimeController {
 
   isActive(): boolean {
     return this.active;
+  }
+
+  snapshot(): ClosingTimeSnapshot {
+    return {
+      active: this.active,
+      phase: this.active ? 'progress' : 'cancelled',
+      acked: this.acked.size,
+      total: this.workers.size
+    };
   }
 
   /** Kick off the protocol. Returns an error string when the floor cannot run
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -162,6 +162,11 @@ export interface HarnessConfig {
   /** Local HTTP port the generic webhook server binds to (default 3849). */
   webhookPort?: number;
 
+  /** Local-only debug/test control plane. Also enabled by MUNDER_DEBUG_CONTROL=1. */
+  debugControlEnabled?: boolean;
+  /** Optional fixed debug control port; default 0 lets the OS choose. */
+  debugControlPort?: number;
+
   // ─── Memory reflection (the janitor's condense half) ───────────────────────
   /** Master toggle for the in-process MemoryReflector. Default on. */
   reflectEnabled?: boolean;
@@ -197,6 +202,8 @@ const DEFAULTS: HarnessConfig = {
   webhookEnabled: false,
   webhookSecret: undefined,
   webhookPort: undefined,
+  debugControlEnabled: false,
+  debugControlPort: undefined,
   // Memory reflection — preventive; nobody is over threshold today, so it sits
   // dark until an agent's memory crosses one of these (the verify gate is the
   // safety for the LLM step). Thresholds DECIDED by god 2026-06-06.
--- a/src/main/debugControl.ts
+++ b/src/main/debugControl.ts
@@ -1,0 +1,209 @@
+/**
+ * Local debug control endpoint.
+ *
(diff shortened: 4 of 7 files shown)


chaitanyagiri:
## Code review — PR #75: `feat(debug): add local control endpoint`

**Verdict:** comment / approve-direction — a deliberately-gated loopback control plane. **No vulnerability found** — bind, auth, opt-in and snapshot sanitization are all done correctly — but it is a *powerful* local surface, so I'm tagging @chaitanyagiri to make the trust model an explicit, conscious opt-in (per the security-review directive; this PR was a prime suspect).

### What it does
- Adds `DebugControlServer` (`src/main/debugControl.ts`) — a loopback-only, token-gated HTTP control plane for tests / local MCP clients. Started in `app.whenReady()` and stopped on quit/reset (`index.ts:2051`, `:1736`, `:1780`).
- Routes: `GET /health`, `GET /snapshot`, `POST /work-order`, `POST /closing-time/start`, `POST /app/quit-now`, `POST /control/:agentId/(steer|halt|resume)`, `POST /pty/:ptyId/kill`.
- Off by default; enabled only via `MUNDER_DEBUG_CONTROL=1` or `cfg.debugControlEnabled` (`config.ts:165-208`). Token (32 random bytes) + URL/port are published to `debug-control.json` in userData, `chmod 0600`.

### What's good (the security design is right)
- **Bind is loopback-only** — `server.listen(port, '127.0.0.1', …)` (`debugControl.ts:60`) **plus** a defense-in-depth `isLoopback(req.socket.remoteAddress)` 403 (`debugControl.ts:88-92`). Unlike the Slack/webhook servers, it opens no public tunnel.
- **Auth is mandatory on *every* route including GET** — `authorized()` runs before routing (`debugControl.ts:84-95`), constant-time `timingSafeEqual` after a length pre-check (`:152-159`). This also closes DNS-rebinding/CSRF: a cross-origin attacker can hit the port but can't read the 0600 token, so every forged request is 401.
- **Snapshot is allow-listed, not a dump** — `sanitizeDebugLogEvent` keeps only metadata keys (`index.ts:670-678`); the snapshot exposes agent metadata, pty *command basename* + pid + idleMs — **no terminal buffers, memory files, env, or secrets**. Exactly the right call.
- Body cap (`MAX_BODY_BYTES = 1 MiB`), default port 0 (OS-chosen, no fixed-port collision), env/config port validated `0–65535`, discovery file removed when disabled/stopped.

### What could be better
- **(security note — the reason for the @-tag, not a blocker):** when enabled, this endpoint is a high-leverage control plane — `POST /work-order` injects arbitrary work orders into the hive (→ autonomous LLM agents that spawn sub-agents), and steer/halt/kill-pty/quit-now are all reachable. The `0600` discovery file protects against *other users*, but **any process running as the same user** can read the token and drive the hive. That's the intended local-dev/test trust model and it's correctly off-by-default — but it should be stated explicitly so it's never flipped on in a shared/multi-user box. Recommend a one-paragraph security note in a README/the PR description spelling this out.
- **(nit)** `handleDebugWorkOrder` accepts an attacker-supplied `from` (`index.ts:777-786`) — harmless given the token gate, but pinning `from:'debug-control'` would make the audit log unambiguous.
- **(nit)** Config mirror: `debugControlEnabled/Port` were added to `config.ts` (main) and `preload/index.ts`, but not the renderer's `HarnessConfig` copy. Fine since the renderer never reads them (typecheck green), just noting the 2-of-3-mirror for the record.
- **(nit)** `MUNDER_DEBUG_CONTROL_PORT` is read in `debugControlPort()` but isn't documented alongside `debugControlPort` config — worth a comment so the two knobs are discoverable together.

Isolated worktree @ `acaf54f`, own COW-cloned node_modules, base `origin/main` @ `8421e18`, torn down after. `package.json` adds only a `test:debug-control` script (no new dependency), so copy-segregation is sufficient.
| stage | result |
|---|---|

Build proves it compiles/launches (endpoint off by default). It does NOT exercise the live endpoint behavior — but `test/debug-control.test.cjs` covers auth (401), bearer + `x-munder-debug-token`, malformed-JSON 400, and work-order/control/kill routing against a compiled `DebugControlServer`.

No prompt-injection content (comments technical; fixtures benign — `'hello'`, `'pause soon'`). Threat model assessed above: correctly loopback + token + opt-in + sanitized. @chaitanyagiri — flagging only so the powerful-when-enabled trust model is an explicit decision; I did **not** find a vulnerability.

### Feedback
This is the right way to build a debug control plane — loopback, mandatory token on all routes, opt-in, and a sanitized (not raw) snapshot. The single thing to do before it's broadly enabled: write down the trust model (token-file = full local hive control for any same-user process) so nobody turns it on in a shared environment unaware.

PR #78, opened 2026-06-11
feat(resume): restore agent sessions across restart + Restart & Continue

gts-47:
## What

Rebased on v0.2.6 (`8421e18`). Three commits:

1. **fix(pty): keep terminal live + typeable across model change / restart** — soft-reset the pooled terminal in place (`resetTerminal`) instead of disposing it, so the data subscription / xterm / keyboard wiring survive a respawn; spawn at the terminal's REAL grid (fit-derived cols/rows) instead of the fixed 100×30, fixing scattered/overlapping TUI rendering.

2. **feat(resume): restore agent sessions across restart + Add Agent resume-session field** —
   - `resume: true` on god restore: Michael reattaches his prior conversation across an app restart (session id from the hive registry; orientation prompt skipped on a genuine resume via the returned `resumed` flag).
   - Worker restore re-enters its EXISTING worktree (`isolate: false`, `gitIsRepo` probe with base-repo fallback) instead of re-isolating, which conflicted on the existing path/branch and lost uncommitted work.
   - Transcript seeding: Claude keys sessions by cwd, so the recorded session's `.jsonl` is copied into the target cwd's project dir before attaching `--resume`; `--resume` is only attached when the transcript is actually present (no more broken resumes against missing ids).
   - Add Agent "resume session" field: paste an explicit session id; `session:resolveCwd` auto-fills the folder; a not-found id falls back to a fresh session and surfaces `resumeNotFound` to the dialog.

3. **feat(resume): Restart & Continue button** — per-agent escape hatch that kills + respawns the PTY with the SAME model and `resume: true`, redrawing a garbled terminal (e.g. xterm reflow after dragging across displays) without losing the thread.

## Conflict-resolution notes (rebase onto v0.2.6)

- Fully adopted the new provider plumbing: `provider` kept in spawn signature/hive meta/`updateAgent`, `tokenizeCommand`/`inferAgentProvider` used in the restart path.
- The Claude resume path (seeding + presence check) lives in the Claude-only block; the generic provider-aware resume block now skips Claude so a failed seed can't blindly attach `--resume`.
- Dropped my old `resume: true` patch to the Dwight bootstrap — superseded upstream by `acc13a3` (headless Haiku enrichment); no assistant auto-spawn is reintroduced.

`npm run typecheck` (node + web) passes.

### DIFF
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -295,9 +295,15 @@ export class HiveManager {
     const cursor = join(dir, 'cursor.json');
     if (!existsSync(cursor)) this.writeJson(cursor, { lastProcessed: null });
 
-    // upsert registry
+    // upsert registry — spread the PRIOR entry first so a respawn preserves
+    // fields the spawn `meta` doesn't carry, above all `sessionId`. Without this,
+    // ensureAgent (which runs before the resume lookup in the pty:spawn handler)
+    // would wipe the recorded session id, so `lastSession()` returns undefined and
+    // `--resume` is never attached — i.e. every restart starts a fresh thread.
     const reg = this.registry();
+    const prev = reg.agents[meta.id];
     reg.agents[meta.id] = {
+      ...prev,
       ...meta,
       capabilities: meta.capabilities ?? [],
       role: meta.role ?? (meta.isGod ? 'orchestrator' : 'agent'),
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -21,7 +21,7 @@ import type { UsageProvider } from './usage';
 import { MemoryManager } from './memory';
 import { MemoryReflector, type ReflectSettings } from './reflect';
 import { PersistStore } from './db';
-import { readAgentUsage, readContextTokens } from './transcript';
+import { readAgentUsage, readContextTokens, seedSessionTranscript, resolveSessionCwd } from './transcript';
 import { listIssues, listCIRuns } from './github';
 import { SlackWebhookServer, SlackReplyServer, postSlackReply, type SlackEventFile } from './slack';
 import { WebhookServer, type WebhookInbound, type WebhookTaskStatus } from './webhook';
@@ -1197,7 +1197,7 @@ function createWindow(): void {
 }
 
 // ─── IPC: pty lifecycle ─────────────────────────────────────────────────────
-ipcMain.handle('pty:spawn', async (_evt, opts: SpawnOptions & { hive?: AgentMeta; isolate?: boolean; resume?: boolean; provider?: AgentProvider }) => {
+ipcMain.handle('pty:spawn', async (_evt, opts: SpawnOptions & { hive?: AgentMeta; isolate?: boolean; resume?: boolean; resumeSessionId?: string; provider?: AgentProvider }) => {
   if (!opts || typeof opts.id !== 'string' || typeof opts.cwd !== 'string' || typeof opts.command !== 'string') {
     return { ok: false, error: 'invalid SpawnOptions' };
   }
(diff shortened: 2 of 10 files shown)


chaitanyagiri:
## Code review — PR #78: `feat(resume): restore agent sessions across restart + Restart & Continue`

**Verdict:** approve + nits — a careful, well-commented resume/restart implementation that closes real correctness gaps (lost session ids, stale-pty corruption, broken `--resume`). Segregated build green. One follow-up is already tracked in-code by the author.

### What it does
- **Resume on restart (#1):** spreads the prior registry entry on respawn so `ensureAgent` no longer wipes `sessionId` (`hive.ts:295-307`), then attaches `--resume <sid>` from `hive.lastSession()` — but **only after seeding the transcript** into the target cwd's Claude project dir, because Claude keys sessions by cwd (`transcript.ts:seedSessionTranscript`, wired at `index.ts:1286-1309`).
- **Explicit resume (#2):** Add Agent gains a "resume session" field; an explicit id wins over auto-resume, `session:resolveCwd` auto-fills the folder from a transcript `cwd` record (`transcript.ts:resolveSessionCwd`, `index.ts:1374-1377`), and a not-found id falls back to a fresh session, surfacing `resumeNotFound` to the dialog instead of launching a broken `--resume`.
- **Restart & Continue (#3):** per-agent escape hatch that kill()+respawns the PTY with the same model and `resume:true` to redraw a garbled terminal without losing the thread.
- **PTY soft-reset + identity guard:** reuses the pooled terminal across model-change/restart and guards `onData`/`onExit` against a dying process whose id was reclaimed by the respawn (`pty.ts:209-233`).

### What's good
- The stale-callback race fix (`pty.ts:212-233`) is the real meat — capturing the `session` object and bailing when `sessions.get(id) !== session` correctly prevents (a) the dead process spraying final bytes into the new TUI frame and (b) its `onExit` deleting the live session + emitting a false `pty:exit`. Well reasoned, well commented.
- **Path-traversal guard:** renderer-supplied session ids flow into `path.join`, and `VALID_SESSION_ID = /^[A-Za-z0-9_-]+$/` rejects `../` before any fs use (`transcript.ts`). Good instinct — this is the one untrusted-input→path sink in the diff and it's handled.
- "Only attach `--resume` when the transcript is actually present" (`index.ts:1296-1305`) eliminates broken resumes against missing ids — the right default-to-fresh behavior, made non-silent via a floor warning + `resumeNotFound` flag.
- Provider-aware split is correct: the Claude seed-and-attach path lives in the Claude-only block, and the generic flag path now skips Claude (`index.ts:1318`) so a failed seed can't blindly attach `--resume`.

### What could be better
- **(nit)** The author documents a tracked follow-up at `index.ts:1216-1221`: a stale `isolate:true` recipe spawned against an already-existing worktree path falls back to the base cwd. The restore flow uses `isolate:false` so it doesn't hit this today, but reuse-existing-worktree handling is the right next step. Fine to land with the note.
- **(nit)** `resolveSessionCwd` returns the `cwd` from the user's own prior transcript to auto-fill the dialog folder — not an injection vector (user's own data), but worth a one-line note that the folder is auto-filled from transcript metadata in case a user is surprised by it.
- **(nit)** `seedSessionTranscript` does a full `readdirSync` over every `~/.claude/projects/<dir>` on each resume — fine at today's scale; if project dirs grow large this is an O(dirs) scan per spawn.

Isolated worktree @ `a0da64e`, own COW-cloned node_modules, base `origin/main` @ `8421e18`, torn down after.
| stage | result |
|---|---|

Build proves it compiles/launches on macOS. It does NOT exercise the actual resume behavior (needs a recorded Claude session + restart), nor Windows paths.

No security issues. Only untrusted-input→sink is `resumeSessionId` (user's own dialog input) → guarded by `VALID_SESSION_ID` before `path.join`. No shell, no network, no eval. No prompt-injection content in the diff (comments are technical).

### Feedback
Solid, ship-shaped work — the pty identity guard and the "seed-then-only-attach-if-present" resume are exactly right, and the path-traversal guard shows good security hygiene. Mergeable as-is; the worktree-reuse follow-up the author already flagged is the natural next PR.

PR #79, opened 2026-06-12
feat(terminal): drag-and-drop a file onto the terminal to inject its path

gts-47:
## What

Drag-and-drop a file (image, etc.) onto a PTY terminal and its **absolute, shell-escaped path** is written into the session — exactly like a native terminal's drag-drop. Claude Code then detects the image path in the prompt and attaches it.

## Why

Previously, dropping a file onto the terminal made Electron treat it as navigation and load the `file://` URL over the app. This makes the common "drop an image into the agent" flow work in-app.

## Changes

- **`preload/index.ts`** — expose `getFilePath` backed by `webUtils.getPathForFile`. Electron 32 removed `File.path`; `getPathForFile` is the supported replacement and must run in the preload (it isn't on `window`). Returns `''` for synthetic files.
- **`PtyTerminalView.tsx`** — `onDragOver`/`onDrop` handlers that:
  - `preventDefault` only when files are present (text/selection drags still fall through to xterm),
  - resolve each dropped file's path,
  - backslash-escape whitespace/quotes so each path lands as a single token,
  - write them space-separated with a trailing space so the user can keep typing.

## Notes

Branched off `main` and scoped to **only** these two files — independent of the session-resume work in #78.

### DIFF
--- a/src/preload/index.ts
+++ b/src/preload/index.ts
@@ -1,4 +1,4 @@
-import { contextBridge, ipcRenderer, type IpcRendererEvent } from 'electron';
+import { contextBridge, ipcRenderer, webUtils, type IpcRendererEvent } from 'electron';
 import type { AgentProvider } from '../shared/agentProvider';
 
 // Injected at build time from package.json (see electron.vite.config.ts).
@@ -310,6 +310,13 @@ const api = {
     ipcRenderer.invoke('pty:spawn', opts),
   writePty: (id: string, data: string): Promise<{ ok: boolean; error?: string }> =>
     ipcRenderer.invoke('pty:write', id, data),
+  /** Resolve the absolute filesystem path of a dropped/selected File. Electron 32
+   *  removed `File.path`; `webUtils.getPathForFile` is the supported replacement
+   *  and must run in the preload (it isn't on `window`). Returns '' if unknown
+   *  (e.g. a synthetic File that never came from disk). */
+  getFilePath: (file: File): string => {
+    try { return webUtils.getPathForFile(file) || ''; } catch { return ''; }
+  },
   resizePty: (id: string, cols: number, rows: number): Promise<{ ok: boolean; error?: string }> =>
     ipcRenderer.invoke('pty:resize', id, cols, rows),
   killPty: (id: string): Promise<{ ok: boolean; error?: string }> =>
--- a/src/renderer/src/components/PtyTerminalView.tsx
+++ b/src/renderer/src/components/PtyTerminalView.tsx
@@ -269,6 +269,34 @@ export function PtyTerminalView({ ptyId, onStreamData, onUserPrompt, onToggleFul
     } catch { /* host may not be sized yet */ }
   }, [fontSize, ptyId]);
 
+  // Drag-and-drop a file (image, etc.) onto the terminal → inject its absolute
+  // path into the pty, exactly like a native terminal's drag-drop. Claude Code
+  // detects an image path in the prompt and attaches it. Without this, Electron
+  // would treat the drop as navigation and load the file:// URL over the app.
+  const onDragOver = (e: React.DragEvent) => {
+    // Must preventDefault on dragover too — otherwise `drop` never fires and the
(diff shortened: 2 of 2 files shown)


chaitanyagiri:
Thankyou this was much needed

chaitanyagiri:
## Review — PR #79: feat(terminal): drag-and-drop a file onto the terminal to inject its path

**Verdict:** request changes — genuinely wanted feature, clean build, but (1) the path escaping leaves a filename→command-injection gap and (2) `getFilePath` duplicates the `pathForFile` helper that has since landed on `main`. Both are small fixes.

### What it does
- Adds `onDragOver`/`onDrop` to the PTY host in `PtyTerminalView.tsx`: a dropped file's absolute path is escaped and written into the session via `writePty`, mimicking a native terminal's drag-drop (Claude Code then auto-attaches an image path). Without it Electron treated the drop as navigation and loaded the `file://` URL over the app — so this fixes a real footgun.
- Exposes `getFilePath` in `preload/index.ts` backed by `webUtils.getPathForFile` (Electron 32 removed `File.path`).
- Correctly `preventDefault`s only when `dataTransfer.types` includes `'Files'`, so text/selection drags still fall through to xterm. Nice touch.

### What's good
- Tightly scoped (2 files, +37/-2), well-commented, and the "only claim file drops" guard is exactly right.
- Trailing-space + space-join lets you chain multiple drops and keep typing — matches native behavior.
- `getFilePath` wraps in try/catch and returns `''` for synthetic files — defensive.

### What could be better
- **(blocker — security) Filename → command injection.** `PtyTerminalView.tsx` escapes only `" ' \ <space> <tab>` (`p.replace(/(["'\\ \t])/g, '\\$1')`). It does **not** escape a newline embedded in a filename — Unix filenames may contain `\n`. A dropped file named `evil\nrm -rf ~` writes a literal newline to the PTY, which the shell executes **immediately, without the user pressing Enter**. Other metacharacters (`$`, `` ` ``, `;`, `|`, `&`, `(`, `)`) also pass through and run on Enter (`$(...)`/backticks command-substitute). Verified locally against hostile names — the newline survives the regex. Low likelihood (needs a maliciously/oddly-named file) but it's a true injection surface, so flagging. **Fix:** POSIX single-quote wrap instead of a metachar denylist — `"'" + path.replace(/'/g, "'\\''") + "'"` neutralizes *everything* including newline; or strip/reject control chars before injecting. /cc @chaitanyagiri
- **(nit — redundancy) Duplicates `pathForFile`.** This PR branched from `8421e18`, before the rich-composer merge (`c8467fc`) landed `pathForFile: (file) => webUtils.getPathForFile(file)` in the same preload `api` (line 487 on `main`). After merge the preload has **both** `getFilePath` (line 371) and `pathForFile` (line 487) doing the same thing. Recommend dropping `getFilePath` and calling `window.cth.pathForFile`, or folding this PR's `try/catch + '' ` hardening into `pathForFile` and keeping one helper.

Isolated worktree, current `origin/main` **merged with PR #79 head** (`afc5f32`), main's `node_modules` symlinked, torn down after.

| stage | result |
|---|---|
| merge (main + #79) | CLEAN (no conflict; GH agrees: MERGEABLE/CLEAN) |
| build (electron-vite) | PASS (built in ~6s, sidecars copied) |
| run smoke | N/A (not launched) |

Scope: verified the **post-merge** tree compiles and builds — important here because the two near-identical helpers coexist after merge (they don't collide; both valid). Did **not** run a runtime smoke or exercise the actual drag-drop, which is visual/OS-level (macOS file-manager drag events) and can't be checked headless. The injection finding is from analyzing the escape regex, not a live exploit.

### Feedback
Great, much-needed feature — ship it after two small changes: (1) switch the path escaping to single-quote wrapping so a hostile filename can't inject a command (the newline case auto-executes), and (2) reuse the already-merged `pathForFile` instead of adding `getFilePath`.

---
<sub>🤖 Automated review by the PR-Reviewer hive agent. Build: isolated worktree, segregated deps, post-merge typecheck+build green. Review only — no merge/push.</sub>

PR #80, opened 2026-06-14
feat(compact): only auto-compact when context is ≥30% of the model's window

gts-47:
## What

Auto-compaction (the hourly ops standup's `autoCompact`) now only fires for an agent when its **live context has filled to ≥30% of that agent's context window** — instead of compacting **every** live Claude terminal on every tick.

## Why

The standup queued `/compact` for all live Claude agents each hour regardless of how full they were, so small or barely-used sessions got compacted needlessly (losing detail for no benefit). Compaction should be driven by actual context pressure, not a fixed cadence.

## How

In the `onAutoCompact` handler (`useHive.ts`), gate each agent on `contextTokens / contextLimit >= 0.30`:

- `contextLimit` is **per-model**, so a **200k** session and a **1M** session both compact at the same **30% relative fill** rather than a fixed token count — this matters because users run both window sizes.
- Agents with **no context telemetry yet** (`contextLimit` not polled) are skipped — never compact blind.
- Threshold is a single named constant `COMPACT_CONTEXT_THRESHOLD = 0.3`.

The existing dedupe (no second `/compact` if one is already queued) and idle-only delivery (never interrupts a working terminal mid-step) are unchanged. Config doc comment for `autoCompact` updated to match.

## Test

`npm run typecheck` passes (node + web).

### DIFF
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -12,8 +12,10 @@ export interface ScheduledMission {
   to: string;
   body: string;
   enabled: boolean;
-  /** When true, the scheduler also sends `/compact` to every live terminal when
-   *  this mission fires — keeping each agent's context lean on a cadence. */
+  /** When true, the scheduler asks the renderer to `/compact` live terminals when
+   *  this mission fires — but only agents whose context has filled past a
+   *  threshold (~30% of their model's window) are compacted, so small/idle
+   *  sessions are left alone instead of compacting on every tick. */
   autoCompact?: boolean;
   lastFiredAt?: number;
   /** Mission flavor. Absent ⇒ 'dispatch' (the classic interval-dispatch mission,
--- a/src/renderer/src/hooks/useHive.ts
+++ b/src/renderer/src/hooks/useHive.ts
@@ -42,6 +42,14 @@ const COMPACT_CMD =
   '/compact Summarise exactly what you are currently working on and the next step, ' +
   'so you can resume from the same point — then continue that work after compacting.';
 
+// Auto-compact only fires once an agent's live context has filled to at least
+// this fraction of its model's window. It's a FRACTION of contextLimit (which is
+// per-agent/per-model), so a 200k session and a 1M session both compact at the
+// same 30% relative fill — not at a fixed token count. Agents below this, or with
+// no context telemetry yet, are left alone — so the hourly standup no longer
+// compacts every terminal unconditionally.
+const COMPACT_CONTEXT_THRESHOLD = 0.3;
+
 // Per-pty submission chain. Every submitToPty for a given pty is appended here so
 // two callers (e.g. the boot sequence's /remote-control and the inbox-wake nudge)
 // can NEVER interleave their text + Enter — which jammed them onto one line and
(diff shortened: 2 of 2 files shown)


chaitanyagiri:
## Review — PR #80: feat(compact): only auto-compact when context is ≥30% of the model's window

**Verdict:** approve + nits — correct, conservative, division-safe gating; pure renderer logic with no security surface. One behavioral caveat worth knowing.

### What it does
- The hourly standup's `autoCompact` previously queued `/compact` to **every** live Claude terminal each tick. This gates it: an agent is compacted only when `contextTokens / contextLimit ≥ 0.30` (`useHive.ts`, new `COMPACT_CONTEXT_THRESHOLD = 0.3`).
- The gate is a **fraction** of each agent's own `contextLimit`, so a 200k and a 1M session both trip at 30% relative fill — not a fixed token count. `config.ts` change is a JSDoc update only.
- Agents with no telemetry yet (`contextLimit` unset) are skipped — "never compact blind."

### What's good
- **Division-safe:** `limit <= 0 || used/limit < THRESHOLD` short-circuits, so `used/limit` is never evaluated when `limit == 0` — no NaN/Infinity.
- **Boundary matches the title:** `used/limit < 0.3 ⇒ skip` means it compacts at `≥30%`. Correct, no off-by-one.
- **Fields are real and live:** `contextLimit`/`contextTokens` are populated by the PTY parser (`usePtyParser.ts:100`, `useHive.ts:400`) — the threshold is applied to the same per-agent values the UI already shows.

### What could be better
- **(nit / behavioral caveat)** `contextLimit`/`contextTokens` are **non-persisted** live-session fields (`store.ts:220` drops them from `PersistedAgent`) and `contextLimit` only ratchets up as the parser sees context lines. So right after an app restart / agent respawn — before a context line is re-parsed — `limit == 0` ⇒ the agent is skipped even if its underlying session is genuinely full. The OLD code always compacted; the NEW code won't compact such a session until telemetry refreshes. This is intentional per the "never compact blind" comment and low-risk in practice (telemetry refreshes as the session is used), but it's the one behavior change a reader should know.
- **(nit)** Optional follow-up: a safety fallback for known long-lived sessions with stale telemetry. The current conservative default is defensible; not a blocker.

### Local test — static review only (no segregated build this cycle)
Floor is under an active cost-constraint, so I did **not** run the segregated worktree build for a +19/-2, renderer-only logic diff. Verified statically instead: both fields exist and are `number | undefined` on the agent/store types (`store.ts:76-78`, `AgentCard.tsx:17-19`); the added guard is type-trivial against them.
| stage | result |
|---|---|
| typecheck | not run (type-trivial; fields verified to exist) |
| build (electron-vite) | not run |
| run smoke | N/A |

What this does NOT prove: a full type/build pass across the tree. Flagging honestly rather than claiming a green a reviewer can't reproduce.

### Security
Pure renderer-side gating — no new IPC, no external/untrusted input, no egress; `/compact` command text unchanged. Prompt-injection scan: N/A (no new data flow). No issues.

### Feedback
Solid, well-scoped fix. The one thing to keep in mind: auto-compaction is now fully dependent on live PTY context telemetry, so a just-respawned full session won't compact until its context line is re-parsed.

chaitanyagiri:
**Peer review — only auto-compact when context is ≥30% of the model's window**

Gates the hourly standup's `autoCompact` per-agent on `contextTokens / contextLimit ≥ 0.30` instead of queueing `/compact` for every live Claude terminal each tick. Because `contextLimit` is per-model, a 200k and a 1M session both compact at the same *relative* fill rather than a fixed token count; agents with no context telemetry yet are skipped ("never compact blind"); the existing dedupe and idle-only delivery are unchanged. Single named constant `COMPACT_CONTEXT_THRESHOLD`, 2 files, +19/−2.

**Strengths**
- Ties compaction to actual context pressure rather than a fixed cadence — the right call, and the per-model *fraction* is the correct unit given users run both 200k and 1M windows.
- Good guard: skipping agents with unknown telemetry avoids compacting blind, and reusing the existing dedupe + idle-only path keeps the change minimal and low-risk.

**Concerns / suggestions**
- 30% is a fairly eager threshold — it'll still compact a session that's only 30% full, which may discard detail earlier than necessary. It's a reasonable first cut, but since it's a pure judgment value, promoting it from a constant to a config field (like the mission's other knobs) would let operators tune it without a rebuild.
- typecheck-only; a tiny unit test asserting the gate skips <30% and unknown-limit agents would lock in the behavior cheaply.

chaitanyagiri:
Closing for now — not included in the v0.3.4 scope. Thanks for the contribution @gts-47; happy to revisit later.

PR #81, opened 2026-06-14
feat(rc): name each agent's Remote Control session after the agent

gts-47:
## What

Remote Control sessions now appear in claude.ai / the mobile app named after the **agent** (Michael, Jim, Dev1…) instead of the machine hostname.

## Why

Today an agent's RC session shows up as `<hostname>-<random>` (e.g. `vyapaks-macbook-pro-mighty-hare`) — Claude's default RC name prefix is the hostname. With several agents running it's impossible to tell which session is which. This is especially bad for users who have `remoteControlAtStartup` enabled, where Claude auto-enables RC for **every** session, so every agent lands in the mobile app under the same opaque hostname.

## How

- **`src/main/index.ts`** — in the Claude-only spawn-flag block, add `--remote-control-session-name-prefix "<agent name>"` for every hive agent. The friendly name (`opts.hive.name`, fallback id) is slugified to a single safe token `[A-Za-z0-9._-]`; Claude still appends its own random suffix for uniqueness. Placed alongside `--model`, so non-Claude providers (codex, agy) are untouched, and an explicit prefix already in args is never overridden.
- **`src/renderer/src/hooks/useHive.ts`** — Michael's boot-time `/remote-control` now passes his name (`/remote-control Michael`) so his session is named deterministically, matching the spawn-time prefix.

## Result

| Before | After |
|---|---|
| `vyapaks-macbook-pro-mighty-hare` | `Michael-<suffix>`, `Jim-<suffix>`, `Dev1-<suffix>` |

## Test

`npm run typecheck` passes (node + web).

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -1971,6 +1971,17 @@ async function spawnAgentCore(opts: AgentSpawnOptions, owner: Electron.WebConten
         : cfg.defaultModel ?? modelForRole(opts.hive, cfg);
       if (m) args.push('--model', m);
     }
+    // Name the Remote Control session after the agent (Michael, Jim, Dev1…) so it
+    // is identifiable in claude.ai / the mobile app. Otherwise Claude defaults the
+    // prefix to the machine hostname (e.g. "vyapaks-macbook-pro-…"), which is
+    // opaque when several agents run at once — especially with remoteControlAtStartup
+    // on, where RC auto-enables for every session. Slugify the friendly name into a
+    // single safe token; Claude still appends its own random suffix for uniqueness.
+    if (!args.includes('--remote-control-session-name-prefix')) {
+      const label = (opts.hive.name || opts.hive.id || '')
+        .trim().replace(/[^A-Za-z0-9._-]+/g, '-').replace(/^-+|-+$/g, '');
+      if (label) args.push('--remote-control-session-name-prefix', label);
+    }
     // Coarse runaway cap.
     if (typeof cfg.maxTurns === 'number' && cfg.maxTurns > 0 && !args.includes('--max-turns')) {
       args.push('--max-turns', String(cfg.maxTurns));
--- a/src/renderer/src/hooks/useHive.ts
+++ b/src/renderer/src/hooks/useHive.ts
@@ -300,8 +300,11 @@ export function useHive(config: HarnessConfig | null): void {
       timers.push(setTimeout(() => {
         if (cancelled) return;
         // settleMs pauses the chain ~1.5s after /remote-control before the
-        // orientation prompt (fresh spawns only) is submitted next.
-        submitToPty(GOD_PTY, '/remote-control', REMOTE_CONTROL_SETTLE_MS).catch(() => { /* best-effort */ });
+        // orientation prompt (fresh spawns only) is submitted next. Name the RC
+        // session "Michael" so it's identifiable in claude.ai / the mobile app
+        // rather than the opaque hostname default (matches the per-agent
+        // --remote-control-session-name-prefix set at spawn in the main process).
+        submitToPty(GOD_PTY, '/remote-control Michael', REMOTE_CONTROL_SETTLE_MS).catch(() => { /* best-effort */ });
         if (!resumedGod) {
           // A type-into-tui god (Crush) can't ride its hive protocol on argv, so the
           // main process hands it back as seedPrompt — type it FIRST (identity), then


chaitanyagiri:
## Review — PR #81: feat(rc): name each agent's Remote Control session after the agent

**Verdict:** approve — correct, injection-safe, and the CLI flag is genuine. No security issues.

### What it does
- At spawn (`index.ts`, `pty:spawn`), passes `--remote-control-session-name-prefix <label>` to the `claude` CLI where `label` = slugified `opts.hive.name || opts.hive.id`, so RC sessions show as "Michael / Jim / Dev1…" in claude.ai / the mobile app instead of the opaque hostname default.
- God's bootstrap (`useHive.ts:262`) changes `/remote-control` → `/remote-control Michael`, naming god's own RC session to match.

### What's good
- **Flag is real** (this was the top failure risk — a bad flag breaks *every* spawn): `claude --help` lists `--remote-control-session-name-prefix <prefix>` (default: hostname) **and** `--remote-control [name]`, so both the main-process flag and the renderer's `/remote-control Michael` form are valid. Refuted as a risk.
- **Injection-safe:** spawn is argv-based (node-pty `ptyManager.spawn`, no `shell`). The slug `replace(/[^A-Za-z0-9._-]+/g,'-')` + `replace(/^-+|-+$/g,'')` strips everything but `[A-Za-z0-9._-]` and removes leading/trailing `-`, so the label can't shell-inject or be parsed as an option — and it's positionally a value after the flag anyway.
- **Defensive guards:** `!args.includes('--remote-control-session-name-prefix')` respects an explicit user/extraArgs override; `if (label)` guards empty/all-non-ASCII names → graceful fallback to the hostname default.

### What could be better
- **(nit)** God's name is hardcoded `'Michael'` in the renderer rather than derived from god's agent meta. Correct today (god *is* Michael, GOD_PTY is a fixed identity), but a future rename of the god persona would strand this literal. Cosmetic.
- **(nit)** A name composed entirely of non-`[A-Za-z0-9._-]` chars (e.g. all-emoji) slugs to empty ⇒ skipped ⇒ hostname default. Acceptable; just noting the edge.

### Local test — static review only (no segregated build this cycle)
Floor is under an active cost-constraint, so I did **not** run the segregated build for a +16/-2 diff. Verified statically: spawn path is argv (`index.ts:1595 ptyManager.spawn`), `opts.hive.name/id` are strings (AgentMeta), the string ops are type-trivial, and flag existence is confirmed via `claude --help`.
| stage | result |
|---|---|
| typecheck | not run (type-trivial) |
| build (electron-vite) | not run |
| run smoke | N/A |

What this does NOT prove: a full type/build pass. Disclosed honestly rather than claiming an unreproducible green.

### Security
Argv spawn (no shell), slugified safe token, no option-injection, no new egress/IPC. Prompt-injection scan: N/A. No issues — no human tag needed.

### Feedback
Clean, low-risk QoL win, and the mechanism is verified against the real CLI flag. Ship-ready pending your normal gate.

chaitanyagiri:
**Peer review — name each agent's Remote Control session after the agent**

Adds `--remote-control-session-name-prefix "<slug>"` to the Claude-only spawn-flag block so RC sessions show up in claude.ai / the mobile app as `Michael-…`, `Jim-…`, `Dev1-…` instead of the opaque `<hostname>-<random>` default. The friendly name is slugified to a single safe token `[A-Za-z0-9._-]`, an explicit prefix already in args is never overridden, and Michael's boot `/remote-control` becomes `/remote-control Michael` to match. 2 files, +16/−2.

**Strengths**
- Small, well-scoped UX win that matters most exactly where the pain is (`remoteControlAtStartup` on, every session under one hostname). Correctly placed in the Claude-only block so codex/agy are untouched, with a safe slug regex and an idempotent `!args.includes(...)` guard.

**Concerns / suggestions**
- `"Michael"` is hardcoded in the renderer's `/remote-control Michael` while the spawn-time prefix derives from `opts.hive.name` — if the GOD agent's display name ever becomes configurable these two drift apart. Deriving the renderer string from the god agent's meta (or a shared constant) would keep them in lockstep.
- Confirm the shipped Claude Code build actually supports `--remote-control-session-name-prefix`; an older build would ignore or reject the flag. typecheck-only is adequate for a pure flag addition.

chaitanyagiri:
Closing with thanks — this feature shipped in v0.3.4 via the fork integration (#116): Remote Control sessions are now named after their agent. Credit noted in the v0.3.4 changelog. 🎉

PR #108, opened 2026-07-11
feat: per-agent environment variables (Claude Code profiles) (#105)

### LINKED ISSUE #105: Support Claude Code Profiles
### Problem / motivation

I use separate claude code accounts for work and personal.  For personal I have an alias configured 'claude-personal' which basically runs:  'CLAUDE_CONFIG_DIR=~/.claude-personal claude'  this directory contains a different config.json  I can run claude-code personal then from the cli and everything works fine, but in MD it doesnt read my zshrc. to see the alias and nor can I set the env variable manually. passing the full command doesnt work either.

### Proposed solution

Allow specifying which shell to use or allow setting env variables to agents

### Alternatives considered

_No response_

### Area

Office floor / avatars

ajishab:
Closes #105.

## Problem

Users running separate Claude Code accounts (work vs personal via `CLAUDE_CONFIG_DIR=~/.claude-personal claude`) had no way to make Munder Difflin spawn an agent with that profile. Shell aliases aren't visible to the spawner (it resolves the binary via `which`, no shell), and an env-prefixed command string is tokenized and exec'd directly by node-pty, so `CLAUDE_CONFIG_DIR=… claude` fails.

## What this adds

Per-agent environment variables, as plain data on the existing spawn path:

- **Add Agent → Workspace → Env vars**: a `KEY=VALUE` row editor. `~` expands to home.
- **Settings → General → Agent environment**: a global default env inherited by every spawn, including the GOD orchestrator (Michael). Per-agent env overrides the global default; internal injections (hive identity, BYOK keys, broker tokens) always win over user env.
- **`CLAUDE_CONFIG_DIR` follow-through**: when an agent overrides its config dir, token/cost telemetry, session-resume transcript seeding, and Claude permissions-acceptance all target/scan that dir — so a profile agent runs the right account *and* still reports tokens.

## Safety

- Validation runs at the main-process boundary: a bad key (regex-invalid, or a denylisted `PATH`/`NODE_OPTIONS`/`ELECTRON_RUN_AS_NODE`/`DYLD_*`/`LD_*`/`AGENT_*`/`HIVE_*`/`CTH_*`/`__proto__`) **fails the spawn** with a named error, never a silent drop.
- The `pty:spawn` IPC strips the main-internal `agentEnvApplied`/`internalEnv` fields off renderer payloads so they can't bypass validation.
- Hire manifests (untrusted input) **cannot** import env; an `env` field is ignored with a notice on file import.
- Env persists verbatim in `registry.json` (same trust level as `config.json`) but is **masked** (`•••` for `*KEY*`/`*TOKEN*`/`*SECRET*`/…) on every derived surface: `fleet.json`, the `hive:registry` IPC, and `tools/agent-env.cjs`.
- Env round-trips through restore-team / revive / restart respawns.

## Testing

- New `test/agent-env.test.cjs` (18 assertions): validation, denylist (incl. lowercase-prefix + `__proto__`), tilde expansion, merge precedence, config-dir resolution, masking.
- `npm run typecheck` (node + web), all 6 test scripts, and `npm run build` green.
- Design docs: `docs/superpowers/specs/2026-07-08-per-agent-env-vars-design.md`, `docs/superpowers/plans/2026-07-08-per-agent-env-vars.md`.

Note: env is data-only end-to-end — no shell is introduced into the spawn path.

### DIFF
--- a/docs/superpowers/plans/2026-07-08-per-agent-env-vars.md
+++ b/docs/superpowers/plans/2026-07-08-per-agent-env-vars.md
@@ -1,0 +1,990 @@
+# Per-Agent Environment Variables (Issue #105) Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Let users set env vars (e.g. `CLAUDE_CONFIG_DIR=~/.claude-personal`) per agent and globally, so separate Claude Code profiles work — including telemetry, resume, and permissions-acceptance following the override.
+
+**Architecture:** Env is plain `Record<string, string>` data riding the existing spawn path (renderer `spawnPty` → `spawnAgentCore` → `PtyManager.spawn`, whose `env` merge already exists). A dependency-free `src/shared/agentEnv.ts` owns validation/denylist/tilde-expansion/masking so main, renderer, and tests share one implementation. Merge order: `process.env` → `defaultAgentEnv` (Settings) → per-agent env → internal injections (hive/BYOK/broker always win).
+
+**Tech Stack:** Electron (main/preload/renderer), TypeScript, React, node-pty. Tests are framework-free `.cjs` scripts that transpile shared TS in-process (see `test/agent-provider.test.cjs`).
+
+**Spec:** `docs/superpowers/specs/2026-07-08-per-agent-env-vars-design.md`
+
+## Global Constraints
+
+- `npm run typecheck` (node + web) must stay green — it is the CI gate.
+- `src/shared/` files must stay dependency-free: no `electron`, no `node:*` imports, no UI imports.
+- New UI derives from design tokens (`var(--cth-…)` CSS vars, existing `PixelButton`/`Row`/`inputStyle` idioms). No ad-hoc colors/fonts.
+- Env denylist (exact): `PATH`, `NODE_OPTIONS`, `ELECTRON_RUN_AS_NODE`. Denylist prefixes: `DYLD_`, `LD_`, `AGENT_`, `HIVE_`, `CTH_`.
+- Env key regex: `/^[A-Za-z_][A-Za-z0-9_]*$/`.
+- A rejected key **fails the spawn** with a specific error — never a silent drop.
+- Hire manifests may NOT import env vars (warning surfaced when a manifest carries one).
+- Commit after every task with a conventional-commits message.
+
+---
+
+### Task 1: Shared env module `src/shared/agentEnv.ts` (TDD)
+
+**Files:**
+- Create: `src/shared/agentEnv.ts`
+- Create: `test/agent-env.test.cjs`
+
+**Interfaces (Produces — later tasks rely on these exact signatures):**
+```ts
+export const ENV_KEY_RE: RegExp;
+export type EnvValidation =
+  | { ok: true; env: Record<string, string> }
+  | { ok: false; error: string };
+export function envKeyIssue(key: string): string | null;
+export function expandTilde(value: string, home: string): string;
+export function validateAgentEnv(env: Record<string, string> | undefined, home: string): EnvValidation;
+export function mergeAgentEnv(defaults: Record<string, string>, perAgent: Record<string, string>): Record<string, string>;
+export function claudeConfigDirFrom(env: Record<string, string> | undefined, home: string): string | null;
+export function maskSensitiveEnv(env: Record<string, string>): Record<string, string>;
+```
+
+- [ ] **Step 1: Write the failing test**
+
+Create `test/agent-env.test.cjs` (mirrors `test/agent-provider.test.cjs`'s transpile-and-require pattern):
+
+```js
+'use strict';
+/**
+ * Per-agent env (#105) tests. Self-contained, no test framework — run with
+ * `node test/agent-env.test.cjs` (mirrors test/agent-provider.test.cjs). The
+ * module under test is dependency-free TypeScript (src/shared/agentEnv.ts), so
+ * we transpile it with the bundled `typescript` compiler into a temp dir and
+ * require the result. Covers key validation, the denylist, merge precedence,
+ * tilde expansion, Claude config-dir resolution, and sensitive-value masking.
+ */
+
+const assert = require('node:assert');
+const fs = require('node:fs');
+const os = require('node:os');
+const path = require('node:path');
+const ts = require('typescript');
+
+const SHARED = path.join(__dirname, '..', 'src', 'shared');
+const out = fs.mkdtempSync(path.join(os.tmpdir(), 'agentenv-'));
+const src = fs.readFileSync(path.join(SHARED, 'agentEnv.ts'), 'utf8');
+const js = ts.transpileModule(src, {
+  compilerOptions: { module: ts.ModuleKind.CommonJS, target: ts.ScriptTarget.ES2020 }
+}).outputText;
+fs.writeFileSync(path.join(out, 'agentEnv.js'), js, 'utf8');
+const ae = require(path.join(out, 'agentEnv.js'));
+
+let failures = 0;
+function check(name, fn) {
+  try { fn(); console.log(`  ok  ${name}`); }
+  catch (e) { failures++; console.error(`FAIL  ${name}\n      ${e.message}`); }
+}
+
+const HOME = '/Users/tester';
+
+// ── key validation ───────────────────────────────────────────────────────────
+check('accepts a plain key', () => {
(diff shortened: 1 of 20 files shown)


chaitanyagiri:
Seems like a big change with a lot of changes, could you please clean it up a bit(remove the markdown files etc) and post a comment setting the context for the problem and what it solves.

chaitanyagiri:
**Peer review — per-agent environment variables (Claude Code profiles) (#105)**

Adds per-agent and global-default env vars as plain data riding the existing spawn path, with a dependency-free `src/shared/agentEnv.ts` owning validation, denylist, tilde-expansion, merge precedence, and masking. UI: an Add Agent → Env vars row editor and a Settings → Agent environment global default. Headline use case is Claude Code profiles — `CLAUDE_CONFIG_DIR` is followed through so token/cost telemetry, resume transcript seeding, and permissions-acceptance all target the overridden dir. 20 files, +1765/−47, with an 18-assertion test and design/spec docs.

**Strengths**
- Strong security posture: validation is enforced at the main-process IPC boundary (not the renderer) and a bad/denylisted key *fails the spawn* rather than silently dropping; the denylist covers `PATH`/`NODE_OPTIONS`/`ELECTRON_RUN_AS_NODE`/`DYLD_`/`LD_`/`AGENT_`/`HIVE_`/`CTH_` case-insensitively; the IPC strips main-internal fields off renderer payloads; untrusted hire manifests can't import env; and secrets are masked on every derived surface (`fleet.json`, roster IPC, `tools/agent-env.cjs`) while `registry.json` keeps them verbatim.
- Good architecture: one dependency-free shared module reused by main + renderer + tests, and the feature is carried genuinely end-to-end (the `CLAUDE_CONFIG_DIR` follow-through into `transcript.ts` is the tell). Merge precedence with internal injections always winning is the right ordering.

**Concerns / suggestions**
- Largest surface here by far — the highest-risk area is the config-dir plumbing touching telemetry/resume/permissions; worth a focused reviewer pass that *every* path which previously assumed `~/.claude` now honors the override, not just the ones in the diff.
- Masking is a key-name heuristic (`KEY|TOKEN|SECRET|PASSWORD|CREDENTIAL`); a secret stored under a non-matching key (e.g. `MY_PAT`) would surface unmasked on `fleet.json`. Fine as best-effort, but worth documenting the limitation so users don't assume all values are masked.

chaitanyagiri:
Quick status: not forgotten — per-agent environment variables stayed out of v0.3.4/v0.3.5 deliberately (those were a community-integration wave and a fixes/content release respectively). It's queued for an upcoming feature release. Thanks for the patience!

chaitanyagiri:
## TL;DR

Thanks @ajishab, and sorry this has sat so long — opened 11 July, and it's the oldest PR in the queue.

@chaitanyagiri already left a full peer review on 24 July, so I won't repeat it. What I've added instead is the thing that review explicitly asked for and couldn't do statically: **an audit of whether every path that assumes `~/.claude` now honours a `CLAUDE_CONFIG_DIR` override** — run against *today's* `main`, not July's. I also ran the gates rather than reading them.

**Two findings that change the picture:**

1. **The answer to that open concern is now "no, and it got worse."** Since this PR was written, `src/main/skills.ts` landed (the v0.4.4 Skills browser) and hardcodes `join(homedir(), '.claude', 'skills')` in **five** places, plus one more in `index.ts`. None of them are in this diff, because they didn't exist when it was written.
2. **The branch is 174 commits behind `main` and conflicts in five files**, including `src/main/transcript.ts` — which is exactly the config-dir plumbing the peer review flagged as the highest-risk surface.

**Verdict:** no malicious code, no NSFW, no prompt injection. Typecheck, tests and `npm run build` are all green *on the branch as it stands*. The design is good and the security posture is genuinely strong. But this can't merge as-is, and the rebase is substantial.

---

## What I did

Checked out `worktree-agent-env-vars-105`, ran typecheck, the full test directory and a production build, measured the gap to `main` and the merge conflicts, then grepped current `main` for every remaining `~/.claude` assumption to answer the peer review's open question with data.

## Finding 1 — the `~/.claude` audit

The 24 July review asked for *"a focused reviewer pass that every path which previously assumed `~/.claude` now honors the override."* Here is that pass, against `main` at `d02c48c`:

| Site on current `main` | In this PR's diff? |
|---|---|
| `src/main/skills.ts:135` — user skills scan | ❌ file didn't exist |
| `src/main/skills.ts:141` — project skills scan | ❌ |
| `src/main/skills.ts:356` — skills install root | ❌ |
| `src/main/skills.ts:435`, `:439` — skill roots | ❌ |
| `src/main/index.ts:3241` — `skillRoots` for the tools/prereq panel | ❌ |
| `src/main/config.ts:717` — `ensureClaudePermissionsAccepted` | ⚠️ handled, but the function has moved |
| `src/main/transcript.ts:44`, `:80`, `:106` — `~/.claude/projects` | ⚠️ handled, but **this file conflicts** |
| `src/main/hive.ts:578` — per-agent `.claude/skills` | ✅ n/a, agent-dir relative |

So a profile agent on `CLAUDE_CONFIG_DIR=~/.claude-personal` would today run the right account, report the right tokens (once rebased) — and then be shown **the wrong machine's skills**, from `~/.claude/skills`, in the Skills browser and the Prerequisites panel.

That isn't a flaw in your design; it's what happens when a cross-cutting feature waits five weeks. It does mean the "follow-through" list in the PR description needs a fourth item, and it's an argument for centralising the resolution — one `claudeConfigDir(agentId)` helper that every call site goes through — so the next subsystem to arrive can't quietly reintroduce the assumption.

## Finding 2 — the branch has drifted a long way

```
  174 commits behind main

  conflicts:
    src/main/index.ts                                   ← spawn path
    src/main/transcript.ts                              ← the config-dir plumbing
    src/renderer/src/components/SettingsModal.tsx
    src/renderer/src/components/CommandCenterPanel.tsx
    src/renderer/src/components/AgentStrip.tsx
```

For scale: the test directory had **8** files when this branch was cut; `main` now has 39. `test/agent-env.test.cjs` passes here, but it's running against a suite a fifth the current size, so "green" means much less than it did in July.

**There's also a collision worth knowing about before you rebase.** PR **#157** rewrites the same `pty.spawn` env construction to strip inherited Claude Code session markers from `process.env`. That PR and this one both restructure how the child env is assembled. On #157 I suggested the layering should be `inherited (stripped) → app defaults → per-agent`, which is exactly the shape your merge-precedence rule already describes (*"internal injections always win over user env"*). If #157 lands first it does the stripping and you plug into it; if this lands first, #157 has to plug into yours. Either order works — it just shouldn't be discovered during a merge.

## Finding 3 — the requested cleanup hasn't happened

@chaitanyagiri asked on 18 July to *"clean it up a bit (remove the markdown files etc)"*. The diff still carries `docs/superpowers/plans/…` (990 lines) and `docs/superpowers/specs/…` (130 lines) — **1,120 of the PR's 1,813 added lines are those two documents**, and the repo has no `docs/superpowers/` directory otherwise. Removing them takes the reviewable diff from ~1,800 lines to under 700, which is a different PR to review. Worth doing in the same pass as the rebase.

## What I tested

| Check | Result |
|---|---|
| `node --test test/*.test.cjs` | ✅ all pass — but only **8** test files existed at this commit |
| Env can't be smuggled in from untrusted input | ✅ confirmed — hire manifests ignore `env`, and `pty:spawn` strips `agentEnvApplied`/`internalEnv` off renderer payloads before validation |
| Denylist | ✅ `PATH`, `NODE_OPTIONS`, `ELECTRON_RUN_AS_NODE`, `DYLD_*`, `LD_*`, `AGENT_*`, `HIVE_*`, `CTH_*`, `__proto__`, case-insensitive. A bad key **fails the spawn** rather than dropping silently — the right call |
| Mergeability | ❌ **CONFLICTING** — five files, 174 commits behind |

On the peer review's masking point (`MY_PAT` wouldn't match the `KEY|TOKEN|SECRET|…` heuristic): I'd go further than documenting it. Since `registry.json` stores values verbatim and `fleet.json` is read by god's own prompt, an unmasked secret doesn't just leak to a file — it lands in another agent's context. A short note in the Env vars editor UI ("values are masked in shared views only when the key looks secret-ish; prefer `*_TOKEN` / `*_KEY` names") would put the caveat where the user actually types.

## Questions

@ajishab — realistically, is a 174-commit rebase something you still have appetite for? Completely fine either way, and the work is good; I'd rather ask than leave it sitting another five weeks. If yes, the order I'd suggest is: (1) drop the two `docs/superpowers/` files, (2) rebase, (3) centralise the config-dir resolution into one helper, (4) route `skills.ts`'s six sites through it. If not, say so and the design doc in the PR body is enough for someone else to pick up — it shouldn't just expire.

@chaitanyagiri — you said in August this was queued for an upcoming feature release. Given the drift and the `skills.ts` interaction, is it still targeted, and does it want the `skills.ts` follow-through **in this PR** or as a follow-up? That answer decides how big the rebase actually is. Also worth deciding the #108 / #157 merge order before either author starts.

PR #110, opened 2026-07-23
fix(procs): release PIDs on every explicit kill path

qschmick:
# fix(procs): release PIDs on every explicit kill path

## What / why

Every explicit kill in the app was a bare node-pty `proc.kill()`, which sends
a single `SIGHUP` to the direct child only. This branch adds a small
`procKill.ts` helper (`ensureKilled(pid)`) and routes every explicit kill path
through it.

## The failure it fixes

`proc.kill()` signals only the immediate child. Two concrete leaks followed:

- a child that ignores/traps `SIGHUP` (common for TUIs and long-lived helpers)
  never dies — its PID is held for the machine's uptime;
- even when the direct child does die, *its* children (MCP servers, helper
  daemons) reparent to PID 1 and keep running.

Spawned-availability checks and killed sessions therefore accumulated live PIDs
across the whole run. `ensureKilled` sends a grace signal, then escalates to
`SIGKILL` of the entire process group (POSIX) / `taskkill /T /F` (Windows). The
pty child is a session leader, so its group covers its descendants.

Applied to `PtyManager` `kill` / `killAll` / `killByOwner`, the hidden-Claude
availability check sessions, and the mempalace miner (which also gains a hard
timeout — a wedged mine previously held its PID forever *and* silently stopped
all future mining passes). mempalace search / wake-up recall is converted from
`spawnSync` (a 120s main-thread freeze on a cold model load) to async `spawn`
with the same timeout + group sweep.

Natural exits are deliberately **not** swept: a daemon an agent intentionally
started must survive the session.

## Test coverage

`test/proc-kill.test.cjs` — self-contained, no framework (`node
test/proc-kill.test.cjs`). Builds a real process tree with a `SIGHUP`-trapping
session leader + a live child, proves the leak on the old path (leader
survives) and the fix on the new one (whole group reaped, zero survivors).
POSIX assertions; win32 exits 0 after a smoke import. Passes locally.

## Notes

Standalone, clean cherry-pick. No partial extraction.

### DIFF
--- a/src/main/hiddenClaude.ts
+++ b/src/main/hiddenClaude.ts
@@ -3,6 +3,7 @@ import { existsSync, readdirSync, readFileSync, statSync } from 'node:fs';
 import path from 'node:path';
 import { resolveCommand, userShellPath } from './shellEnv';
 import { projectDir } from './transcript';
+import { ensureKilled } from './procKill';
 
 /**
  * Shared helper: run a HIDDEN interactive claude session (ephemeral PTY) and
@@ -150,7 +151,14 @@ export function runHiddenClaude(prompt: string, opts: HiddenClaudeOptions): Prom
     let bootMaxTimer: NodeJS.Timeout;
     let globalTimer: NodeJS.Timeout;
 
-    const kill = () => { try { ptyProc.kill(); } catch { /* noop */ } };
+    // Hidden sessions are ephemeral CHECKS — nothing they spawn (MCP servers,
+    // helpers) may outlive them. Kill politely, then sweep the process group so
+    // every check releases its PIDs even if `claude` shrugs off the SIGHUP.
+    const kill = () => {
+      const pid = ptyProc.pid;
+      try { ptyProc.kill(); } catch { /* noop */ }
+      ensureKilled(pid);
+    };
 
     const finish = (r: HiddenClaudeResult) => {
       if (settled) return;
--- a/src/main/memory.ts
+++ b/src/main/memory.ts
@@ -16,6 +16,7 @@
 import { existsSync, statSync, readdirSync, readFileSync, writeFileSync } from 'node:fs';
 import { join } from 'node:path';
 import { spawn, spawnSync } from 'node:child_process';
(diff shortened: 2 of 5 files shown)


chaitanyagiri:
**Peer review — release PIDs on every explicit kill path**

Replaces bare node-pty `proc.kill()` (SIGHUP to the direct child only) with a new `procKill.ts` helper: `ensureKilled(pid)` waits a grace, then SIGKILLs the whole process group (POSIX) / `taskkill /T /F` (Windows), reaping descendants that reparented to PID 1. Wired into `PtyManager` `kill`/`killAll`/`killByOwner`, the hidden-Claude check sessions, and the mempalace miner (now with a 10-min hard timeout). Also converts `MemoryManager.search`/`wakeUp` from `spawnSync` (a 120s main-thread freeze on a cold model load) to async `spawn` with the same timeout + group sweep. Natural exits are deliberately not swept. Ships a self-contained `test/proc-kill.test.cjs`.

**Strengths**
- Correct root-cause fix: a SIGHUP-trapping leader or orphaned grandchildren really do leak PIDs, and group-SIGKILL of the session leader is the right mechanism. The deliberate "don't sweep natural exits" scope protects intentionally-backgrounded daemons.
- The test builds a real HUP-trapping tree and proves both the leak (bare SIGHUP survives) and the fix (whole group reaped) — genuinely convincing, not a smoke test. Moving the recall CLIs off `spawnSync` is a nice unblocking of the main thread on top.

**Concerns / suggestions**
- `search`/`wakeUp` change from returning a value to returning a `Promise` — a breaking signature change. This diff doesn't touch the callers, so please confirm every caller is updated to `await` (an un-awaited caller would now get a Promise where it expected `{ok, output}`). Worth calling out explicitly since it's the one behavior change hiding in an otherwise-additive PR.
- `ensureKilled` fires an unref'd timer with no handle, so a very fast app quit could exit before the grace elapses; that's acceptable (the OS reaps on exit), just noting the sweep is best-effort during shutdown.

PR #111, opened 2026-07-23
perf(usage): incremental per-file transcript cache

qschmick:
# perf(usage): incremental per-file transcript cache

## What / why

`readAgentUsage` re-read and re-`JSON.parse`d **every** transcript `.jsonl` in
the project dir on each call. The breaker/cost beat calls it roughly every 30s
per agent whenever there's no live OTel session. This branch makes the read
incremental.

## The failure it fixes

On a 40MB project dir the full re-parse was ~70ms of main-thread stall per agent
per beat, and a multi-agent floor paid it ×N. Transcripts are append-only JSONL,
so the totals for a byte range that was already parsed never change. The cache
keys per `(dir | file | session)` with size + mtime validation and parses only
the appended tail — the stored offset always lands on a newline boundary, so a
torn final write is counted exactly once. A rewrite/shrink triggers a full
re-parse. Warm read drops to ~0.2ms.

## Test coverage

`test/transcript-usage.test.cjs` — written against the OLD implementation first,
then held green across the rewrite. Covers appends, torn lines, rewrites,
deletes, and session-id filters (`node test/transcript-usage.test.cjs`). Passes
locally.

## Notes

Standalone, clean cherry-pick. No behavior change — only the cost of computing
the same totals.

### DIFF
--- a/src/main/transcript.ts
+++ b/src/main/transcript.ts
@@ -114,13 +114,120 @@ export interface ReadUsageOptions {
   sessionId?: string;
 }
 
+/** Per-file incremental parse state for the usage cache. Transcripts are
+ *  append-only JSONL, so a parsed byte range's totals never change — repeat
+ *  reads only stat the file and parse the appended tail. Keyed by
+ *  dir|file|sessionFilter since the filter changes what a range sums to. */
+interface FileUsageEntry {
+  size: number;
+  mtimeMs: number;
+  /** Bytes parsed so far — always ends on a newline boundary, so a torn
+   *  trailing line is simply re-read once the writer completes it. */
+  offset: number;
+  totals: AgentUsage;
+}
+
+const usageCache = new Map<string, FileUsageEntry>();
+/** Soft bound; when crossed, the oldest-inserted half is dropped (entries
+ *  rebuild on demand). Real fleets have tens of transcripts, not thousands. */
+const USAGE_CACHE_MAX = 2048;
+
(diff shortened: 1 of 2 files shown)


chaitanyagiri:
**Peer review — incremental per-file transcript usage cache**

`readAgentUsage` previously re-read and re-`JSON.parse`d every transcript `.jsonl` in the project dir on each call (~70ms/agent/beat on a 40MB dir, ×N agents, every ~30s). This adds a per-`(dir|file|session)` cache validated on size+mtime that parses only the appended tail; the stored offset always lands on a newline boundary so a torn final line is counted exactly once, a shrink/rewrite forces a full re-parse, and a soft 2048-entry bound drops the oldest half. Warm read drops to ~0.2ms.

**Strengths**
- Correctly exploits the append-only JSONL invariant — the newline-boundary offset handling for torn writes is exactly right and is directly tested.
- Excellent test discipline: the suite was written against the *old* implementation and held green across the rewrite, covering appends, torn lines, rewrites, deletes, and session filters. Same totals, cheaper.

**Concerns / suggestions**
- The cache key includes `sessionId`, so alternating filtered/unfiltered calls for the same file keep parallel entries and each re-parses the tail independently — bounded and fine given real callers, just worth a comment so it isn't mistaken for a bug later.
- mtime+size validation is safe here because a growing file always changes size; the only theoretical miss (same size *and* same mtime tick after a rewrite) can't happen for appends. Solid, well-scoped cherry-pick.

Issue #109, opened 2026-07-15
Circuit breaker fires on idle agents and on compaction / inbox-ack token bursts (v0.3.3) — spend-based detection can't tell housekeeping from looping


spid3rstan:
# Circuit breaker fires on idle agents and on compaction / inbox-ack token bursts (v0.3.3) — spend-based detection can't tell housekeeping from looping

**Type:** Bug
**Version:** v0.3.3 (latest), Windows 11, Claude Code agent engine
**Related:** #57 (fixed the *assistant-agent* case; this is the broader case), #56 (idle-agent ledger duplicates), #59 (open — skip checkpoint for idle agents)

## Summary

On v0.3.3 the cost/runaway circuit breaker repeatedly fires on agents that are demonstrably **not** looping. #57 added an `isAssistant` skip to the breaker's iteration loop, which fixed inert *assistant* agents. But the same misclassification still happens for **ordinary idle agents** and for **any agent during a `/compact` or inbox-ack**, because the detector keys on **token spend**, and spend cannot distinguish four different things that all emit tokens:

1. real work, 2. a runaway loop, 3. a `/compact` summary, 4. the breaker's own ack traffic.

## Evidence (one ~13h session, 4 agents)

- **12 breaker events, 12 false positives, 0 real loops caught.**
- An **idle agent with no dispatched task, no open card, and zero artifacts** was escalated to `steering`, then returned to `healthy` an hour later **with no intervention and no work done in between** — the alarm raised and cleared while the thing it claims to measure never changed.
- Multiple firings hit agents that were **fully stopped under a stand-down order**; the only artifact one produced all session was the **ack the breaker itself demanded**.
- Firings recurred **hourly, on the same minute** — matching the **scheduled auto-compaction** introduced in v0.2.7. The harness schedules the compaction whose token output then trips its own breaker.

## Root cause (hypothesis)

The breaker measures the wrong quantity. Spend is not a proxy for looping: compaction, inbox processing, and the breaker's own ack messages all raise the token counter identically to a runaway loop. #57's `isAssistant` guard only exempts one *category of agent*; it does not exempt these *categories of activity* on normal agents.

## Proposed fix

**Distinguish on OUTPUT, not on SPEND.**
- An agent that is *working* produces a **new artifact** in the window — a file written/edited, a task card moved, a changed test/pass count.
- An agent that is *looping* emits tokens and produces **no new artifact**.
- Only escalate when spend is high **and** no artifact was produced in the window.

[...]

PR #112, opened 2026-07-23
fix(breaker): eliminate the no-progress false-positive storm (#109)

qschmick:
# fix(breaker): eliminate the no-progress false-positive storm (#109)

## What / why

The circuit breaker's no-progress arm fired constantly on healthy agents — in a
live 8-agent floor ~10% of all hive messages were breaker steers. This branch
carries the #109 fix plus a small follow-on that closes the last observed
residual false positive.

## The failures it fixes

**Commit 1 — the #109 fix.** Three exemptions for the no-progress arm:

- *compaction exemption*: `PreCompact` opens a grace window (closed by
  `PostCompact`/`SessionStart`, with a 5-min safety cap) so `/compact` and the
  auto-compact mission's token burst can't trip the Δoutput arms — the exact
  self-inflicted FP reported in #109.
- *recent DISTINCT tool calls count as progress*: background workflows and
  interactive sessions do real work that never touches hive coordination files.
  A genuine single-call loop never refreshes this clock, and the loop/velocity
  arms still backstop.
- inbox/`.done` and the outbox dir now count as coordination mtimes (handling
  mail *is* coordination), and the arm is debounced to 2 consecutive beats so a
  one-beat blip never steers.

**Commit 2 — residual FP + a hot-path allocation.** Two hunks:

- `runBreakerBeat` now treats a recent OTel tool span as progress alongside
  fresh coordination-file mtimes. Background Workflow/subagent tool calls never
  reach the parent session's `PostToolUse` hook (so the breaker's own
  distinct-tool clock stays stale), but their spans do flow through the
  collector under the parent's id via `telemetry.getSpans(id)` — an idle parent
  supervising a busy background fleet is progressing, not wedged.
- `toolKey` uses a truncating JSON replacer (cap each string field at 250
  chars). It runs synchronously in the hook reply path on every `PostToolUse`,
  and a Write/Edit `tool_input` carries the whole file body (up to MBs), so
  serializing it all just to keep 200 chars was a multi-MB transient allocation
  per large write. Key semantics are unchanged.

## Test coverage

`test/breaker.test.cjs` (`node test/breaker.test.cjs`) — 16 cases including
regression guards for the loop / error-storm / velocity trips, the compaction
grace window, and the two new huge-input toolKey cases (loop arm still fires on
identical huge Writes; huge inputs differing early still count as distinct).
Passes locally.

The second commit was originally bundled inside a larger main-process commit on
our fork. Only the two hunks above were extracted here (the `runBreakerBeat`
span check and the `toolKey` replacer, plus their two test cases). The span
change references `telemetry.getSpans`, which already exists upstream and
returns `ToolSpan[]` with a `ts` field. Typechecks cleanly.

### DIFF
--- a/src/main/breaker.ts
+++ b/src/main/breaker.ts
@@ -73,6 +73,19 @@ const DEFAULTS = {
   tokenVelocityPerMin: 60_000 // output tokens/min — coarse backstop, deliberately high
 };
 
+/** Safety cap on the PreCompact exemption: if PostCompact never arrives (crash,
+ *  or a Claude build that doesn't emit it), the Δoutput trips re-arm on their own. */
+const COMPACT_GRACE_MS = 5 * 60_000;
+/** Trailing grace after PostCompact: the compaction burst lands in the NEXT
+ *  beat's cumulative diff, so the exemption must outlive the compaction itself. */
+const POST_COMPACT_GRACE_MS = 90_000;
+/** How recent a DISTINCT tool call must be to count as progress for the
+ *  no-progress arm. Mirrors the beat's file-mtime progress window (300s). */
+const PROGRESS_TOOL_WINDOW_MS = 300_000;
+/** Consecutive tripping beats the no-progress arm needs before it fires — a
+ *  one-beat blip (inbox ack, statusline burst) never steers on its own. */
+const NO_PROGRESS_BEATS = 2;
+
 interface AgentBreakerState {
   level: BreakerLevel;
   reason: string;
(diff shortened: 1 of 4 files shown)


chaitanyagiri:
**Peer review — eliminate the breaker no-progress false-positive storm (#109)**

Adds three exemptions to the breaker's no-progress arm: a `PreCompact`→`PostCompact` grace window (5-min safety cap + 90s trailing grace) so the compaction token burst can't trip the Δoutput arms; recent *distinct* tool calls count as progress; and inbox/`.done` + outbox mtimes now count as coordination, with a 2-beat debounce so a one-beat blip never steers. A follow-on commit also treats a recent OTel tool span as progress (so an idle parent supervising a busy background fleet doesn't read as wedged) and truncates each string field to 250 chars in `toolKey` to avoid a multi-MB transient allocation when hashing large Write/Edit inputs on the hook hot path.

**Strengths**
- Attacks a measured problem (~10% of hive traffic being breaker steers); each exemption is well-reasoned and still backstopped by the loop/error/velocity arms, so the safety net isn't removed, just the false arm.
- The `toolKey` truncation is a nice real catch — serializing an entire file body to keep 200 chars was pure waste on every `PostToolUse`.
- Good regression coverage (16 cases, incl. the compaction window and the huge-input `toolKey` cases).

**Concerns / suggestions**
- The grace caps widen the window in which a genuinely stuck *compaction-adjacent* agent goes unsteered by this arm — acceptable given the other arms run, but the no-progress arm is now the weakest by design; worth a comment stating that explicitly.
- The span-as-progress path depends on `telemetry.getSpans(id)` surfacing background subagent spans under the parent's id — worth confirming that attribution holds in the shipped build, since it's the linchpin of that exemption.

PR #113, opened 2026-07-23
fix(renderer): never type into a prompted agent; stop the queue drain destroying mail

qschmick:
# fix(renderer): never type into a prompted agent; stop the queue drain destroying mail

## What / why

Two adversarially-verified interaction bugs in the renderer's auto-typers. The
first is a safety issue — programmatic typing could confirm a tool-permission
prompt — and is the strongest reason to take this branch.

## The failures it fixes

**SAFETY — typing could land on a permission prompt.**
The inbox-wake nudge fired on agents in status `idle` **or** `waiting`.
`waiting`/`blocked` mean a real approval/notification prompt is on screen (the
hook `Notification` needs-human path; `BLOCK_HINTS` like `❯ 1. Yes` detected by
`usePtyParser`). `submitToPty` always follows its text with a bare `\r`, and at
a Claude Code menu that Enter **silently confirms** the highlighted "1. Yes" on
the pending tool approval. The nudge is now gated to strictly `idle`; it simply
defers (it stamps `nudged` only when it actually types) and fires once the
prompt clears and the agent drifts idle. The Crush seed typer had the same
trailing-Enter hazard while its TUI booted — it now re-queues the seed and
retries on a later tick if the agent is `waiting`/`blocked`, and skips a
vanished (killed-mid-boot) agent entirely. (Queue drain was already idle-only.)

**RELIABILITY — the queue drain silently destroyed mail.**
`window.cth.writePty` never rejects for a dead pty; it *resolves*
`{ ok: false, error }`. `submitToPty` awaited it unchecked, so a failed delivery
looked successful. Combined with the drain removing a message from the queue
*before* writing it, one message was silently destroyed per cooldown for any
crashed-but-`idle` agent. Now `submitToPty` throws on `!ok`, and the drain
removes a message only *after* its write chain resolves — an `inFlightSends`
guard prevents double-sends, and a 3-strike drop `console.warn`s the lost
message (naming it) instead of vanishing it.

## Test coverage

None added — these are timing/interaction paths against the live PTY and store,
not currently covered by the repo's node test harness. Verified by reading
through the state machine; typechecks cleanly. Manual review of the two gates is
worthwhile.

Extracted from a larger renderer commit on our fork. This branch is a
single-file diff (`src/renderer/src/hooks/useHive.ts`) and deliberately excludes
the unrelated fixes from that commit (terminal disposal, dismissed-agent
tombstones, the inbox-listing perf change). The inbox-wake hunk keeps the
existing `hiveInbox`-based newest-id logic — only the strictly-idle gate is
changed here.

### DIFF
--- a/src/renderer/src/hooks/useHive.ts
+++ b/src/renderer/src/hooks/useHive.ts
@@ -91,9 +91,16 @@ function submitToPty(ptyId: string, text: string, settleMs = 250): Promise<void>
     // commands) is sent raw — some TUIs (Antigravity's agy) treat the paste
     // markers as literal input and never submit, so skipping them is more robust.
     const payload = text.includes('\n') ? `\x1b[200~${text}\x1b[201~` : text;
-    await window.cth.writePty(ptyId, payload);
+    // writePty NEVER rejects for a dead pty — it resolves { ok:false, error:
+    // 'no pty: …' } — so an unchecked await here made every failed delivery look
+    // successful (the queue-drain then destroyed the message it had already
+    // popped, #36). Surface the failure as a rejection; the chain itself is
+    // immune (the prev.catch above absorbs it for the next writer).
+    const wrote = await window.cth.writePty(ptyId, payload);
+    if (!wrote?.ok) throw new Error(wrote?.error ?? `pty write failed: ${ptyId}`);
     await new Promise((r) => setTimeout(r, 140));
-    await window.cth.writePty(ptyId, '\r');
+    const submitted = await window.cth.writePty(ptyId, '\r');
+    if (!submitted?.ok) throw new Error(submitted?.error ?? `pty write failed: ${ptyId}`);
     await new Promise((r) => setTimeout(r, settleMs));
   });
   writeChains.set(ptyId, next);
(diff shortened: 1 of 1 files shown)


chaitanyagiri:
**Peer review — never type into a prompted agent; stop the queue drain destroying mail**

Two renderer auto-typer fixes in `useHive.ts`. (1) **Safety:** the inbox-wake nudge is now gated to strictly `idle` (was `idle || waiting`), and the Crush seed typer re-queues + retries if the agent is `waiting`/`blocked` and skips a vanished agent — because `submitToPty` always follows its text with a bare `\r`, and at a Claude Code permission menu that Enter silently confirms the highlighted "1. Yes". (2) **Reliability:** `submitToPty` now throws on `!ok` (`writePty` resolves `{ok:false}` for a dead pty rather than rejecting), and the drain removes a message only *after* the write chain resolves, with an `inFlightSends` dedupe guard and a 3-strike drop that `console.warn`s the lost message instead of silently destroying it.

**Strengths**
- The safety fix is the headline and genuinely matters — programmatically confirming a tool-permission prompt is a real hazard, and gating on `idle` + deferring (stamping `nudged` only when it actually types) is the right shape.
- The remove-*after*-confirm reordering correctly closes the silent one-message-per-cooldown data loss; `inFlightSends` + bounded retries are the right primitives, and the drop is loud and names the message.

**Concerns / suggestions**
- `lastFlush` is stamped before the async write resolves, so a fast-failing head still burns the full cooldown — 3 strikes take ~3×4.5s before the drop. Intended and safe, just noting the latency.
- No automated tests (acknowledged — live-PTY/store timing paths). The two gates are worth a manual pass, as the author suggests; a small state-machine unit test around `dispatch`'s in-flight/failure bookkeeping would be a good future add.

PR #114, opened 2026-07-23
perf(pty): memoize command resolution + reuse the cached shell PATH on spawn

qschmick:
# perf(pty): memoize command resolution + reuse the cached shell PATH on spawn

## What / why

Agent spawns paid a few seconds of synchronous interactive-shell launches on
the main process — freezing every window — that were entirely avoidable. This
branch caches both.

## The failure it fixes

- `resolveCommand` ran `$SHELL -ilc which <cmd>`, which sources the user's whole
  `zshrc` (nvm/asdf init is routinely ~1s), and every spawn hit it **twice**:
  once in the `isCommandAvailable` pre-check and once in `spawn()` — ×N on a
  team restore. Successful resolutions are now cached for the session (the
  method is split into a thin `resolveCommand` cache wrapper over
  `resolveCommandUncached`). A positive hit is revalidated with `existsSync`, so
  an uninstall/update between spawns re-probes; negatives are deliberately not
  cached, so the missing-CLI auto-install re-check still sees a just-installed
  binary.
- the per-spawn `$SHELL -ilc echo $PATH` block is replaced with the already
  session-cached `shellEnv.userShellPath()`, which the codebase already uses for
  headless child launches — `pty.ts` was the last inline copy.

Results are byte-identical to before; they're just computed once.

## Test coverage

None added — this is pure memoization over existing, already-tested resolution
logic, and the results are unchanged. Typechecks cleanly.

Extracted from a larger main-process commit on our fork; this branch is a
single-file diff (`src/main/pty.ts`) and depends only on the existing
`shellEnv.userShellPath()` already present upstream.

### DIFF
--- a/src/main/pty.ts
+++ b/src/main/pty.ts
@@ -2,6 +2,7 @@ import * as pty from 'node-pty';
 import type { WebContents } from 'electron';
 import { existsSync } from 'node:fs';
 import { spawnSync } from 'node:child_process';
+import { userShellPath } from './shellEnv';
 
 interface PtySession {
   id: string;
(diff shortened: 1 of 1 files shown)


chaitanyagiri:
**Peer review — perf(pty) command-resolution + PATH memoization**

Caches successful `resolveCommand` results (split into a thin cache wrapper over `resolveCommandUncached`, revalidated with `existsSync` on every hit, negatives intentionally not cached) and replaces the per-spawn `$SHELL -ilc echo $PATH` block with the session-cached `userShellPath()`. Single-file diff in `src/main/pty.ts`.

**Strengths**
- Targets a real cost: the double interactive-shell launch per spawn (pre-check + spawn) was a multi-second main-thread freeze, ×N on a team restore — a clean win.
- Correct on the subtle bits: not caching negatives keeps the missing-CLI auto-install re-probe working, and the `existsSync` revalidation re-probes after an uninstall/update between spawns.

**Concerns / suggestions**
- The cache is unbounded, but the command set is tiny and fixed, so that's fine — a one-line note on the `Map` would make that explicit.
- Edge case worth a thought: `existsSync(cached.path)` only checks the *old* path still exists, so a binary relocated to a new dir while a stale copy lingers at the old path would keep the stale resolution. Realistic only under uninstall+reinstall-elsewhere, so low risk.
- No tests, but this is pure memoization over already-tested logic with byte-identical results — acceptable here.

PR #123, opened 2026-08-10
fix(memory): read the transcript directory Claude Code actually writes

gts-47:
## The bug

`projectDir()` encodes a project's cwd using an older Claude Code convention: the leading slash is
**dropped** and the rest are dashed (`/Users/me/app` → `Users-me-app`). Claude Code now dashes the
leading slash too (`/Users/me/app` → `-Users-me-app`).

So every consumer of `projectDir()` reads a directory Claude Code never writes to. Nothing asserts the
directory exists, so an empty read looks like "no data" rather than "wrong path", and it fails silently.

## What it breaks

`MemoryReflector` calls `extractLastAssistantText`, which reads the transcript through `projectDir()`,
finds nothing, and aborts with `no assistant response found in transcript`. On this machine that is
**4039 condense-aborts and zero successes since 17 Jun 2026** — memory condensation has never once run.
The summarizer was never at fault: the hidden Haiku session's own transcript is present, valid, and in
the dashed directory the whole time.

Evidence from a live machine:

```
~/.claude/projects/-Users-...-HarnessAgents   1518 transcripts, mode 700, written today
~/.claude/projects/Users-...-HarnessAgents       4 stale files, mode 755, newest 4 Aug
```

The no-dash twin is mode 755 because *our own* Electron process created it via `mkdirSync`/`copyFileSync`
in `transcript.ts`. Both variants exist side by side for the root and for ~10 worktrees.

It also affects the offline usage reconciler (`readAgentUsage`) and cross-cwd session-resume seeding,
which writes into a directory Claude Code never reads.

## The fix

- `projectKey()` dashes every separator, including the leading slash.
- `legacyProjectKey()` keeps the old spelling.
- `projectDir()` prefers the current spelling, falls back to legacy **only** when legacy exists and the
  current one does not, and returns the current spelling when neither exists — so callers that create the
  directory create the one Claude Code will actually read.

Windows behaviour is unchanged.

## Tests

`test/transcript-project-dir.test.cjs`, four cases: an unseen cwd yields the dashed key; the current
spelling wins when both twins exist; legacy-only installs still resolve; and the exact path that was
failing. Hermetic — each case gets a throwaway `$HOME`, so the real `~/.claude` is never touched.

`test/load-ts.cjs` needed `esModuleInterop: true` to match `tsconfig.node`/`tsconfig.web`; without it a
default import of a CJS builtin compiles to `path_1.default`, which is undefined at run time. Test
harness only — no shipped code is affected.

`npm run typecheck` (node + web) exits 0. `npm run test:focused` passes 37/37.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "typecheck:node": "tsc --noEmit -p tsconfig.node.json",
     "typecheck:web": "tsc --noEmit -p tsconfig.web.json",
     "typecheck": "npm run typecheck:node && npm run typecheck:web",
-    "test:focused": "node --test test/auto-compact.test.cjs test/codex-remote.test.cjs test/control.test.cjs test/provider-automation.test.cjs test/provider-config.test.cjs test/queue-delivery.test.cjs test/terminal-automation.test.cjs test/terminal-recovery.test.cjs",
+    "test:focused": "node --test test/auto-compact.test.cjs test/codex-remote.test.cjs test/control.test.cjs test/provider-automation.test.cjs test/provider-config.test.cjs test/queue-delivery.test.cjs test/terminal-automation.test.cjs test/terminal-recovery.test.cjs test/transcript-project-dir.test.cjs",
     "dist": "npm run build && electron-builder",
     "dist:mac": "npm run build && electron-builder --mac",
     "dist:win": "npm run build && electron-builder --win",
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -147,8 +147,19 @@ function shortRand(): string {
 
 /** Non-memory files `mempalace mine` must not ingest (Claude Code hooks config,
  *  cursor, raw inbox/outbox JSON). `mempalace mine` honors .gitignore, so we drop
- *  one in each agent dir; written on birth here and refreshed by the mine loop. */
-const MINE_IGNORE_LINES = ['settings.json', 'cursor.json', 'inbox/', 'outbox/'];
+ *  one in each agent dir; written on birth here and refreshed by the mine loop.
+ *
+ *  `.codex/` is here for a second reason as well, and it is the load-bearing one:
+ *  a Codex worker's CODEX_HOME lives INSIDE its agent dir (see installCodexHooks —
+ *  Codex can only be given hooks through a config.toml in its own home, so it
+ *  cannot share the user's ~/.codex). Codex then fills that folder with full
+ *  session transcripts, an 80MB+ logs sqlite and a plugin cache, and the hive's
+ *  git repo was faithfully versioning every revision of all of it. Twenty Codex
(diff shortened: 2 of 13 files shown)


chaitanyagiri:
Thanks for this — I've read it through and the diagnosis reproduces on my machine as well: `condense-abort` / `no assistant response found in transcript`, with zero successes on record. The root cause you've identified is real and worth fixing.

This is **under review** and will be picked up in **v0.3.8**.

I have some notes on the project-key encoding and on the unrelated commits riding along on this branch — I'll follow up with those separately before it lands.

chaitanyagiri:
Following up as promised. Superseded by **#125**, which is this fix rebased onto current `main` — full credit to you, @gts-47: the investigation, the root cause, the resolver design with its legacy fallback, the `esModuleInterop` catch and the original tests are all your work, and the commit carries your `Co-Authored-By`.

Two reasons for a fresh PR rather than iterating here:

**1. The project key needed one more correction.** `cwd.replaceAll('/', '-')` dashes the leading slash but nothing else. Claude Code dashes *every* non-alphanumeric — dots included — which is the rule this file's own Windows branch already used. Checked against `~/.claude/projects` on the maintainer's machine: all 30 live Claude-Code-written dirs are `[a-zA-Z0-9-]` with zero dots, while the stale harness-written twins keep theirs.

Running your resolver against real paths there:

| cwd | this PR returns | Claude Code writes | |
|---|---|---|---|
| `…/Documents/MDv0.3.0` | `Users-…-MDv0.3.0` | `-Users-…-MDv0-3-0` | ❌ |
| `/Users/vyapakgoyal/Documents/HarnessAgents` | matches | matches | ✅ |

It works for `HarnessAgents` because that path has no dots. Every project path on the maintainer's machine contains `MDv0.3.0`, so it would have changed nothing there — and the legacy fallback then finds the stale twin, so the miss returns a real directory and reads as a hit. A silent fix for a silent bug. Your four tests all used dot-free fixtures, which is why they passed; #125 adds four more that cover the dot rule.

**2. Scope.** This branch also carried five unrelated fixes, and its base was 204 commits behind `main`. #125 is the transcript fix alone.

To be clear about those five: **none are superseded** — I checked, all five are still absent from `main` — and they were dropped for scope, not merit. The `.codex/` gitignore in particular is a real fix for a 7.5GB-`.git`, OOM-the-machine problem. Please do reopen them as separate PRs; #125's description lists all five with notes on each, including two things worth a second look (the gitignore doesn't untrack what's already committed, and hand-spelling the Edit menu drops `pasteAndMatchStyle`, `delete` and the macOS Speech / Start Dictation / Emoji & Symbols items that `{ role: 'editMenu' }` supplies).

Closing in favour of #125. Genuinely good catch — this had been silently killing memory condensation, and nothing surfaced it.

PR #128, opened 2026-08-13
fix(hive): keep Codex worker homes out of the hive's git history

gts-47:
A Codex worker's `CODEX_HOME` lives **inside** its agent directory (`installCodexHooks` — Codex can only be given hooks through a `config.toml` in its own home, so it cannot share the user's `~/.codex`). Codex then fills that folder with full session transcripts, an 80MB+ logs sqlite and a plugin cache, and the hive's git repo faithfully versions every revision of all of it.

On my fleet, twenty Codex agents took the hive's `.git` to **7.5 GB**, at which point git's own auto-gc tried to repack it and took **22 GB of RAM** doing so. The machine swapped and the app stopped responding. Untracking brought the hive from 13,638 tracked files to 3,901.

None of it was ever wanted in history: it is Codex's private scratch state, and it stays on disk either way, so `codex --resume` is unaffected.

**Two commits, because the ignore alone is not enough:**

1. Add `.codex/` to `MINE_IGNORE_LINES` in **both** copies (`hive.ts`, written on spawn; `memory.ts`, written by the mine loop — only the latter reaches agents that are not currently running).
2. Untrack what is already committed. Adding a path to `.gitignore` does not stop git tracking a file it already tracks, so an existing hive keeps recording every revision exactly as before — the ignore reads as a fix while the repo goes on growing. Once per process, refresh every agent's ignore file and drop any tracked `agents/*/.codex` path from the index. Probed with `ls-files` first, so a clean repo does not rewrite its index on every launch.

### DIFF
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -191,8 +191,19 @@ function shortRand(): string {
 
 /** Non-memory files `mempalace mine` must not ingest (Claude Code hooks config,
  *  cursor, raw inbox/outbox JSON). `mempalace mine` honors .gitignore, so we drop
- *  one in each agent dir; written on birth here and refreshed by the mine loop. */
-const MINE_IGNORE_LINES = ['settings.json', 'cursor.json', 'inbox/', 'outbox/'];
+ *  one in each agent dir; written on birth here and refreshed by the mine loop.
+ *
+ *  `.codex/` is here for a second reason as well, and it is the load-bearing one:
+ *  a Codex worker's CODEX_HOME lives INSIDE its agent dir (see installCodexHooks —
+ *  Codex can only be given hooks through a config.toml in its own home, so it
+ *  cannot share the user's ~/.codex). Codex then fills that folder with full
+ *  session transcripts, an 80MB+ logs sqlite and a plugin cache, and the hive's
+ *  git repo was faithfully versioning every revision of all of it. Twenty Codex
+ *  agents took the hive's .git to 7.5GB, at which point git's own auto-gc tried to
+ *  repack it and took 22GB of RAM doing so — the machine swapped, the app stopped
+ *  responding. None of it was ever wanted in history: it is Codex's private
+ *  scratch state, and it stays on disk (so resume still works) either way. */
+const MINE_IGNORE_LINES = ['settings.json', 'cursor.json', 'inbox/', 'outbox/', '.codex/'];
 
 /** Idempotently ensure `<agentDir>/.gitignore` excludes the non-memory files.
  *  Append-only: writes only the missing lines, leaving any existing entries. */
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -1870,10 +1870,43 @@ export class HiveManager {
     return { ok: res.status === 0, out: res.stdout ?? '', err: res.stderr ?? '' };
   }
 
+  /** Has the one-time Codex-home untrack pass run in this process yet? */
+  private untrackedCodexHomes = false;
+
+  /**
+   * Stop versioning Codex worker homes that are ALREADY in the index.
+   *
+   * Adding `.codex/` to each agent's .gitignore only keeps NEW paths out; git
+   * happily keeps recording a file it is already tracking, so a hive that
+   * predates that ignore line goes on committing every SQLite and transcript
+   * revision exactly as before — the .gitignore reads as a fix while the repo
+   * keeps growing. This closes that: once per process, refresh every agent's
+   * ignore file (agents that are not running never pass through spawn, and the
+   * mine loop only reaches them if mempalace is installed) and drop any tracked
+   * `.codex` path from the index. The files stay on disk, so `codex --resume`
+   * is unaffected; only their history stops.
+   */
+  private untrackCodexHomes(root: string): void {
+    if (this.untrackedCodexHomes) return;
+    this.untrackedCodexHomes = true;
+    const agentsDir = join(root, 'agents');
+    if (!existsSync(agentsDir)) return;
+    try {
+      for (const id of readdirSync(agentsDir)) ensureMineIgnore(join(agentsDir, id));
+    } catch { /* best-effort */ }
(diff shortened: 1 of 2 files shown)


chaitanyagiri:
## TL;DR

Thanks @gts-47 — this is a genuinely important fix and the write-up is exactly the kind that makes a reviewer's job easy.

**What it does:** a Codex worker's `CODEX_HOME` lives **inside** its agent directory (Codex can only be given hooks via a `config.toml` in its own home, so it can't share `~/.codex`). Codex then fills that folder with full session transcripts, an 80MB+ logs SQLite and a plugin cache — and the hive's git repo faithfully versioned every revision of all of it. On your fleet: twenty Codex agents took the hive's `.git` to **7.5 GB**, at which point git's own auto-gc tried to repack it and consumed **22 GB of RAM**. The machine swapped and the app stopped responding.

The PR is two commits, and the second is the one that matters: adding `.codex/` to `.gitignore` does **not** stop git tracking files it already tracks, so an existing hive would keep growing while the ignore line reads like a fix. So it also untracks — once per process, refreshing every agent's ignore file and dropping tracked `agents/*/.codex` paths from the index.

**Verdict:** no malicious code, no NSFW, no prompt injection, no dependency changes. Typecheck green, 149/149 tests pass, both `MINE_IGNORE_LINES` copies are still unpatched on `main`. **Blocked only on staleness** — the branch is **75 commits behind** and conflicts in `src/main/hive.ts`. One real gap worth closing, and one missing test I'd push for.

---

## What I did

Checked out `pr/codex-home-out-of-git`, ran typecheck and the full suite, confirmed the bug is still live on `main`, read `ensureMineIgnore` and the commit path for ordering and failure modes, and tested the merge against current `main`.

## Why the second commit is the load-bearing one

```
   commit 1 only:                          commit 1 + 2 (this PR):

   .gitignore gains ".codex/"              .gitignore gains ".codex/"
        │                                       │
        │  git already tracks                   │  git rm -r --cached agents/*/.codex
        │  agents/x/.codex/**                   │       │
        ▼                                       ▼       ▼
   `git add -A` keeps committing          index drops them; add -A won't re-add
   every sqlite + transcript revision      (files stay on disk → `codex --resume` fine)
        │                                       │
        ▼                                       ▼
   .git keeps growing, silently            growth stops
```

I checked the ordering in `commit()`: `untrackCodexHomes(root)` runs **before** the retry loop that does `git add -A`, so the staged deletions can't be immediately re-added. The `ls-files` probe before mutating is the right instinct too — without it, `rm --cached` would rewrite the index on every launch inside the commit retry path. And `ensureMineIgnore` internally try/catches both its read and its write, so a stray `.DS_Store` in `agents/` can't abort the refresh loop. That all holds up.

## What I tested

| Check | Result |
|---|---|
| Bug still present on `main` | ✅ both copies still read `['settings.json', 'cursor.json', 'inbox/', 'outbox/']` — `hive.ts:196` and `memory.ts:25` |
| Ordering (untrack before `add -A`) | ✅ correct |
| `ensureMineIgnore` failure safety | ✅ fully guarded; a non-directory entry is a harmless no-op |
| Mergeability | ❌ **CONFLICTING** — `src/main/hive.ts`. Branch is **75 commits behind**, the largest gap of any open PR |

I can corroborate the general shape of the problem from the hive on this machine: **142 MB of `.git` for 1,036 tracked files.** No Codex agents here, so no `.codex` paths are tracked — but that ratio shows hive repos bloat readily even without the Codex case, which makes your fix more valuable, not less.

## The gap: existing users keep the 7.5 GB

Untracking stops the *growth*, but every object already written stays in `.git`. Your own numbers say tracked files went 13,638 → 3,901; the pack didn't shrink. So a user who hit this still has a multi-gigabyte `.git` and is still one auto-gc away from the 22 GB repack that took their machine down — which is the symptom they actually noticed.

That's arguably out of scope for this PR, and I'd rather it landed than grew. But it shouldn't go unrecorded. Options, roughly in order of how much I'd like them:

1. Log a one-time actionable hint next to the existing `console.warn` when untracking finds tracked Codex homes — `.git` is at N MB; history rewrite needed to reclaim it.
2. A `git gc --prune=now --aggressive` after the untrack. Reclaims loose objects but **not** packed history, so it's a partial measure.
3. A documented `git filter-repo` recipe in `HIVE.md`. Full reclaim, but destructive enough that it should never be automatic.

My vote is (1) here, and (3) as a docs follow-up.

## The missing test

The two `MINE_IGNORE_LINES` copies now carry a comment saying **"MUST STAY IN SYNC"**. That's exactly the kind of invariant that drifts the moment someone adds a line to one file — and the consequence is silent (only agents that aren't currently running are affected, so nobody notices until a repo bloats again). A four-line test asserting the two arrays are `deepEqual` would pin it permanently, and it's the cheapest possible insurance against a repeat of a bug that cost you 22 GB of RAM.

Longer term the duplication itself is the smell: one exported `MINE_IGNORE_LINES` in `src/shared/` imported by both would make the test unnecessary. Either is fine; the current state — two copies plus a comment — is the one I'd avoid.

## Questions

@gts-47 — (1) can you rebase? 75 commits is a lot, and `hive.ts` has moved substantially, so it's better done by you than resolved by someone guessing at intent. (2) Any objection to adding the two-copy sync test while you're in there? (3) Does `untrackCodexHomes` need to run on hives with no Codex agents at all — the `ls-files` probe makes it cheap, but it also calls `ensureMineIgnore` for every agent directory on the first commit of every process. Worth confirming that's negligible on a large fleet.

@chaitanyagiri — this one is worth prioritising: it's a **data-loss-adjacent, machine-wedging bug** with a confirmed real-world reproduction, and it's been open since 13 July. Two things for you: (a) do you want the reclaim story (option 1/3 above) in this PR or as a follow-up? (b) `CODEX_HOME` living inside the agent dir is the root cause of the whole class — is there a reason it can't live under the harness home *outside* the git-tracked hive, with only `config.toml` inside? That would make the ignore unnecessary rather than load-bearing. Happy to open that as a separate issue if it's worth exploring.

PR #129, opened 2026-08-13
fix(restart): let "Restart & Continue" work on an agent that already died

gts-47:
`restartWithModel` kills the current PTY and treats any failure from `killPty` as fatal, so when the process is **already gone** it throws `no pty: pty-god` and never respawns.

That is precisely the case the button exists for: the CLI crashed, the Mac slept, or the user hit Ctrl-C twice. The agent whose session you most want back is the one you cannot restart.

Treat "there was no pty to kill" as success and carry on to the resume spawn; every other kill failure still aborts.


chaitanyagiri:

Your change is live for everyone: "Restart & Continue" now works on an agent that has already died, instead of dead-ending on the one situation the button exists for.

PR #130, opened 2026-08-13
fix(inbox): stop one odd message id from silencing an agent's wake nudge

gts-47:
The inbox-wake nudge dedupes on the **lexicographically largest** id in an agent's inbox, read as "the newest". Ids are normally `<timestamp>-<rand>` so that holds — but an agent may set its own `id` in the outbox JSON and `normalize()` keeps it verbatim (`id: partial.id ?? …`, unchanged since inception).

My orchestrator's inbox has held `dev15-progress-canvas-v4` since Aug 10. Letters sort above digits, so that id outranks every `2026-*` timestamp and, because it never drains, it is permanently "the newest". Result: he was nudged once per app launch and then never again, while **45 real messages** piled up behind it unannounced. Agents finished work, reported back, and he never heard. Measured across his inbox: 34 of 1611 messages carry a hand-written id — 2%, which is all it takes.

It self-healed in the past only because the offending message eventually got moved to `.done`.

Track the ids already nudged for instead of a high-water mark. No ordering assumption, and it keeps what the high-water mark was there for: a set that only shrinks as messages drain contains nothing new, so a drain stays quiet.

### DIFF
--- a/src/renderer/src/hooks/useHive.ts
+++ b/src/renderer/src/hooks/useHive.ts
@@ -253,10 +253,20 @@ function passesContextPressure(a: Agent, rule: ContextRule): boolean {
  *      doesn't stall while an agent sits at its prompt.
  */
 export function useHive(config: HarnessConfig | null): void {
-  // Per-agent dedup key for the inbox-wake nudge: the newest inbox message id we
-  // last nudged about. Keyed by id (not count) so an oscillating count after a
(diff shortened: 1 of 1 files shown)


chaitanyagiri:

Your change is live for everyone: a single odd message id can no longer silence an agent's wake nudge permanently.

PR #131, opened 2026-08-13
fix(paste): paste what was just dictated, not the clipboard's previous text

gts-47:
Dictation tools (Muesli, Wispr Flow, etc.) work by stashing the clipboard, writing the transcript into it, sending a paste, then **restoring the old clipboard immediately**. Anything that reads the clipboard asynchronously loses that race and pastes the previous text.

Two independent late reads, both fixed:

1. **Terminal paste** — `pasteClipboard` awaited `window.cth.readClipboard()`, an async IPC round-trip. By the time it resolved the tool had already restored the old text. Added a synchronous `app:readClipboardSync` (`ipcMain.on` + `event.returnValue`) and read through it, keeping the async path as a fallback.
2. **The app-wide Cmd+V** — `{ role: 'editMenu' }` registers Cmd+V as a **menu accelerator**, and Electron replays it via `webContents.paste()` on the next tick. That fires app-wide, including over the Queue composer, and loses the same race. Replaced with an explicit Edit submenu whose clipboard items carry `registerAccelerator: false`, so the keystroke is left to the focused element and the menu items still work from the menu bar.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -2269,7 +2269,34 @@ function installAppMenu(): void {
         ? [newFloorItem, { type: 'separator' as const }, { role: 'close' as const }]
         : [newFloorItem, { type: 'separator' as const }, { role: 'quit' as const }]
     },
-    { role: 'editMenu' },
+    // The Edit menu is spelled out rather than `{ role: 'editMenu' }` for one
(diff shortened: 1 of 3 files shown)


chaitanyagiri:

Your change is live for everyone: dictation pastes what you just said, not whatever the clipboard held before it.

PR #132, opened 2026-08-13
fix(transcript): never resolve a root cwd to the projects directory itself

gts-47:
For cwd `/` the legacy key is the empty string, and `path.join(root, '')` collapses to `~/.claude/projects` — the directory holding **every** project, which always exists. The legacy fallback therefore returns it, and callers go on to read and seed `~/.claude/projects/<session>.jsonl` instead of the real per-project location.

An empty key is not a project name; treat it as no legacy candidate at all. Covered by a new test.

Found by an independent review pass over my branch. Typecheck clean, 131/131 focused tests pass.


chaitanyagiri:

Your change is live for everyone: a root cwd is never resolved to the projects directory itself.

PR #134, opened 2026-08-13
perf(office): stop rendering the floor nobody is looking at

gts-47:
A fullscreen terminal or file editor covers the office floor completely, and a hidden window shows nothing at all — but the Pixi ticker goes on running the whole scene regardless: every character, thought cloud, coffee run and envelope animating, and the renderer drawing every frame, into pixels nobody can see.

On a floor of twenty agents that is the app's single largest continuous cost. Working the normal way with one agent — living in the fullscreen terminal — pays it 100% of the time, which is a real hit on a busy machine (I chase paste races that need the renderer to answer a keystroke inside 500ms, and this is what eats that budget).

**Stop the ticker rather than unmount.** The WebGL context, textures and the whole scene graph stay alive, so leaving fullscreen is instant instead of a full theme reload. Nothing in the scene reads wall-clock time — every update is driven by the ticker's own delta — so a paused floor resumes exactly where it left off.

Paused when `fullscreenAgentId` or `fullscreenFilePath` is set, or `document.hidden`. `init()` is async and starts the ticker itself, so it re-applies the paused state on the way out; otherwise a floor that went fullscreen mid-load would come back running.

Typecheck clean, focused tests pass.

### DIFF
--- a/src/renderer/src/scene/office/OfficeFloor.tsx
+++ b/src/renderer/src/scene/office/OfficeFloor.tsx
@@ -168,6 +168,38 @@ export function OfficeFloor() {
   // tears down and rebuilds the whole scene on the new map/cast (see deps below).
   const officeTheme = useStore((s) => s.officeTheme);
 
+  // Is the floor actually on screen? A fullscreen terminal or file editor covers
+  // it completely, and a hidden window shows nothing at all — but the Pixi ticker
+  // went on running the whole scene regardless: every character, thought cloud,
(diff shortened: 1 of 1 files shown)


gaby:
@chaitanyagiri Probably worth merging.

chaitanyagiri:
Hey @gaby , I'll pick it up tomorrow busy in the product hunt launch. Would you mind taking a look at it and post your findings here?

chaitanyagiri:

Your change is live for everyone: the office floor stops rendering while nobody is looking at it.

PR #136, opened 2026-08-13
Sandbox integrations: keep the OS sandbox on, close the tunnel, run agent code off-machine

chaitanyagiri:
Three related pieces of the same story: agents in this app are real CLI processes on the user's machine, and today almost nothing bounds them.

Nothing here changes default behaviour — item 1 is off by default, item 2 falls back to today's provider, item 3 is inert until registered.

## 1 · Keep the engine's OS sandbox on in auto mode (`f3d2841`)

Auto mode drops the sandbox every engine ships (`bypassPermissions`, `codex --dangerously-bypass-approvals-and-sandbox`, `--yolo`). The only isolation left is a git worktree, which bounds nothing outside the repo.

The reason was never that autonomy requires it — it was a path layout. The PROTOCOL makes every worker write to `$AGENT_DIR`, a different path tree from its PTY cwd, which a workspace-scoped sandbox blocks (`src/shared/codexCommands.ts:45` says so). Both CLIs accept extra writable roots, so we declare them in the per-agent config we already author for each engine:

| | flag | writable roots declared in |
| --- | --- | --- |
| claude | `--permission-mode acceptEdits` | `settings.json` — `permissions.additionalDirectories` (Edit/Write tools) **and** `sandbox.filesystem.allowWrite` (bash subprocesses) |
| codex | `-a never -s workspace-write` | `--add-dir <root>` on argv |

Both Claude keys are required or an agent deadlocks mid-protocol: the tool is allowed but `mv … .done/` is denied. Mode is `acceptEdits`, **not** `auto` — `auto` is account/plan gated and rejected at startup where unavailable, which would break the spawn.

god is the one agent granted `HIVE_ROOT` (sole scribe of board.md/tasks.json, drops spawn-requests); that subsumes its own agent dir, so the two collapse into one grant.

**Verified against the installed CLIs, not just asserted.** Claude ran bash with no prompt, wrote the declared agent dir, and got `Operation not permitted` touching `$HOME`. Codex reports `sandbox: workspace-write [workdir, /tmp, $TMPDIR, <agent dir>]`. (Codex's behavioural write test couldn't finish — the account is out of API credits — so its half is verified at the flag/policy level only.)

Toggle: **Settings → Autonomy & Budgets**. Off ⇒ spawn and settings file are byte-identical to before. Engines with no scoped mode keep their bypass flag rather than silently losing autonomy.

## 2 · Give the public tunnel a close handle (`a931eab`)

`slack.ts` and `webhook.ts` each had their own `openTunnel()` and both carried the same admission — *"tunnelmole has no documented close handle; teardown is best-effort"*. Stopping the bridge left the public URL alive until the app quit.

New `src/main/tunnel.ts` returns a handle with a real `close()` and defaults to a cloudflared quick tunnel — an ordinary child process, killed through the same `ensureKilled` ladder the PTYs use. tunnelmole stays as the fallback, unchanged. Also fixes two ways the old code stranded a child: a silent cloudflared is killed on timeout, and an early exit rejects with the tail of its log instead of hanging.

## 3 · Brokered sandbox exec (`16d024f`)

Sandboxed autonomy bounds an agent's own writes; it does nothing about code the model just *wrote* and nobody has read. The loopback broker already forwards to a registered baseUrl with the credential injected in main, so this is a template, not machinery:

- one `INTEGRATION_TEMPLATES` entry (`sandbox-exec`) — the only template whose baseUrl the user must supply, since it points at their own service;
- the capability catalog gains the call plus the boundary that makes it matter, and documents `MD_BROKER_URL`/`MD_BROKER_TOKEN`, which were injected but never written down;
- `examples/sandbox-worker/` is the missing half: a ~90-line Worker over `@cloudflare/sandbox` behind a constant-time bearer check, with `runInSandbox()` as the single swap point for [`@cloudflare/computer`](https://github.com/cloudflare/computer)'s isolate backends. Not compiled by the app build (tsconfigs include only `src/**`).

## Tests

`npm run typecheck` clean; `npm run test:focused` **154 pass / 0 fail** (130 before, +24 across three new files, all registered in the focused suite).

`test/integration-templates.test.cjs` also closes a pre-existing gap: nothing covered the template catalog, so an entry that couldn't pass the upsert gate would have shipped as a dead button.

## Not in scope

Running agents *themselves* in a remote sandbox. Every engine here is a long-lived interactive TUI over a PTY; `@cloudflare/computer` documents no PTY or long-lived-process support and `@cloudflare/sandbox` needs a deployed Worker + Containers. That belongs behind an execution-backend seam, in the Cloud tier — not here.

### DIFF
--- a/examples/sandbox-worker/Dockerfile
+++ b/examples/sandbox-worker/Dockerfile
@@ -1,0 +1,9 @@
+# Container image for the sandbox exec service. The base image ships the agent
+# @cloudflare/sandbox talks to; add whatever toolchain your agents need on top.
+FROM docker.io/cloudflare/sandbox:0.7.0
+
+# Example: give agent code a Python as well as the bundled Node.
+# RUN apt-get update && apt-get install -y --no-install-recommends python3 \
+#     && rm -rf /var/lib/apt/lists/*
+
+EXPOSE 3000
--- a/examples/sandbox-worker/README.md
+++ b/examples/sandbox-worker/README.md
@@ -1,0 +1,75 @@
+# Sandbox exec service (reference)
+
+The other half of the **Sandbox exec (Cloudflare)** integration template. It gives a
+hive worker one capability: *run this code somewhere that isn't the user's laptop.*
+
+Agents in Munder Difflin run as real CLI processes on the user's machine. Sandboxed
+autonomy (Settings → Autonomy & Budgets) bounds what those processes can touch, but the
+riskiest thing an agent does isn't its own file writes — it's executing code the model
+just **wrote** and nobody has read. That belongs off-machine.
+
+## How a worker reaches it
+
+The worker never holds this service's token and never calls it directly. It calls the
+app's loopback broker, which injects the credential and forwards the request:
+
+```
+worker ──▶ $MD_BROKER_URL/i/sandbox-exec/exec                    (capability token)
+              └──▶ broker (Electron main) ──▶ https://<your worker>/exec
+                                                Authorization: Bearer <AUTH_TOKEN>
+```
+
+```bash
+curl -s -X POST "$MD_BROKER_URL/i/sandbox-exec/exec" \
+  -H "Authorization: Bearer $MD_BROKER_TOKEN" \
+  -H 'content-type: application/json' \
+  -d '{"source":"python3 -c \"print(sum(range(10)))\"","backend":"shell"}'
+# → {"ok":true,"stdout":"45\n","stderr":"","exitCode":0}
+```
+
+## Routes
+
+| Route | Body | Returns |
+| --- | --- | --- |
+| `POST /exec` | `{"source": "...", "backend": "shell" \| "js", "timeoutMs"?: number}` | `{"ok":true,"stdout","stderr","exitCode"}` |
+| `GET /health` | — | `{"ok":true,"backend":"sandbox-sdk"}` |
+
+Both require `Authorization: Bearer <AUTH_TOKEN>`. Source is capped at 256 KB and the
+timeout at 120 s, whatever the caller asks for.
+
+## Deploy
+
+```bash
+cd examples/sandbox-worker
+npm install
+wrangler secret put AUTH_TOKEN      # any long random string
+wrangler deploy
+```
+
+Then in Munder Difflin: **Settings → Integrations → Sandbox exec (Cloudflare)**, set the
+base URL to your Worker origin (`https://md-sandbox-exec.<you>.workers.dev`) and paste
+the same token as the secret. Enable it, and it appears in every worker's capability
+catalog.
+
+## Which Cloudflare sandbox?
+
+This implementation uses [`@cloudflare/sandbox`](https://github.com/cloudflare/sandbox-sdk)
+(**beta**): a real container per sandbox id, with streaming exec, background processes,
+git clone, `/workspace` file APIs, and `sleepAfter` lifetimes. It needs Workers +
+Containers, and Docker locally for `wrangler dev`.
+
+[`@cloudflare/computer`](https://github.com/cloudflare/computer) (**preview**, MIT) is
+the other option — a virtual filesystem plus `workspace.runtime.exec(source, {backend})`
+inside Durable Objects, with the authoritative state in SQLite and container /
+isolate-bash / isolate-JS backends. The isolate backends need no container at all, so
+they start faster and cost less for small snippets, at the price of a much smaller
+userland. `runInSandbox()` in `src/index.ts` is the single function to swap; upstream
+calls its APIs unstable, so pin a version if you go that way.
+
+Neither is required by the app: anything that answers these two routes works, including
+a service on your own infrastructure.
+
+## Not production code
+
+One bearer token, no tenancy, no quotas, no audit log. It is deliberately small enough
+to read in one sitting before you trust it with anything.
--- a/examples/sandbox-worker/package.json
+++ b/examples/sandbox-worker/package.json
@@ -1,0 +1,17 @@
+{
+  "name": "md-sandbox-exec",
+  "private": true,
+  "version": "0.1.0",
+  "description": "Reference sandbox exec service for the Munder Difflin 'Sandbox exec (Cloudflare)' integration. Deployed separately; not part of the app build.",
+  "scripts": {
+    "dev": "wrangler dev",
+    "deploy": "wrangler deploy"
+  },
+  "dependencies": {
+    "@cloudflare/sandbox": "^0.7.0"
+  },
+  "devDependencies": {
+    "@cloudflare/workers-types": "^4.20260801.0",
+    "wrangler": "^4.0.0"
+  }
+}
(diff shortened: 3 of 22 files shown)


chaitanyagiri:
## TL;DR

@chaitanyagiri — reviewing your own PR, so I've been deliberately harder on it than on the contributor ones.

**What it does:** three related pieces, none of which change default behaviour. (1) **Keep the engine's OS sandbox on in auto mode** — instead of dropping it wholesale (`bypassPermissions`, `--dangerously-bypass-approvals-and-sandbox`), declare the writable roots the PROTOCOL actually needs (`$AGENT_DIR`, the palace, `HIVE_ROOT` for god) via Claude's `settings.json` and Codex's `--add-dir`, and run `acceptEdits` / `-a never -s workspace-write`. (2) **A tunnel with a real `close()`** — a new `src/main/tunnel.ts` defaulting to a cloudflared quick tunnel (an ordinary child process, killable through the same `ensureKilled` ladder as the PTYs), with tunnelmole as the fallback. (3) **A brokered sandbox-exec template** plus a ~90-line reference Cloudflare Worker.

**Verdict:** the app-side code is the most security-conscious in the repo — injectable deps, an anchored URL regex, prefix-collapsing writable roots, and the honest note that Codex's half is verified at flag level only. **173/173 tests, typecheck and build all green.** Conflicts on `package.json` only, 60 commits behind.

**One real bug, in the example Worker:** `backend: 'js'` is broken for essentially all modern JavaScript, and I reproduced it. Details below.

---

## What I did

Checked out `feat/sandbox-integrations`, ran typecheck, the full suite and a production build, read `tunnel.ts`, `sandboxWritableRoots` and the Worker line by line looking for the ways each could fail open, and **reproduced the one bug I found** in a shell rather than reasoning about it.

## The bug: `backend: 'js'` mangles or executes its own input

`examples/sandbox-worker/src/index.ts`:

```ts
const command = backend === 'js' ? `node -e ${JSON.stringify(source)}` : source;
const result = await sandbox.exec(command, { timeout: timeoutMs });
```

`JSON.stringify` escapes `"` and `\` for a **JS string literal**. It does not escape backticks or `$`, and the result is then handed to `sandbox.exec` as a **shell command string** — where a double-quoted region still honours command substitution and parameter expansion. Reproduced:

```
$ source = 'console.log(`hello`)'
  command = node -e "console.log(`hello`)"
  actual  = sh: hello: command not found        ← the shell ran `hello`, node got console.log()

$ source = 'const x=1; console.log(`v=${x}`)'
  command = node -e "const x=1; console.log(`v=${x}`)"
  actual  = (no output at all)                  ← $x expanded to empty, backticks substituted
```

Template literals and `${}` are in almost every modern JS snippet, so `backend: 'js'` fails on realistic input — silently, with wrong output rather than an error, which is the worse failure. Note this is **not a privilege escalation**: `backend: 'shell'` passes `source` straight to `exec` by design, so an authenticated caller already has a shell. It's a correctness bug that makes the documented JS backend unusable.

The fix is to stop building a shell string: write the source to `/workspace/x.js` via the sandbox's file API and run `node /workspace/x.js`, or use an argv-array form if `sandbox.exec` exposes one. Either way the code never goes through a shell parser.

## Also in the Worker, smaller

- **`tokenOk` isn't constant-time**, despite the name and comment. `if (presented.length !== expected.length) return false` returns early, so it leaks the token's length by timing; the comment defends the *response shape*, which is a different property. Workers expose `crypto.subtle.timingSafeEqual` — using it makes the name true. For a random high-entropy token this is minor, but the file is a reference people will copy.
- **No rate limiting.** A leaked `AUTH_TOKEN` is unmetered container spend on the user's own Cloudflare account. Worth one line in `README.md` even if the code stays as-is.
- **`MAX_TIMEOUT_MS` (120s) exceeds `sleepAfter: '1m'`.** I assume `sleepAfter` is idle-based so an executing job isn't reaped, but it reads like a contradiction — worth a comment.

The framing is right, though: "NOT PRODUCTION CODE… small enough to read in one sitting" is exactly the correct disclaimer for a reference implementation, and auth is checked **before** routing, so `/health` isn't an unauthenticated oracle. That ordering is the thing most people get wrong.

## Sandboxed auto mode — the strongest part

`sandboxWritableRoots` is careful in the ways that matter: absolute-only (a `~/…` entry would be silently ignored by Codex and rejected by Claude, so shipping one would make the policy file read broader than it is), and prefix-collapsing so god's `HIVE_ROOT` grant doesn't imply two separate permissions when it's really one. `--add-dir` goes on argv rather than into a shell string, so paths with spaces are safe. And the reasoning for `acceptEdits` over `auto` — that `auto` is account/plan-gated and would fail the spawn where unavailable — is the kind of detail that only comes from actually trying it.

Two questions rather than objections:

- **No symlink resolution.** If `$AGENT_DIR` is reached through a symlink, the declared root and the path the kernel enforces on can differ, and the grant silently doesn't cover what it names. A `realpathSync` before collapsing would close that.
- **Case-insensitive filesystems.** `covers()` compares with `startsWith`, so on default APFS `/Users/x/Hive` and `/Users/x/hive` are the same directory but won't collapse. Harmless (a redundant grant, not a wider one), but it means the file can state the grant twice.

## `tunnel.ts`

The clearest win of the three, and the comment explaining *why* — that both call sites carried the same "no documented close handle; teardown is best-effort" admission while the public URL outlived the server — is the right way to record a fix. `QUICK_TUNNEL_RE` being anchored to `.trycloudflare.com` so a hostile-looking log line can't be mistaken for the tunnel URL is a nice touch, as is killing a silent cloudflared on timeout instead of stranding it.

One thing worth naming in the release notes rather than fixing: `'auto'` silently switches an existing user's tunnel provider the moment `cloudflared` appears on their PATH — different domain, different URL shape, and Slack app config may be pinned to a URL. Not wrong, just invisible.

## What I tested

| Check | Result |
|---|---|
| `node --test test/*.test.cjs` | ✅ **173/173**, 0 fail |
| New tests | ✅ `sandbox-mode` (181), `tunnel` (156), `integration-templates` (78) lines |
| `backend: 'js'` quoting | ❌ **reproduced broken** (above) |
| `tokenOk` constant-time claim | ⚠️ length-leaking early return |
| Malware / obfuscation / exfiltration | ✅ none. The Worker is the only network-egress code added and it's an example, excluded from the app build (`tsconfig` covers `src/**` only) |
| Mergeability | ❌ **CONFLICTING** — `package.json` only; 60 commits behind |

## Questions

Since this is yours, these are decisions rather than requests:

1. **Fix or descope `backend: 'js'`?** Fixing it is a few lines. Alternatively drop the `js` backend from the example and let `shell` be the only documented mode — `node -e` via shell was always going to be fragile.
2. **Is `sandboxedAutoMode` meant to become the default eventually?** Right now it's an opt-in toggle in Settings → Autonomy & Budgets, which means the shipped default is still full-bypass autonomy. If the answer is "yes, once Codex's behavioural test can run", that belongs in the PR description — it changes how a reviewer weighs the whole thing.
3. **Codex's half is verified at flag/policy level only** (you say so, and the honesty is appreciated). Is that blocking for merge, or do you want to land it behind the off-by-default toggle and verify when credits allow? Given nothing changes unless a user turns it on, I'd land it.
4. This has been open since 13 August and is 60 commits behind, but conflicts on `package.json` alone. Worth a rebase soon — the sandbox work touches `hive.ts`, and **#128** (still open) touches `hive.ts` too, so the gap only gets more expensive.

PR #137, opened 2026-08-14
fix(terminal): strip the CLI's quote rail from copied selections

chaitanyagiri:
## The bug

Copying a line of agent output from a terminal pastes a stray block glyph with it:

```
▎ Munder Difflin — clones for you and your team, working 24/7.
```

Claude Code paints a markdown blockquote **into the character grid** — the `▎` rail is a real cell in column 0, not a border drawn beside the text. A triple-click or a full-width drag therefore selects it along with the prose, and xterm's `getSelection()` hands it straight to the clipboard.

## The fix

Sanitize on the way out, in `copySelection()` — the single place every copy path already funnels through (Ctrl/Cmd+C with a selection, Ctrl/Cmd+Shift+C, right-click). Per line: drop a leading run of gutter rails, keep the indent.

New `src/renderer/src/components/terminalSelection.ts` (pure string module, electron-free, unit-testable).

Kept deliberately narrow so it can't corrupt legitimate copies:

- **A rail only counts when a space follows it, or it ends the line** — the shape a quote gutter always has. Bar charts and block art (`▌▌▌▌ 40%`) keep their bars, because those run together.
- **Only the left/right BLOCK elements are in the set** (`▌▍▎▏▐▕`). Box-drawing verticals (`│`, `┃`) are *not*: `tree`, `git log --graph` and every framed table indent continuation lines with `│   `, and stripping that would break far more copies than it fixes.
- **No rail anywhere → returned byte-for-byte unchanged** (fast path, no split).
- A rail-only selection sanitizes to empty but still **counts as a copy**, so right-click doesn't fall through to paste and inject clipboard text into the agent's prompt.

## Verification

- 10 new tests in `test/terminal-selection.test.cjs`, added to `test:focused`.
- **Mutation-checked both directions:** 4 fail against a no-op sanitizer (the fix does something); 3 of the guard tests fail against a naive strip that also eats box drawing (the guards have teeth).
- `npm run test:focused` → **140/140**; `npm run typecheck` clean.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "typecheck:node": "tsc --noEmit -p tsconfig.node.json",
     "typecheck:web": "tsc --noEmit -p tsconfig.web.json",
     "typecheck": "npm run typecheck:node && npm run typecheck:web",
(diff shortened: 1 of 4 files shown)


chaitanyagiri:
Shipped in **[v0.4.4](https://github.com/chaitanyagiri/munder-difflin/releases/tag/v0.4.4)** — thank you, @chaitanyagiri. 🙏

Your change is live for everyone: the CLI's quote rail is stripped from copied selections, so terminal copy comes back clean.

PR #142, opened 2026-08-15
Fix renderer task-ledger lost updates

baziyer:
## Problem

Renderer task actions can replace `tasks.json` with a snapshot read seconds earlier. If a webhook or Slack card is appended between that read and write, answering an Ask Me question, moving a card, or dismissing a card silently deletes the new work. This was reproduced live: a webhook POST returned a task ID, then an Ask Me write changed the ledger from 9 cards back to 8 and the webhook status capability returned 404.

## Fix

- Add main-process atomic `addTask`, `patchTask`, and `deleteTask` operations that always mutate the latest on-disk ledger.
- Route Ask Me answers/dismissals, task status moves/deletes, and Slack promotion through those operations.
- Remove the renderer bridge and IPC handler for whole-ledger replacement.
- Preserve unrelated fields such as `webhook.tokenHash` and Slack metadata.
- Add a regression that reproduces the stale-snapshot ordering and locks the renderer boundary.

## Verification

- RED: regression failed because atomic APIs did not exist and the unsafe bridge remained exposed.
- GREEN: `node --test test/hive-task-mutation.test.cjs` — 4/4 passed.
- `npm run typecheck` — passed (node + web).
- `npm run build` — passed.
- Complete focused suite: 132 passed; `proc-kill` passes with macOS process access; the existing `hive-hook-node` cancellation reproduces unchanged on upstream main.

Exact head: `b622f0ac56ad4d9f2d6b7bfa3b7b9ce768e4849d`

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "typecheck:node": "tsc --noEmit -p tsconfig.node.json",
     "typecheck:web": "tsc --noEmit -p tsconfig.web.json",
     "typecheck": "npm run typecheck:node && npm run typecheck:web",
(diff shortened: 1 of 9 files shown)


chaitanyagiri:
Shipped in **[v0.4.4](https://github.com/chaitanyagiri/munder-difflin/releases/tag/v0.4.4)** — thank you, @baziyer. 🙏

Your change is live for everyone: renderer task-ledger updates are atomic, so task cards stop going missing when two things edit the board at once.

PR #144, opened 2026-08-15
fix(hive): keep the cost ledger out of the hive's git history

gts-47:
`cost-ledger.jsonl` is append-only and gains a row per usage sample, so a repo that tracks it stores a fresh copy of the **whole file** on every hive commit — and the hive commits constantly.

On my machine that is a 251 MB ledger with ~3,700 copies behind it, which leaves git walking hundreds of gigabytes of blob. A routine `git gc --auto` turned into a `pack-objects` run that climbed to **16 GB resident** and pushed the machine 29 GB into swap. Nothing else in the hive grows that way; every other churny live file (`fleet.json`, `hooks.sock`) is already ignored.

**What this does**

- adds `cost-ledger.jsonl` to the hive root `.gitignore` written by `ensureHive`
- drops the copy already in the index, once per process, from `commit()`

The second half is the part that matters for an existing hive: git keeps recording a file it already tracks no matter what `.gitignore` says, so the ignore line alone would have read as applied while the repo went on growing — the same trap `untrackCodexHomes` exists to close (#128).

The ledger stays on disk, so the cost history the app reads is untouched; only its version history stops.

Note: this stops the growth. A hive that already carries the old copies still needs a history rewrite to reclaim the space — that is a local operation, not something the app should do on its own.

Tests: 150/150 pass.

https://claude.ai/code/session_01DUpadJJuEQz1zpQ6Vk83j3

### DIFF
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -477,7 +477,7 @@ export class HiveManager {
 
     // Keep the churny/ephemeral live files out of the hive git repo.
     const gitignore = join(root, '.gitignore');
-    const want = ['fleet.json', 'hooks.sock', '.DS_Store'];
+    const want = ['fleet.json', 'hooks.sock', 'cost-ledger.jsonl', '.DS_Store'];
     let lines: string[] = [];
     if (existsSync(gitignore)) { try { lines = readFileSync(gitignore, 'utf8').split('\n'); } catch { lines = []; } }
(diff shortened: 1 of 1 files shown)


chaitanyagiri:

Your change is live for everyone: the cost ledger stays out of the hive's git history.

PR #101, opened 2026-06-29
feat(provider): add GitHub Copilot CLI adapter preset

anxkhn:
## What & why

Adds GitHub Copilot CLI (`copilot`, npm `@github/copilot`) as a first-class engine in the agent provider registry, alongside the existing Claude/Codex/OpenCode/Antigravity/Qwen/Crush/Pi presets. Copilot now ships a clean non-interactive print mode, so a hive worker can spawn it the same way it spawns Codex:

```
copilot -p "<prompt>" -s --allow-all-tools --no-ask-user [--model <model>]
```

It mirrors the existing Codex/OpenCode preset shape, so it slots into the same spawn, model-picker, auto-mode, and resume plumbing with no special-casing:

- New `copilot` preset: `defaultCommand: copilot`, `autoFlag: -s --allow-all-tools --no-ask-user` (gated by the floor auto-mode toggle), `-p` initial-prompt injection (non-hiveAware, like codex), `--model` passthrough, `--resume` support, and `npm install -g @github/copilot` for the missing-CLI installer.
- `COPILOT_MODELS` picker (Claude Sonnet 4.5 default, gpt-5.4, auto).
- `inferAgentProvider` / `isAgentProvider` recognize the `copilot` binary; the voice hire tool and `PROVIDER_COMMAND` map include it.
- It's non-hiveAware with no hook bridge, so `canReceiveInbox` is `false` (print mode exits per turn; mail bounces to the god). That keeps the addition honest and surgical rather than claiming a drain path that doesn't exist yet.

Verified live against the real `copilot` v1.0.65: print mode (`-p -s --allow-all-tools --no-ask-user --model claude-sonnet-4.5`) and the `--output-format json` `result` line (sessionId + exitCode) both behave as wired.

A small self-contained registry test (`test/agent-provider.test.cjs`, run with `node`) transpiles the shared TS registry and asserts selection, the command shape, and model/resume passthrough.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

## Screenshots / clips

No visual redesign: the Add Agent provider picker auto-iterates `AGENT_PROVIDER_PRESETS`, so a new "Copilot" chip appears via the existing styling with no token changes.

## Checklist

- [x] `npm run typecheck` passes (the de-facto CI gate).
- [x] `npm run build` succeeds.
- [x] Any new UI derives from `DESIGN.md` / `tokens.ts` (no new UI; existing picker styling).
[...]

PR #145, opened 2026-08-15
feat(provider): add Cursor Agent CLI (`agent`) engine preset

GBurchell27:
## Summary
- Adds a first-class **`cursor`** provider preset that spawns Cursor Agent CLI (`agent`).
- God-eligible (interactive TUI, no `-p`) so Michael can run on Cursor credits; hive protocol is typed in after boot (`seedDelivery: 'type-into-tui'`).
- Default orchestrator model: **`gpt-5.6-luna-high`** (cheap / high-context Cursor catalog model). Auto-mode flags: `--force --trust`.
- Native missing-CLI installer via `curl https://cursor.com/install -fsS | bash` (and Windows `irm … | iex`).
- Model picker includes Luna / Sol / Composer / Opus / Sonnet quick-picks; command field stays editable for any live `agent models` slug.

## Why
Cursor CLI is widely used and already headless-capable (`agent -p` / interactive). Stock Munder cannot select it today. This mirrors the Copilot community-provider pattern (PR #101) while keeping the session alive for inbox delivery (unlike Copilot print mode).

## Notes for reviewers
- No Cursor hook/proxy bridge yet — mail uses the terminal work-order / idle-nudge fallback (same honesty bar as early Crush).
- `gpt-5.6-luna-*` is a Cursor-catalog model (credits), not a separate public OpenAI API slug.
- Slash compact/clear for Cursor left as `NO_CONTEXT_COMMANDS` until a frozen binary command table is transcribed.

## Test plan
- [x] `node test/agent-provider.test.cjs`
- [x] `node --test test/provider-config.test.cjs test/provider-automation.test.cjs test/cli-install-ladder.test.cjs`
- [x] `npm run typecheck`
- [ ] Manual: `npm run dev` → onboarding Step 2 shows **Cursor** → pick `gpt-5.6-luna-high` → Michael spawns `agent --force --trust --model gpt-5.6-luna-high`
- [ ] Manual: Add Agent worker on Cursor; confirm terminal stays interactive and receives a typed work order
- [ ] Manual: missing `agent` on PATH triggers native install banner / curl installer

### DIFF
--- a/docs/hires/scripts/build-data.py
+++ b/docs/hires/scripts/build-data.py
@@ -14,7 +14,7 @@ may use any model string; suggestions never validate).
 """
 import json, os, re, sys, urllib.request
 
-PROVIDERS = ['claude', 'antigravity', 'codex']
+PROVIDERS = ['claude', 'antigravity', 'codex', 'cursor']
 UPSTREAM_CONFIG = ('https://raw.githubusercontent.com/chaitanyagiri/munder-difflin/'
                    'main/src/renderer/src/store/config.ts')
 
--- a/docs/hires/validator.js
+++ b/docs/hires/validator.js
@@ -5,8 +5,8 @@
 
 window.HireSpec = (function () {
   const SPEC = 'munder-difflin/hire@1';
-  const PROVIDERS = ['claude', 'antigravity', 'codex'];
-  const PROVIDER_LABEL = { claude: 'Claude Code', antigravity: 'Antigravity', codex: 'Codex' };
+  const PROVIDERS = ['claude', 'antigravity', 'codex', 'cursor'];
+  const PROVIDER_LABEL = { claude: 'Claude Code', antigravity: 'Antigravity', codex: 'Codex', cursor: 'Cursor' };
   const FLAG_RE = /^[A-Za-z0-9._\/=:,@+-]{1,100}$/;
   // model flows onto the spawn command line — reject shell metacharacters
   // (mirror of MODEL_RE in the app's src/shared/hire.ts).
@@ -43,7 +43,7 @@ window.HireSpec = (function () {
     cap(raw.homepage, 300, 'homepage');
     if (raw.provider !== undefined) {
       const p = normalizeProvider(raw.provider);
-      if (!PROVIDERS.includes(p)) errors.push('"provider" must be claude, antigravity (or agy), or codex');
+      if (!PROVIDERS.includes(p)) errors.push('"provider" must be claude, antigravity (or agy), codex, or cursor');
     }
     if (raw.commandFlags !== undefined) {
       if (!Array.isArray(raw.commandFlags) || raw.commandFlags.length > 16) {
--- a/electron.vite.config.ts
+++ b/electron.vite.config.ts
@@ -52,21 +52,11 @@ function copyMainSidecars() {
 export default defineConfig({
   main: {
     plugins: [externalizeDepsPlugin(), copyMainSidecars()],
-    define: defineMain,
-    build: {
-      rollupOptions: {
-        input: { index: resolve(__dirname, 'src/main/index.ts') }
-      }
-    }
+    define: defineMain
   },
   preload: {
     plugins: [externalizeDepsPlugin()],
-    define,
-    build: {
-      rollupOptions: {
-        input: { index: resolve(__dirname, 'src/preload/index.ts') }
-      }
-    }
+    define
   },
   renderer: {
     define,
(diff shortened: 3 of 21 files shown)


chaitanyagiri:
### Review — `feat(provider): add Cursor Agent CLI (agent) engine preset`

Thanks for this — the preset itself is well-shaped and follows the house pattern closely (good comments on *why* `-p` is avoided, `seedDelivery: 'type-into-tui'` matching the Crush precedent, `autoModeFlag` gated by `config.autoMode`, multi-flag string is precedented by `pi`'s `-s --allow-all-tools --no-ask-user`, and `nativeInstallCommand` mirrors the `claude` preset). The provider plumbing (`hire.ts`, `realtimeActions.ts`, `AddAgentModal`, `docs/hires/validator.js`, `CURSOR_MODELS`, the four test files) is complete and consistent.

Two things need to come out before this can land, plus a couple of questions.

#### 1. Blocking — undisclosed toolchain overhaul rides along with the feature

`package.json` + `package-lock.json` (+2553 / −3957) change dependencies that have nothing to do with Cursor:

| package | main | this PR |
|---|---|---|
| electron | ^32.2.0 | **^43.4.0** |
| vite | ^5.4.8 | **^8.2.1** |
| electron-vite | ^2.3.0 | **^5.0.0** |
| electron-builder | ^25.1.8 | **^26.15.3** |
| better-sqlite3 | ^11.10.0 | **^13.0.3** |
| @electron/rebuild | ^3.7.0 | **^4.2.0** |
| localtunnel | ^2.0.2 | **^1.8.3** (downgrade) |

None of this is in the PR description. An 11-major Electron jump plus `better-sqlite3` 11→13 changes the native ABI for `better-sqlite3` and `node-pty` (which this app rebuilds in `postinstall`), and it lands in a release-tagged repo. Please revert `package.json` / `package-lock.json` to the base and re-generate the lock with only what the feature needs (which looks like: nothing). If the upgrade is genuinely wanted, it deserves its own PR with its own smoke test.

#### 2. Blocking (coupled to #1) — `electron.vite.config.ts` drops the explicit entry points

```diff
-    define: defineMain,
-    build: { rollupOptions: { input: { index: resolve(__dirname, 'src/main/index.ts') } } }
+    define: defineMain
```
…same for `preload`. This reads as an accommodation for electron-vite v5 defaults rather than an intentional change. With the dependency bump reverted this must revert too. (Good that `copyMainSidecars()` was left in place — dropping it is a recurring source of `Cannot find module './slack-trigger.cjs'` at runtime.)

#### 3. Blocking — unrelated feature bundled in

`src/renderer/src/components/ProjectNoticeboard.tsx` (+112, new), the `projects` tab added to `SidebarTabs.tsx`, and the `App.tsx` right-pane restructure (+37 / −15) are a separate feature. Please split them into their own PR so this one is reviewable as "add a provider preset".

#### 4. Question — is the binary really `agent`?

```ts
defaultCommand: 'agent',
…
if (bin === 'agent' || bin === 'cursor-agent') return 'cursor';
```
The official installer (`curl https://cursor.com/install`) ships **`cursor-agent`** on PATH; `agent` is the shorter alias. Two concerns: (a) if `cursor-agent` is the canonical name, the default command should be `cursor-agent` with `agent` as the alias, not the reverse; (b) `agent` is a very generic name to claim in `providerFromBinary` — any unrelated `agent` binary on a user's PATH would be labelled as Cursor. Worth flipping the precedence.

#### 5. Non-blocking notes

- **Model slugs** (`gpt-5.6-luna-high`, `composer-2.5`, …) can't be validated in CI and will drift with Cursor's catalog. The editable command field mitigates this — no change requested, just noting the maintenance cost.
- **`recommendedOrchestratorModel: 'gpt-5.6-luna-high'`** is a product call: onboarding Step 2 tells users to give Michael "a longer-context, higher-capability model", so defaulting the orchestrator to the *cheap* tier is slightly at odds with the surrounding copy. @chaitanyagiri's call.
- **Security**: no secrets or credentials introduced; auto-mode flags are gated behind `config.autoMode` like every other engine; the `curl … | bash` installer is a trust-on-first-use vector but is a hardcoded official URL and matches the existing `claude` preset, so it's consistent with the current bar.
- **CI**: no checks have reported on this PR (`statusCheckRollup` is empty). I could not reproduce a build/typecheck locally because the branch swaps the entire toolchain — worth re-running the listed test plan after the dependency revert.

**Verdict: request changes** — drop items 1–3, answer 4, and the Cursor preset itself looks good to merge.

<sub>Automated review by Jim (Munder Difflin hive) — recurring PR review pass.</sub>

chaitanyagiri:
Hey, I was gonna review, this myself too. Why did you delete this?

PR #149, opened 2026-08-16
Fix seven hive/agent harness defects: task-ledger data loss, worker spawn, queue delivery, teardown, and token accounting

bluepanda999:
Seven fixes to the hive/agent harness, found by running a multi-agent hive on v0.4.3 for a day and diagnosing each failure against the live app rather than from reading alone. Each commit is standalone and independently revertable.

## What is fixed

| Commit | Defect | Impact |
|---|---|---|
| `75d1c0e` | `hiveWriteTasks` whitelist overwrite drops every non-whitelisted field from **every** card on any single task edit | Data loss |
| `c7ba7fe` | Main-spawned (non-GUI) workers never receive their objective, and do not appear in the fleet view | Feature is dead on arrival |
| `e84dc4a` | Main-only spawns miss the permission posture a GUI hire gets, so the agent waits forever on an approval nobody can see | Agent unreachable |
| `cfa59f7` | Queued messages never drain to an agent whose terminal has fallen silent | Messages lost |
| `242f283` | `PtyManager.kill()` deletes its session synchronously while node-pty's `onExit` arrives later, so the stale-id guard trips and `teardownPty` is skipped on every explicit kill | Dead agents stay on the roster as active |
| `18d83c7` | Mempalace installed after app boot never arms the mine loop, so semantic memory sits on "getting ready" forever | Feature stuck |
| `06879a3` | The transcript usage fallback sums every `.jsonl` transcript ever written in an agent's cwd, across all sessions and all agents | Token accounting is wrong by orders of magnitude |

## The last one is worth expanding

`telemetry.ts`'s `transcriptFallback` is the path `getAgentUsage()` takes whenever an agent has no live OTel data yet, which is the common case immediately after spawn. It called `readAgentUsage(cwd)` with no session filter, so it summed every transcript ever left in that cwd's project directory: other agents' history, and previous runs'. Shared cwds are routine, since several spawn paths skip git isolation and even isolated workers can share a base repo cwd.

The fix resolves the agent's current session id and passes it to the `sessionId` filter that `readAgentUsage` already supports. Where no session id is known yet, it now reports no data instead of someone else's history. The code's own comment already described the unfiltered behaviour as "the legacy behavior".

Measured on a real shared project directory: the unfiltered sum across four sessions is 65.7M tokens, mostly `cacheRead`, while the per-session figures are 6.9M, 47.5M and 18M.

## Verification

- `npm run typecheck` clean on both the node and web projects.
- `npm run test:focused` passes 159/159, exit 0.
- Each fix that is unit-testable has a dedicated `test/<defect>.test.cjs`, registered in `test:focused`, following the convention already in the repo.
- The D11 regression test was checked against the pre-fix code and fails there with the exact bug shape, so it is a real regression test rather than a tautology.

Note for anyone re-running the gates: this suite uses `node --test` with the spec reporter, so grepping for TAP-style `not ok` lines matches nothing and reads green regardless of outcome. The trailing `fail` tally plus the exit code are the signals that mean anything.

## An observation we could not fully explain

The worker spawn path (`c7ba7fe`, `e84dc4a`) cannot be exercised by unit tests, so it was verified by spawn probes against a locally built app. During that testing, two spawn requests identical in every field except `tokenCap` produced very different outcomes: with a finite cap of 200,000 the worker registered, made zero LLM calls, and was reaped citing 143,369,766 tokens used; with `tokenCap: 0` the same worker ran to completion and the fleet snapshot reported its real usage as 251,901 tokens.

That is reported as an observation, not a diagnosis. It is consistent with the reaper reading the unfiltered cross-session total that `06879a3` fixes, and the two figures differ by roughly the ratio you would expect, but we did not capture a log proving which branch of `getAgentUsage()` the reaper actually took. Worth knowing if you see workers dying before doing any work. Note also that 251,901 legitimately exceeds a 200,000 cap, so a worker's boot baseline can genuinely exceed a small cap; the surprising part is the 143M figure, not the decision to reap.

## Deliberately not included

- `breaker.ts`'s `tokensOf` still sums `input + output + cacheRead + cacheCreation` cumulatively. Narrowing it to an output-only or windowed measure was considered and rejected for now, because `CommandCenterPanel.tsx` independently recomputes the same four-field sum from the same sample to drive each agent's token meter against the same cap. Changing only the breaker's comparison would desync the meter from what actually trips. The traced root cause was the cross-session summation, so the fix stays scoped to that.
- Two other unfiltered `readAgentUsage` callers are left alone because neither is live: `StubUsageProvider` in `usage.ts` is never instantiated now that `telemetry.ts` provides usage, and the `hive:agentUsage` IPC handler is wired but called by no renderer component. Flagging them in case either is wired up later.
- No latch-once-per-crossing was added to the breaker. Scoping the usage read removes the driver of repeated re-escalation, so a separate latch looked like belt-and-braces rather than a fix. Happy to add one if you would prefer the defence in depth.
- A token cap set below an agent's boot baseline still fails silently rather than loudly. That is a UX concern and out of scope here.

Happy to split this into separate PRs, reorder it, or drop any commit you would rather not take.

AI-assisted; human-reviewed


chaitanyagiri:
## TL;DR

Thanks @bluepanda999 — this is the strongest engineering in the open queue. Seven defects, each its own revertable commit, each diagnosed against a running app rather than by reading, with a dedicated test wherever one is possible and an explicit list of what you deliberately *didn't* fix. That last section is rarer than it should be.

**Verdict:** no malicious code, no NSFW, no prompt injection, no dependency changes. **178/178 tests pass**, typecheck is clean on both projects, and `npm run build` succeeds. I independently confirmed the three defects I could verify statically are still live on `main`, and the token-accounting one (`06879a3`) is, in my reading, **the highest-impact single fix across all twenty open PRs**.

**Two things need a maintainer decision, not a code change:** one of your seven commits is a **duplicate of PR #159** by a different author (yours is the more complete version), and the PR conflicts on three files. Plus one design question about the merge semantics.

---

## What I did

Checked out `fix/askme-data-loss-and-queue-drain`, ran typecheck, the full suite and a production build, then verified each statically-checkable claim against current `main` — including reading `readAgentUsage`'s signature to confirm the `sessionId` filter really does already exist.

Also: thank you for the note about `node --test`'s spec reporter and TAP grepping. It's correct, and it's the kind of thing that silently invalidates a whole verification pass. I read the `# fail` tally and the exit code throughout.

## Verified against `main`

| Commit | Claim | My check |
|---|---|---|
| `75d1c0e` | Ledger writes drop non-whitelisted fields from every card | ✅ `hive.ts:1371` `writeTasks()` does `writeJson(tasks.json, { tasks })` — wholesale replace, no merge. No `taskLedger.ts` exists on `main` |
| `06879a3` | The transcript fallback sums every session in a shared cwd | ✅ `telemetry.ts:392` calls `readAgentUsage(cwd)` with **no options**, while `transcript.ts:270` is `readAgentUsage(cwd, opts: ReadUsageOptions = {})` and threads `opts.sessionId` into `readFileUsage`. The filter was already there and simply wasn't passed. `main`'s collector is wired with `resolveCwd` only |
| `242f283` | Explicit `kill()` skips `teardownPty` via the stale-id guard | ✅ confirmed — and see the duplication note below |

## Why `06879a3` is the one I'd merge first

```
   getAgentUsage(agentId)
        │  no live OTel yet — the common case right after spawn
        ▼
   transcriptFallback(agentId)
        │
        ▼
   readAgentUsage(cwd)          ← no sessionId
        │
        ▼
   ~/.claude/projects/<key>/*.jsonl
        │
        ├── this agent, this session          ✅ wanted
        ├── this agent, three previous runs   ✗
        └── every OTHER agent that shared     ✗
            the cwd, ever
```

Your measured numbers make it concrete: 65.7M unfiltered against per-session figures of 6.9M / 47.5M / 18M. And the consequence isn't cosmetic — `breaker.ts` reads `agentTokenCaps` and reaps against this figure, so a worker that has made **zero** LLM calls can be killed for exceeding a cap using a stranger's history.

On your "observation we could not fully explain": I think your instinct is right and it's more than a coincidence of ratios. `tokenCap: 0` disables the cap check entirely, so that worker never consults `getAgentUsage()` and survives; a finite cap forces the lookup, hits the unfiltered fallback, reads the shared directory's entire history (143M), and reaps immediately. The two outcomes differ *because* one path reads the number and the other doesn't. I can't prove which branch the reaper took either — but the mechanism is sufficient, and it means the bug's real symptom is **"finite token caps are broken for any worker in a shared cwd"**, which is worth putting in the CHANGELOG in those words. Your restraint in reporting it as an observation was correct; I'd just say the diagnosis now looks solid.

Your point that 251,901 legitimately exceeds a 200,000 cap is also worth keeping — a worker's boot baseline can genuinely blow a small cap, and that's a separate UX problem from this bug.

## The ledger fix

`mergeTaskLedger` / `patchTaskInLedger` are well-designed, and the doc comments explain the *second* trap — that re-serializing the display model produces real coerced values (`priority: "high"` → `3`) which then beat any merge — which is the part most people would miss. Splitting persistence-side backstop from caller-side rule is right.

One question about the semantics: **the merge is shallow.** For a card with a nested object (`slack: { channel, thread_ts, ts }`), a writer that sends a partial `slack` object replaces the whole thing and the unsent keys are gone — exactly the class of loss this function exists to prevent, one level down. Your comment argues shallow is correct *for missing top-level keys*, and I agree; I'm asking whether nested objects were considered and judged safe (because no writer ever sends a partial nested object), or not yet encountered.

Minor: `byId` keeps the **first** entry for a duplicate id. Given the ledger is hand-written by the god, duplicate ids are plausible. First-wins is a defensible choice, but it's silent — worth a line in the comment.

## Duplication with PR #159 — needs a decision

Your commit `242f283` and @gpechieu's #159 (fix 3) are **the same fix**: the `wasWorker` flag read before cleanup, the `hive:agentArchived` broadcast in `teardownPty`, and the explicit `teardownPty(workerId)` after each `ptyManager.kill(workerId)` in `ephemeralWorkerTick`. Two contributors independently found the same root cause and wrote near-identical code.

**Yours is the more complete one.** #159 fixes the three sites inside `ephemeralWorkerTick`; you *also* fix `ipcMain.handle('workers:stop')` — the manual "stop worker" button, whose own doc-comment still asserts the belief both of you disproved ("kill → teardownPty runs"). I flagged that gap on #159. Whichever lands first, the other needs a rebase; if #159 lands first, its `workers:stop` gap must still be closed.

## What I tested

| Check | Result |
|---|---|
| `node --test test/*.test.cjs` | ✅ **178/178**, 0 fail |
| New tests registered in `test:focused` | ✅ yes, following repo convention |
| Mergeability | ❌ **CONFLICTING** — `package.json`, `AskMeTab.tsx`, `TaskDetailOverlay.tsx`. Branch is 44 commits behind |

I did not re-verify the D11 regression against pre-fix code as you did, nor the two spawn-path fixes (`c7ba7fe`, `e84dc4a`) — you're right that they aren't unit-testable, and I have no second machine to run spawn probes on. I'm taking those on the strength of the rest, which held up under every check I *could* run.

## Questions

@bluepanda999 — (1) nested-object shallow merge: considered and safe, or not yet hit? (2) Would you rebase and drop `242f283` if #159 lands first — or would you rather #159 drop its version, since yours covers the fourth kill site? Genuinely your call and I'll back whichever you and @gpechieu agree. (3) On `breaker.ts`'s `tokensOf` (your "deliberately not included"): with `06879a3` in place, is the cumulative `input + output + cacheRead + cacheCreation` sum now *correct per session*, so the remaining question is only whether cumulative is the right measure? That reframes it from a bug to a product decision.

@chaitanyagiri — three things. **(a)** Please pick a resolution for the #149/#159 overlap so neither author redoes the other's work. **(b)** `06879a3` deserves to be prioritised on its own merits — finite per-agent token caps are, as far as I can tell, currently unusable for any worker in a shared cwd, and the failure mode is a worker dying instantly having done nothing. **(c)** Seven fixes in one PR is a lot to hold at once; the commits are clean and standalone, so if you'd rather merge `06879a3` and `75d1c0e` today and take the rest after a rebase, the branch supports that.

PR #153, opened 2026-08-17
feat(provider): Cursor CLI preset + standing goal editor

GBurchell27:
## What & why

Sorry for the churn on #145 — I closed that PR after your review because I realised it bundled unrelated changes (dependency upgrades, a personal sidebar feature) and I wanted to resubmit something cleaner. That was my mistake; I should have fixed it in place instead of closing it.

This PR replaces #145 with a focused set of changes:

1. **Cursor Agent CLI provider preset** — register `cursor` as a first-class, god-eligible provider so Michael and workers can run on Cursor credits. Interactive spawn (no `-p`) keeps the TUI alive for hive mail via `seedDelivery: 'type-into-tui'`; auto-mode uses `--force --trust`.
2. **Standing goal editor** — agents' standing goals are injected on session start (via hooks where supported, or prepended in `useHive` for engines without hook injection). Adds an Edit Agent modal from the agent detail panel.

**Review feedback addressed from #145:**
- No `package.json` / `package-lock.json` / `electron.vite.config.ts` changes
- No Project Noticeboard / sidebar tab / App layout restructure
- Default binary is now **`cursor-agent`** (canonical installer name); `agent` remains an alias in `inferAgentProvider`

Closes nothing — supersedes the intent of #145.

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

## Screenshots / clips

Standing goal editor (visual change) — screenshot to follow in a comment after manual smoke test. Cursor provider appears in onboarding Step 2 model picker like other providers.

## Checklist

- [ ] `npm run build` succeeds. *(not run in this environment — will verify locally)*
- [x] If I added art, it's my own or compatibly licensed, and listed in `ATTRIBUTION.md`.
- [x] PR is focused on a single change. *(two related agent-management features; happy to split standing goals if you prefer)*

## Test plan

- [x] `node test/agent-provider.test.cjs`
- [x] `node --test test/provider-config.test.cjs test/provider-automation.test.cjs test/cli-install-ladder.test.cjs`
- [x] `npm run typecheck`
- [ ] Manual: `npm run dev` → onboarding Step 2 shows **Cursor** → pick `gpt-5.6-luna-high` → Michael spawns `cursor-agent --force --trust --model gpt-5.6-luna-high`
- [ ] Manual: Edit an agent's standing goal from the detail panel; confirm it appears on next session start
- [ ] Manual: missing `cursor-agent` on PATH triggers native install banner

### DIFF
--- a/docs/hires/scripts/build-data.py
+++ b/docs/hires/scripts/build-data.py
@@ -14,7 +14,7 @@ may use any model string; suggestions never validate).
 """
 import json, os, re, sys, urllib.request
 
-PROVIDERS = ['claude', 'antigravity', 'codex']
+PROVIDERS = ['claude', 'antigravity', 'codex', 'cursor']
 UPSTREAM_CONFIG = ('https://raw.githubusercontent.com/chaitanyagiri/munder-difflin/'
                    'main/src/renderer/src/store/config.ts')
 
--- a/docs/hires/validator.js
+++ b/docs/hires/validator.js
@@ -5,8 +5,8 @@
 
 window.HireSpec = (function () {
   const SPEC = 'munder-difflin/hire@1';
-  const PROVIDERS = ['claude', 'antigravity', 'codex'];
-  const PROVIDER_LABEL = { claude: 'Claude Code', antigravity: 'Antigravity', codex: 'Codex' };
+  const PROVIDERS = ['claude', 'antigravity', 'codex', 'cursor'];
+  const PROVIDER_LABEL = { claude: 'Claude Code', antigravity: 'Antigravity', codex: 'Codex', cursor: 'Cursor' };
   const FLAG_RE = /^[A-Za-z0-9._\/=:,@+-]{1,100}$/;
   // model flows onto the spawn command line — reject shell metacharacters
   // (mirror of MODEL_RE in the app's src/shared/hire.ts).
@@ -43,7 +43,7 @@ window.HireSpec = (function () {
     cap(raw.homepage, 300, 'homepage');
     if (raw.provider !== undefined) {
       const p = normalizeProvider(raw.provider);
-      if (!PROVIDERS.includes(p)) errors.push('"provider" must be claude, antigravity (or agy), or codex');
+      if (!PROVIDERS.includes(p)) errors.push('"provider" must be claude, antigravity (or agy), codex, or cursor');
     }
     if (raw.commandFlags !== undefined) {
       if (!Array.isArray(raw.commandFlags) || raw.commandFlags.length > 16) {
--- a/src/main/hooks.ts
+++ b/src/main/hooks.ts
@@ -67,7 +67,10 @@ export class HookServer {
     private control?: ControlRegistry,
     /** Circuit breaker (Lane A #6.6b) — fed the hook-derived signals (session id,
      *  repeated identical tool calls). Optional so the server still runs without it. */
-    private breaker?: CircuitBreaker
+    private breaker?: CircuitBreaker,
+    /** Standing goal text for an agent (from the durable roster). Optional so
+     *  tests can omit it; when set, injected on SessionStart / UserPromptSubmit. */
+    private getStandingGoal?: (agentId: string) => string | null
   ) {}
 
   start(): void {
@@ -258,12 +261,22 @@ export class HookServer {
       && !!agentId && this.hive.isGod(agentId);
     const roster = wantsRoster ? this.hive.rosterContext() : null;
 
-    if (steer || roster) {
+    // Standing goal (hire Briefing) — durable roster field, re-read every cycle so
+    // an Edit Agent save is picked up on the next SessionStart / UserPromptSubmit
+    // without restarting the worker. Kept out of --append-system-prompt (volatile-
+    // free cache invariant); lives on the live hook channel instead.
+    const wantsGoal = (event === 'SessionStart' || event === 'UserPromptSubmit') && !!agentId;
+    const goalRaw = wantsGoal ? (this.getStandingGoal?.(agentId) ?? null) : null;
+    const goal = goalRaw
+      ? `<goal>\n${goalRaw}\n</goal>`
+      : null;
+
+    if (steer || roster || goal) {
       this.emit(agentId, event, p);
       return {
         hookSpecificOutput: {
           hookEventName: event,
-          additionalContext: [roster, steer].filter(Boolean).join('\n\n')
+          additionalContext: [roster, goal, steer].filter(Boolean).join('\n\n')
         }
       };
     }
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -258,9 +258,30 @@ let breakerBeatTimer: ReturnType<typeof setInterval> | null = null;
 // Feed the breaker's api_error-storm trip from Oscar's OTel api_error spans —
 // Jim's one breaker input with no on-branch source (telemetry.onApiError seam).
 telemetry.onApiError((agentId) => breaker.recordError(agentId));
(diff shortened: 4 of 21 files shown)


chaitanyagiri:
## TL;DR

Thanks @GBurchell27 — and thank you for the note about #145. Resubmitting clean rather than arguing the point is the right instinct, and this PR is visibly the better version: no `package.json` churn, no unrelated layout work, `cursor-agent` as the canonical binary with `agent` kept as an alias.

**What it does:** two things. (1) A **Cursor Agent CLI provider preset** — `cursor-agent`, interactive (no `-p`) so the TUI stays alive for hive mail via `seedDelivery: 'type-into-tui'`, `--force --trust` under auto-mode, `curl | bash` native install, god-eligible. (2) A **standing goal editor** — a new Edit Agent modal, with the goal injected via `additionalContext` on `SessionStart`/`UserPromptSubmit` for hook engines and prepended to queued PTY deliveries for engines without hooks.

**Verdict:** no malicious code, no NSFW, no prompt injection. Typecheck clean, **150/150 tests pass**, and `npm run build` **succeeds** — that's your unticked checklist item, now verified. You also updated `provider-config.test.cjs` for the new god-eligible provider, which is the exact step another provider PR in this queue ([#169](../pull/169)) missed. **One performance issue I'd fix before merge**, one token-cost question, and the missing screenshot.

---

## What I did

Checked out `feat/cursor-provider-standing-goals`, ran typecheck, the full test suite and a production build, read the preset against the repo's own conventions, and traced the standing-goal read path from `HookServer` down to the filesystem.

## Fix this first: a disk read on every turn

```
   agent submits a prompt
        │
        ▼
   HookServer.handle('UserPromptSubmit')
        │
        ▼
   standingGoalFromRoster(agentId)
        │
        ▼
   roster.read()
        ├─ existsSync(roster.json)
        ├─ readFileSync(roster.json, 'utf8')      ← synchronous
        └─ JSON.parse(...)                        ← the WHOLE roster
             (agents + archived + restorable + queues + notes + worktrees)
```

The hook server sits on the **critical path of the agent's turn** — the CLI blocks on the hook response. So every prompt from every agent now costs a synchronous read and full parse of a file that holds far more than the one string being looked up. On the machine I'm reviewing from, `roster.json` is **29 KB for 6 agents**; the queues and notes are what make it big, and it scales with fleet size and message backlog, not with the number of goals.

The comment explains the intent — *"re-read every cycle so an Edit Agent save is picked up without restarting the worker"* — and that requirement is right. Two ways to keep it without the cost:

1. **Cache on mtime.** `statSync().mtimeMs` and only re-parse when it changes. One `stat` per hook instead of a read+parse; the Edit Agent save still lands on the very next turn.
2. **Push instead of pull.** The renderer already knows when a goal changes — have it send goals to main on save, and keep a `Map<agentId, string>` in memory.

(1) is a five-line change and keeps your "always fresh" property exactly.

## The Cursor preset

This reads as carefully as the existing presets, and the reasoning is written down where a future maintainer will need it:

| Choice | Assessment |
|---|---|
| `defaultCommand: 'cursor-agent'` | ✅ canonical installer name; `agent` alias kept in `inferAgentProvider` |
| No `-p` (interactive TUI) | ✅ and the *why* is stated — print mode exits per turn, which would break god-eligibility and inbox drain |
| `initialPromptFlag: undefined` + `seedDelivery: 'type-into-tui'` | ✅ correct, and the Cobra-subcommand reasoning matches the Crush precedent exactly |
| `nativeInstallCommand` (`curl \| bash`, `irm \| iex`) | ✅ right field, right shape — matches Claude Code and OpenCode. Notably **not** invented as an npm package |
| `canReceiveInbox: true` with `hiveAware: false` | ✅ honest, and the comment says so |
| `autoModeFlag: '--force --trust'` | ✅ multi-token is fine (`grok` already uses `'--permission-mode bypassPermissions'`), and it's gated on `config.autoMode` |
| Gallery + hire spec updated (`hire.ts`, `docs/hires/validator.js`, `build-data.py`) | ✅ **thorough** — most provider PRs forget the public Agent Gallery entirely |

One thing to name plainly rather than object to: `--force` means tool calls run without confirmation and `--trust` skips the workspace-trust prompt. That's the same posture as codex's `--dangerously-bypass-approvals-and-sandbox`, it's behind the operator's own auto-mode toggle, and the comment says what each flag does. It should still be in the CHANGELOG when this ships, because "Cursor support" reads more innocuous than "Cursor support that can run unattended with trust prompts suppressed".

## What I tested

| Check | Result |
|---|---|
| `node --test test/*.test.cjs` | ✅ 150/150 |
| `npm run build` | ✅ **exit 0** — your unticked box |
| God-eligibility test updated for the new provider | ✅ yes — `provider-config.test.cjs` amended |
| New tests | ✅ `agent-provider` +26, `cli-install-ladder` +11, `provider-automation` adjusted |
| Mergeability | ✅ **MERGEABLE**, no conflicts |
| NSFW / profanity | ✅ none |
| Prompt injection | ✅ none *by the PR*. The standing goal is operator-authored text injected as `additionalContext` — that's an intended instruction channel, not an injection vector. Worth noting there's no length cap and no handling of a literal `</goal>` in the text; neither is exploitable by a third party, but see below |

## Two smaller things

**1. Injecting the goal on every `UserPromptSubmit` is a recurring token cost.** `SessionStart` alone would put it in context once per session. On every prompt, a 300-word goal is re-sent every turn for the whole session — and this app ships a circuit breaker with per-agent token caps, so that cost is visible to users. If per-turn re-injection is deliberate (goals that must survive compaction, say), that's a good reason — but it should be written in the comment, because the cheaper option looks obviously correct to a future reader.

**2. Missing screenshot.** `EditAgentModal.tsx` is 329 new lines of UI and CONTRIBUTING marks a screenshot **REQUIRED** for any visual change. You said one would follow — please do add it; it's also the fastest way for the maintainer to check the modal against `DESIGN.md`.

Minor: `withStandingGoal` skips prepending if `text.includes('<goal>')`. A queued message that happens to contain that literal would silently suppress the goal. A dedicated marker (`<!--md:goal-->`) or a per-delivery flag avoids that.

## Credit where it's due — this PR fixes a bug #154 only papers over

`usePtyParser.ts` writes live status into the durable `description` field in **four** places, and `mockEvents.ts` in one. **This PR deletes all five.** That is the actual root cause of the "on standby overwrote my hire role" problem that your other PR ([#154](../pull/154)) addresses with a `TRANSIENT_ROLE_RE` denylist. I flagged the missing root-cause fix on #154 before I'd read this one — the fix exists, it's just here.

Worth saying explicitly in both PRs, because the merge order matters: if **#153 lands first**, #154's regex is a one-time migration for rosters that already carry captions, which is a much easier thing to justify keeping. If #154 lands first, the denylist has to hold the line permanently against a writer that's still firing every four seconds.

## Questions

@GBurchell27 — (1) will you switch `standingGoalFromRoster` to an mtime-cached read? (2) Is per-`UserPromptSubmit` injection deliberate, or would `SessionStart` do? (3) Are `gpt-5.6-luna-high` / `composer-2.5` verified against a real `cursor-agent --help`, or curated guesses? The repo's convention for guesses is an explicit `// TODO-verify` marker (see `COPILOT_MODELS`). (4) You offered to split the standing-goal editor out — I'd only take that if the maintainer wants Cursor to land faster; the two are independently reviewable as they stand.

@chaitanyagiri — one decision: **merge #153 before #154** for the reason above. And a sanity check from you on the Cursor auto-mode posture (`--force --trust`) before it ships, since it's the first preset that suppresses a *trust* prompt as well as tool confirmations.

PR #154, opened 2026-08-17
fix(hive): keep hire role out of live status captions

GBurchell27:
## Summary
- Roster `description` is the durable hire/job one-liner (same string as hive `registry.role`). Live state belongs on `status` / `action`. Status captions like `"on standby"` were able to overwrite that job string in the floor roster, and a restart passed it back into `registry.json` + `identity.md`.
- `ensureAgent` now keeps the last real role (and capabilities) when a respawn sends a transient caption. Restart / restore / wake-respawn omit a status-like `description` instead of sending it as `role`.
- Boot heals roster descriptions from `registry.role` when the floor caption looks like status. Add Agent labels the field **Role**. The demo mock loop no longer writes live action into `description`.

## Test plan
- [ ] Hire an agent with a real role one-liner (or import a hire JSON). Confirm Add Agent shows **Role**, not Description.
- [ ] Pause and resume the agent. Open the floor card / any role field: it still shows the hire one-liner, not `"on standby"` / idle.
- [ ] Restart & Continue (or quit and restore the team). `hive/registry.json` `role` and `hive/agents/<id>/identity.md` still have the hire role; capabilities are unchanged.
- [ ] If a roster already has `"on standby"` but the hive registry still has the real role, reload the app once and confirm the floor description is healed from the registry.
- [ ] `node --test test/agent-role.test.cjs` and `npm run typecheck`

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "typecheck:node": "tsc --noEmit -p tsconfig.node.json",
     "typecheck:web": "tsc --noEmit -p tsconfig.web.json",
     "typecheck": "npm run typecheck:node && npm run typecheck:web",
-    "test:focused": "node --test test/codex-remote.test.cjs test/control.test.cjs test/provider-automation.test.cjs test/provider-config.test.cjs test/queue-delivery.test.cjs test/terminal-automation.test.cjs test/terminal-recovery.test.cjs test/roster.test.cjs test/proc-kill.test.cjs test/expand-tilde.test.cjs test/hive-cwd.test.cjs test/hive-hook-node.test.cjs test/hive-roster-injection.test.cjs test/hive-runtime-path.test.cjs test/cli-install-ladder.test.cjs test/node-install.test.cjs test/office-gl-recovery.test.cjs test/update-state.test.cjs test/transcript-project-dir.test.cjs test/commit-graph.test.cjs",
+    "test:focused": "node --test test/codex-remote.test.cjs test/control.test.cjs test/provider-automation.test.cjs test/provider-config.test.cjs test/queue-delivery.test.cjs test/terminal-automation.test.cjs test/terminal-recovery.test.cjs test/roster.test.cjs test/proc-kill.test.cjs test/expand-tilde.test.cjs test/hive-cwd.test.cjs test/hive-hook-node.test.cjs test/hive-roster-injection.test.cjs test/hive-runtime-path.test.cjs test/cli-install-ladder.test.cjs test/node-install.test.cjs test/office-gl-recovery.test.cjs test/update-state.test.cjs test/transcript-project-dir.test.cjs test/commit-graph.test.cjs test/agent-role.test.cjs",
     "check:links": "node tools/check-release-links.cjs",
     "dist": "npm run build && electron-builder",
     "dist:mac": "npm run build && electron-builder --mac",
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -5,8 +5,9 @@
  * process commits to (agents never call git — they just write files). See
  * HIVE.md for the full design. Responsibilities:
  *   - per-agent workspace (identity.md, memory.md, inbox/, outbox/, cursor.json)
- *   - a roster (registry.json), shared blackboard (board.md), task ledger,
- *     and an append-only event log (log.jsonl)
+ *   - hive identity (registry.json: id/role/cwd/session — what agents read),
+ *     separate from the UI floor roster (`<harnessHome>/roster.json`)
+ *   - shared blackboard (board.md), task ledger, and an append-only event log (log.jsonl)
  *   - a router that drains each agent's outbox into recipients' inboxes
  *
  * Human-in-the-loop is native to each agent's Claude Code session: permission
@@ -36,6 +37,7 @@ import {
   type AgentProvider
 } from '../shared/agentProvider';
 import { MCP_CATALOG } from '../shared/mcpCatalog';
+import { preferredAgentRole } from '../shared/agentRole';
 import { expandTilde } from './fs';
 
 /** The subset of HarnessConfig the hive consumes for the default-MCP merge.
@@ -548,6 +550,15 @@ export class HiveManager {
     mkdirSync(join(dir, 'inbox', '.done'), { recursive: true });
     mkdirSync(join(dir, 'outbox', '.sent'), { recursive: true });
 
+    // Resolve role BEFORE writing identity.md. A restart passes the floor
+    // roster's `description`, which can be a status caption ("on standby").
+    // identity.md and registry.role are the durable job from the hire.
+    const reg = this.registry();
+    const prev = reg.agents[meta.id];
+    if (meta.cwd) meta = { ...meta, cwd: expandTilde(meta.cwd) };
+    const role = preferredAgentRole(meta.role, prev?.role, !!meta.isGod);
+    meta = { ...meta, role };
+
     const identity = join(dir, 'identity.md');
     writeFileSync(identity, this.identityText(meta), 'utf8'); // refresh on each spawn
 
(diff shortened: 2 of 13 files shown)


chaitanyagiri:
## TL;DR

Thanks @GBurchell27 — the distinction you're drawing is the right one, and it's worth making: **`description` / `registry.role` is the durable job from the hire; live run-state belongs on `status` / `action`.** Documenting that in `roster.ts` and `hive.ts`, adding `preferredAgentRole`, keeping the last real role across respawns, and relabelling the Add Agent field **Role** are all good calls. The tests pass and the typecheck is clean — I ran the plan you left unticked.

**But the PR treats the symptom and leaves the cause running.** The exact string the PR is named after is written into `description` by a line this PR doesn't touch:

```
src/renderer/src/hooks/usePtyParser.ts:66     description: 'on standby',
```

That fires on **every** 4-second idle drift, for every agent, all session long. So after this merges, the floor roster still gets clobbered exactly as before — the healing effect only runs once at boot, so a user sees the hire role until the agent first goes idle, then "on standby" for the rest of the session. The hive registry *is* protected (that part works), but the floor is not.

**Verdict:** no malicious code, no NSFW, no prompt injection. 152/152 tests pass, typecheck clean, conflicts only on `package.json`. **One line fixes the cause**, and it changes what the rest of the PR is *for*.

---

## What I did

Checked out `fix/preserve-agent-role`, ran typecheck and the full suite, ran your `agent-role` tests, tested the merge, then grepped `main` for every writer of a status string into `description` to find out whether the denylist is a migration or a permanent patch.

## The cause is still there

```
   agent goes quiet 4s
        │
        ▼
   usePtyParser.scheduleIdle()
        updateAgent(id, { status:'idle', action:'awaiting',
                          description:'on standby' })   ← ✗ still in this PR
        │
        ▼
   store.updateAgent persists the roster
        │
        ▼
   roster.json description = "on standby"     ← the bug, re-created every idle
        │
        ├─ hive registry:  protected ✅ (roleForHiveSpawn omits it on respawn)
        └─ floor card:     clobbered ✗ (heal effect only runs at boot)
```

Both writers of the literal survive on your branch — `usePtyParser.ts:66` and `mockEvents.ts:164`. You removed the *other* mockEvents writer (`description: sample.what` in `stepAgent`), which is the right instinct; these two are the same mistake and `usePtyParser` is the one that runs in production. Deleting `description: 'on standby'` from that `updateAgent` call is the whole fix: `status: 'idle'` and `action: 'awaiting'` in the same patch already carry the information, on the fields that are meant to hold it.

With that line gone, `TRANSIENT_ROLE_RE` and `syncDescriptionsFromRoles` become a **one-time migration** for rosters already carrying a caption — which is a much easier thing to defend than a permanent filter.

## Why the denylist shouldn't be permanent

```js
/^(on\s+)?standby$|^(idle|awaiting|paused|resumed|working|thinking|archived|starting up|reconnecting…?|running the floor|a fresh harness)$/i
```

Compare that with the actual status vocabulary in `PixelBadge.tsx`:

```ts
export type StatusKind =
  | 'idle' | 'thinking' | 'working' | 'waiting' | 'blocked' | 'success' | 'ghost'
  | 'compacting' | 'looping';
```

**`waiting`, `blocked`, `success`, `ghost`, `compacting` and `looping` are not in the regex.** If any writer ever puts one of those in `description`, it passes as a durable role and gets written into `registry.json` and `identity.md` as the agent's job. A hand-maintained list of strings will always trail the enum it's shadowing. If the denylist survives at all, derive it from `StatusKind` plus the handful of genuine captions, so adding a status can't silently widen the hole.

Two more things about it worth knowing:

- **`'a fresh harness'` is the Add Agent form's own default** (`useState(pendingHire?.description ?? 'a fresh harness')`, and `description.trim() || 'a fresh harness'` on submit). Filtering it means a user who leaves the field untouched silently gets the role `agent` in the hive. That's probably what you want — but it's a real behaviour change and it isn't mentioned.
- **`preferredAgentRole('on standby', 'idle', true)` returns `'on standby'`**, which your test enshrines. The `isGod` default is only reached when both inputs are *empty*, never when both are non-empty-but-transient. Since the function's entire job is "transient loses", falling through to `'orchestrator (god)'` / `'agent'` when *both* sides are transient would be more consistent. The spawn path dodges this by passing the god default as `fallback`, but `syncDescriptionsFromRoles` calls it directly with the registry value.

## The healing effect writes to git on every boot

Effect 0 calls `hivePatchAgentRole` per agent whose role differs, and `patchAgentRole` ends with `this.commit(\`hive: role ${id}\`)`. On a twenty-agent fleet that's up to **twenty hive commits at every app start**. Given #128 in this same queue documents a hive `.git` reaching 7.5 GB and an auto-gc taking 22 GB of RAM, per-agent commits in a boot loop are worth avoiding. Batch the registry writes and commit once, or skip the commit when the patch is a heal rather than an operator edit.

## What I tested

| Check | Result |
|---|---|
| `npm run typecheck` (node + web) | ✅ exit 0 — your checklist item, now ticked |
| `node --test test/agent-role.test.cjs` | ✅ 3/3 — likewise |
| `node --test test/*.test.cjs` | ✅ 152/152 |
| `test:focused` includes the new test | ✅ yes |
| Cause still present on the branch | ❌ `usePtyParser.ts:66` and `mockEvents.ts:164` both still write `'on standby'` |
| Denylist vs `StatusKind` | ⚠️ 6 of 9 status values unlisted |
| Mergeability | ❌ **CONFLICTING** — `package.json` only (the stale `test:focused` line; re-add `agent-role.test.cjs` to `main`'s current list) |

The parts I'd keep unchanged: `roleForHiveSpawn` returning `undefined` so `ensureAgent` falls back to `prev.role`, `capabilities: meta.capabilities ?? prev?.capabilities ?? []` (that fixes a silent capability wipe on respawn that isn't even mentioned in the description — worth calling out, it's a real fix), and the two doc-comment rewrites explaining roster-vs-registry. That last one is the most valuable thing here long-term.

## Questions

@GBurchell27 — (1) was `usePtyParser.ts:66` a deliberate omission, or missed? If you delete it, does anything on the floor lose text it needs — i.e. is the card rendering `description` where it should be rendering `action`? (2) Is filtering `'a fresh harness'` intentional given it's the form's own default? (3) Could you batch the heal into one hive commit?

@chaitanyagiri — one for you: with the cause removed, `TRANSIENT_ROLE_RE` becomes a **migration** for rosters already carrying captions. Do you want it kept permanently as belt-and-braces (accepting that it will trail `StatusKind`), or run once and deleted? I'd vote run-once-and-delete — a regex that guesses whether a user's own text is "really" a job is the kind of thing that eventually eats someone's legitimate role string.

chaitanyagiri:
**Correction to my review above — an important one, in your favour @GBurchell27.**

I wrote that this PR "leaves the cause running" because `usePtyParser.ts:66` still writes `description: 'on standby'`. That's factually true of *this branch* — but I've since reviewed **#153**, and it deletes **all four** of those writers in `usePtyParser.ts` plus the one in `mockEvents.ts`. You did fix the root cause; it's in the sibling PR, not missing.

So the accurate framing is about **merge order**, not omission:

- **#153 first** → the writers are gone, and `TRANSIENT_ROLE_RE` here becomes a **one-time migration** for rosters that already carry a caption. Easy to justify, easy to delete later.
- **#154 first** → the denylist has to hold the line permanently against a writer still firing every four seconds, and it will keep trailing `StatusKind` as new captions appear.

I've recommended #153-before-#154 on the other PR too.

Everything else in my review stands unchanged, and is worth a look independently of ordering:

1. `TRANSIENT_ROLE_RE` misses **6 of the 9** `StatusKind` values (`waiting`, `blocked`, `success`, `ghost`, `compacting`, `looping`) — derive it from the enum rather than hand-listing, or drop it once #153 lands.
2. `'a fresh harness'` is the Add Agent form's own default, so filtering it silently turns an untouched field into the role `agent`.
3. The boot heal calls `hivePatchAgentRole` per agent, and each one ends in its own `hive.commit()` — up to N hive commits on every app start. Worth batching, especially given #128 in this same queue.
4. `preferredAgentRole(transient, transient, isGod)` returns the first transient value rather than falling through to the god/agent default.

Sorry for the misleading emphasis in the original — reviewing twenty PRs in sequence, I hadn't yet reached the one that contained the fix.

cc @chaitanyagiri

PR #155, opened 2026-08-17
fix(hive): let a broadcast reach agents on hookless providers

tsurutanmen:
## The bug

Broadcast fan-out gated its target list on `canReceiveInbox`, so an agent running
the `custom` provider **never heard a `to: 'broadcast'` message** — while a
**direct** message to that very same agent was delivered fine, as a terminal work
order.

That asymmetry is the surprising part. From the office floor it looks like the
agent is present and reachable (you can mail it, it answers), but every broadcast
quietly skips it.

## Why the gate isn't needed

`deliver()` already knows how to serve a hookless target. For any target where
`canReceiveInbox` is false it calls `emitTerminalHandoff(msg, t)` and only bounces
to the god when the renderer is unavailable:

```ts
if (t !== godId && !canReceiveInbox(reg.agents[t]?.provider)) {
  if (!this.emitTerminalHandoff(msg, t)) {
    this.deliver({ ...msg, to: godId, subject: `[undeliverable — …]` }, godId);
  }
  continue;
}
```

So the extra gate on the fan-out list was redundant for the providers it let
through, and lossy for the ones it excluded.

Fan-out now selects the agents direct mail can reach — skipping the sender, the
send-only prep assistant, and archived agents, exactly as before — and the
existing per-target path decides *how* each one is served.

## Trade-off

With the renderer down, a broadcast to N hookless agents produces N bounces to the
god instead of silence. That is the same thing N direct messages already produce
today, and it is the loud failure rather than the quiet one — nothing is dropped
without the god being told.

## Shape of the change

The rule moved into `src/shared/broadcast.ts` as a pure function, so it is
testable on its own the way `queueDelivery` and `codexRemote` are. `hive.ts` calls
it and keeps everything else unchanged. The old comment in `hive.ts` explained the
hookless exclusion, so it is rewritten rather than left to describe behaviour that
no longer exists.

## Testing

- `npm run typecheck` — exit 0.
- 6 new tests in `test/broadcast-targets.test.cjs`, all passing.
- `npm run test:focused` — the set of failing tests is **identical before and
  after** this change (8 failures, all pre-existing).

### On those 8 pre-existing failures

I developed this on **Windows**, which `CONTRIBUTING.md` notes is untested. On a
clean checkout of `main`, `npm run test:focused` already fails 8 tests here, all
of them platform issues rather than logic ones:

- `codex-remote` — asserts a `unix://` socket path
- `cli-install-ladder` / `node-install` — asserts the npm/native installer rungs
- `transcript-project-dir` (4) — asserts POSIX path normalization, e.g. expects
  `-Users-me-MDv0-3-0` and gets a Windows path

I stashed this change and re-ran to confirm the same 8 fail without it. Happy to
be told if any of those are already known.

## What this PR does not do

No UI change, so no screenshot. It does not give `custom` a bridge — a hookless
agent still has no safe-idle lifecycle state, and still receives via the terminal
work-order path rather than the guarded inbox. This only stops broadcasts from
skipping it.

### DIFF
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -36,6 +36,7 @@ import {
   type AgentProvider
 } from '../shared/agentProvider';
 import { MCP_CATALOG } from '../shared/mcpCatalog';
+import { selectBroadcastTargets } from '../shared/broadcast';
 import { expandTilde } from './fs';
 
 /** The subset of HarnessConfig the hive consumes for the default-MCP merge.
@@ -1142,15 +1143,11 @@ export class HiveManager {
     const resolveTo = (to: string): string => (to === 'human' || to === 'god' ? godId : to);
     const targets = msg.to === 'broadcast'
       // The roster for fan-out is the ACTIVE registry: skip the send-only prep
-      // assistant, any archived agent (closed tab), and providers that can't
-      // expose safe-idle lifecycle state (hookless custom commands), so mail never
-      // piles into a dead inbox. Claude, Codex and Antigravity are included; their
-      // hooks let the renderer wake them only after a safe idle boundary.
-      ? Object.keys(reg.agents).filter((a) =>
-          a !== msg.from
-          && !reg.agents[a]?.isAssistant
-          && !reg.agents[a]?.archived
-          && canReceiveInbox(reg.agents[a]?.provider))
+      // assistant and any archived agent (closed tab). Hookless providers are
+      // NOT skipped — the per-target path below already serves them a terminal
+      // work order, so excluding them here only made a broadcast invisible to an
+      // agent that direct mail reaches fine. See selectBroadcastTargets.
+      ? selectBroadcastTargets(reg.agents, msg.from)
       // Never deliver to self — guards a god → "human" message looping back to god.
       : [resolveTo(msg.to)].filter((t) => t !== msg.from);
     for (const t of targets) {
--- a/src/shared/broadcast.ts
+++ b/src/shared/broadcast.ts
@@ -1,0 +1,48 @@
+/**
+ * Broadcast fan-out target selection.
+ *
+ * Kept as a pure function (and out of `hive.ts`) so the rule is testable on its
+ * own, the way `queueDelivery` / `codexRemote` are.
+ */
+
+/** The subset of a registry agent that fan-out actually looks at. */
+export interface BroadcastCandidate {
+  /** Michael's prep assistant — send-only, drains no inbox. */
+  isAssistant?: boolean;
+  /** PTY tab closed; the record is retained but the agent is not live. */
+  archived?: boolean;
+}
+
+/**
+ * The agents a `to: 'broadcast'` message fans out to.
+ *
+ * Excluded: the sender itself, the send-only prep assistant (direct mail to it
+ * would rot unread), and archived agents (no live PTY).
+ *
+ * NOT excluded: providers without a hook/proxy bridge. Fan-out used to gate on
+ * `canReceiveInbox`, which meant an agent on a hookless provider — `custom`,
+ * i.e. any CLI the presets don't know — silently never heard a broadcast, even
+ * though a DIRECT message to that same agent was delivered fine. That asymmetry
+ * was the bug: `deliver` already routes a hookless target through
+ * `emitTerminalHandoff` (a terminal work order) and only bounces to the god when
+ * the renderer is unavailable. Fan-out now selects the same agents direct mail
+ * can reach, and that existing per-target path decides HOW each one is served.
+ *
+ * The trade-off this accepts: with the renderer down, a broadcast to N hookless
+ * agents produces N bounces to the god instead of silence. That is the same
+ * behaviour N direct messages already produce, and it is the loud failure —
+ * nothing is dropped without the god being told.
+ */
+export function selectBroadcastTargets(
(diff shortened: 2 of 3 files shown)


chaitanyagiri:
## TL;DR

Thanks @tsurutanmen — this is a careful, well-argued PR and I'd merge it after one question is answered.

**What it does:** broadcast fan-out gated its target list on `canReceiveInbox`, so an agent on a hookless provider (`custom` — any CLI the presets don't know) **never heard a `to: 'broadcast'` message**, while a *direct* message to that same agent was delivered fine as a terminal work order. The asymmetry is the bug: from the floor the agent looks present and reachable, but every broadcast silently skips it. The fix drops the redundant gate, moves the rule into a pure `selectBroadcastTargets()` in `src/shared/broadcast.ts`, and lets the existing per-target path decide *how* each agent is served.

**Verdict:** no malicious code, no NSFW, no prompt injection, no dependency changes. Typecheck green, 155/155 tests pass, `main` still has the gate at `hive.ts:1205`, and — importantly — **I chased the one risk this change could plausibly carry and it doesn't apply.** Details below, because it's the thing a reviewer would reasonably worry about and the PR description doesn't address it.

---

## What I did

Checked out `fix/broadcast-reaches-hookless-agents`, ran typecheck and the full suite, confirmed the gate is still on `main`, and then traced `emitTerminalHandoff` all the way into the renderer to check whether removing the gate could inject text into a hookless agent mid-turn.

## The risk I checked, and why it's fine

The gate you removed was justified by a comment about providers that *"can't expose safe-idle lifecycle state"* — hooks *"let the renderer wake them only after a safe idle boundary."* The obvious worry is that a broadcast now types a work order into N hookless agents' REPLs while they're mid-generation, corrupting their input. It doesn't:

```
   hive.deliver → emitTerminalHandoff → IPC 'hive:terminalHandoff'
                                              │
                                              ▼
        useHive.ts  onHiveTerminalHandoff((msg) => …)
                                              │
                    ┌─────────────────────────┴───────────────────┐
                    │                                             │
          target has a live PTY                          no live PTY
                    │                                             │
          enqueueMessage(target, workOrder)            enqueueMessage(GOD, "handoff failed")
                    │
                    ▼
      "queued through the same idle-only PTY drain
       as human-composed messages"     ← the comment in useHive.ts
```

So work orders go through the same **idle-gated queue** a human message does; they are never typed straight into a busy REPL. The handler also dedupes twice — a `seenTerminalHandoffs` set on `msg.id`, plus a `Message: <id>` marker check against the existing queue — so a broadcast can't double-queue. Combined with the provider-agnostic PTY-quiescence fallback that sits right below it in the same file, hookless agents do get a usable idle signal.

That's the argument your PR description is missing, and it's the one that makes the change safe rather than merely symmetrical. Worth adding to the commit message.

## What I tested

| Check | Result |
|---|---|
| `node --test test/*.test.cjs` | ✅ 155/155 |
| `test/broadcast-targets.test.cjs` | ✅ 6/6 |
| Bug still present on `main` | ✅ `hive.ts:1205` still filters on `canReceiveInbox(reg.agents[a]?.provider)` |
| Per-target hookless path still intact | ✅ `hive.ts:1228` unchanged — `emitTerminalHandoff`, bounce to god on failure |
| Mergeability | ✅ **MERGEABLE**, no conflicts (branch is 40 commits behind but the surrounding code hasn't moved) |

Extracting the rule into `src/shared/` as a pure function is the right call and matches how `queueDelivery` and `codexRemote` are structured. The `BroadcastCandidate` interface being *narrower* than the registry entry is also good design — the function can only look at what it's allowed to.

## Two notes

**1. The test fixture doesn't model what its comments claim.** `toby` is commented *"A worker on a hookless provider (`custom`): no hook bridge, no proxy"* — but `BroadcastCandidate` has no `provider` field, so `toby` is just `{}`, identical to `dev1`. The test is correct and the regression comment is accurate about *history*, but as written the fixture can't distinguish hookless from hooked, and a future reader will assume it does. Either drop the provider framing from the fixture comments, or (better) keep `provider: 'custom'` on `toby` as documentation even though the function ignores it.

**2. The bounce amplification is slightly broader than described.** You cover "renderer down → N bounces to god instead of silence", and I agree that's the loud failure. There's a second case: renderer **up**, but a target has no live PTY (crashed, `archived` not yet set) — `useHive` then enqueues a *"Terminal handoff failed for X"* message to god. Previously hookless agents in that state were skipped entirely, so each broadcast now costs god one message per such agent. Still the right direction, still bounded by roster size, but worth a line in the trade-off section since it happens with the renderer perfectly healthy.

## Questions

@tsurutanmen — the one thing I couldn't settle from the code: with the gate gone, is there any provider in `AGENT_PROVIDER_PRESETS` where `canReceiveInbox: false` means *"the CLI is not interactive at all"* rather than *"it has no hook bridge"*? Copilot is spawned in print mode (`initialPromptFlag: '-p'`, `canReceiveInbox: false`) — for a one-shot print-mode process, a queued work order has nothing to be typed into. If so, that provider may still need excluding, on a different criterion than the one you removed. Worth checking `kimi` too.

@chaitanyagiri — sign-off question rather than a code one: this makes broadcast reach **every** live non-archived agent, including `custom` commands you've never presetted. That's what "broadcast" should mean, and it's what people will expect. Just confirming you want the floor-wide semantics rather than "broadcast means everyone the hive can wake via hooks", because the old comment reads like the latter was deliberate.

PR #157, opened 2026-08-17
fix(pty): strip inherited Claude Code session markers from agent PTY env

gpechieu:
## Symptom
Every agent spawned in a run showed **"Transcript saving is off — inherited CLAUDE_CODE_CHILD_SESSION marker"**; no agent session ever reached `~/.claude/projects/<cwd-slug>/`, so `--resume` could never find anything — restores silently fell back to fresh terminals with no conversational context.

## Root cause
The app is often launched from *inside* a Claude Code session (`npm run dev` typed into a claude terminal — presumably common for anyone developing this app with Claude). That parent session's identity markers arrive via `process.env` and flowed into every agent PTY: `CLAUDE_CODE_CHILD_SESSION` makes the agent CLI believe it is a child session and silently disables transcript saving, and `CLAUDE_CODE_SESSION_ID` / `CLAUDE_PID` / messaging socket+token leak the *parent's* identity into fresh agents.

## Fix
Build the child env explicitly and delete the five markers before `pty.spawn`. Agents are top-level sessions regardless of how the app was launched.

## Validation
- Runtime: after restarting the app, `ps eww` shows the child marker on zero agent processes, and agents' session `.jsonl` transcripts appeared on disk for the first time — resume is functional from then on.


chaitanyagiri:
## TL;DR

Thanks @gpechieu — this is a real bug with a correct diagnosis, and the symptom you describe (every agent silently losing its transcript, so `--resume` always fell back to a fresh terminal) is nasty precisely because nothing looks broken. Good find.

**What it does:** the app is often launched from *inside* a Claude Code session (`npm run dev` typed into a claude terminal). That parent session's identity markers arrive via `process.env` and flow into every agent PTY. `CLAUDE_CODE_CHILD_SESSION` in particular makes the agent CLI believe it's a child session and **silently disables transcript saving**. The fix builds the child env explicitly and deletes five markers before `pty.spawn`.

**Verdict:** no malicious code, no NSFW, no prompt injection, no dependency changes. Typecheck green, 149/149 tests pass. **Three things to address before merge:** the branch conflicts with `main`, the delete happens in the wrong order relative to the app's own env injection, and — verified from a live Claude Code session — **the deny-list is missing seven more markers**, including two that matter.

---

## What I did

Checked out the branch, ran typecheck and the full suite, tested the merge against current `main`, and then dumped the actual `CLAUDE*` environment of a live Claude Code session to check the deny-list against reality rather than against memory.

## The leak, and what's still leaking

```
   you type `npm run dev` inside a Claude Code session
        │
        │  process.env carries the PARENT session's identity
        ▼
   Electron main
        │
        │  { ...process.env, PATH, TERM, …, ...opts.env }
        ▼
   pty.spawn  →  agent CLI sees CLAUDE_CODE_CHILD_SESSION
                      │
                      ▼
              "I am a child session" → transcript saving OFF
                      │
                      ▼
              nothing in ~/.claude/projects/<slug>/  →  --resume finds nothing
```

I dumped `env | grep ^CLAUDE` inside a live session. There are **twelve** markers, not five. Your list catches five:

✅ `CLAUDE_CODE_CHILD_SESSION` · `CLAUDE_CODE_SESSION_ID` · `CLAUDE_PID` · `CLAUDE_CODE_MESSAGING_SOCKET` · `CLAUDE_CODE_MESSAGING_TOKEN`

Still inherited by every agent:

| Marker | Why it matters |
|---|---|
| `CLAUDE_CODE_FORCE_SESSION_PERSISTENCE` | **Same family as the bug you're fixing** — it's session-persistence behaviour, inherited from the parent |
| `CLAUDE_EFFORT` | The parent's reasoning-effort setting silently applied to every spawned agent. That's a behaviour *and cost* leak the operator never chose |
| `CLAUDE_CODE_EXECPATH` | Path to the parent's `claude` binary — an agent could resolve the parent's binary instead of the one on its own PATH |
| `CLAUDECODE` | The generic "running under Claude Code" flag; anything that branches on it branches wrongly |
| `CLAUDE_CODE_ENTRYPOINT` | Parent's entrypoint identity |
| `CLAUDE_CODE_ENABLE_TELEMETRY` | Parent's telemetry choice applied to agents |
| `CLAUDE_REMOTE_CONTROL_SESSION_NAME_PREFIX` | Parent's remote-control naming |

A hardcoded list of five will keep drifting as the CLI adds markers — and it already has. I'd strip by **prefix** instead: remove everything matching `/^CLAUDE(CODE|_)/` from the inherited `process.env`, then layer the app's own intentional values on top. That's self-maintaining and expresses the actual rule: *agents are top-level sessions, they inherit none of the parent's Claude identity.*

## The ordering issue

Right now the deletes run **after** `...(opts.env ?? {})` is spread in:

```ts
const childEnv = { ...process.env, PATH, TERM, COLORTERM, FORCE_COLOR, ...(opts.env ?? {}) };
for (const k of [...]) delete childEnv[k];      // ← also deletes anything opts.env set deliberately
```

So if the app (or a user, once #108's per-agent environment variables land) ever sets one of those keys on purpose, it gets silently wiped. The strip should apply to the **inherited** layer only:

```ts
const inherited = { ...process.env };
for (const k of Object.keys(inherited)) if (/^CLAUDE(CODE|_)/.test(k)) delete inherited[k];
const childEnv = { ...inherited, PATH, TERM, COLORTERM, FORCE_COLOR, ...(opts.env ?? {}) };
```

Same behaviour today, correct behaviour tomorrow, and it reads as the rule rather than as a list.

## What I tested

| Check | Result |
|---|---|
| Bug still present on `main` | ✅ confirmed — no `CLAUDE_CODE_CHILD_SESSION` handling anywhere in `main`'s `pty.ts` |
| Deny-list completeness | ❌ 5 of 12 markers stripped (table above) |
| Mergeability | ❌ **CONFLICTING** — `src/main/pty.ts`. `main` has since added locale handling (`LANG`/`__CF_USER_TEXT_ENCODING`) to the same env literal, so it needs a rebase |

## On the missing test

This is the one I'd push on. The repo tests exactly this kind of thing — `win-cmd-shim.test.cjs` and `hive-windows-prompt.test.cjs` both exist to pin spawn-time behaviour, and `buildMissingCliScript()` in `cliInstall.ts` takes `platform` as an explicit parameter *specifically* so it can be tested off-platform. The same trick works here: pull the env construction into an exported pure function —

```ts
export function buildPtyEnv(parentEnv, userPath, agentEnv) { … }
```

— and a five-line test can assert that a parent env containing all twelve markers produces a child env containing none, while `agentEnv` values survive. Without that, this fix has no defence against someone re-simplifying the env literal back into a single spread. Given it was invisible for a whole run of the app before you caught it by hand, that defence is worth having.

## Questions

@gpechieu — (1) any objection to switching to the prefix strip? It subsumes your five and the seven I listed. (2) Was `CLAUDE_EFFORT` deliberately left in — is there a case where an agent *should* inherit the parent's effort setting? I can't think of one, but you've been running this as a daily driver and I haven't. (3) Could you rebase onto `main` and fold the env literal into an exported helper with a test while you're in there?

@chaitanyagiri — flagging an interaction: this PR and **#108 (per-agent environment variables)** both rewrite the same `pty.spawn` env block, and #108 is also conflicting. Whichever lands second will have to redo this work. Worth deciding an order — my suggestion is this one first, since it's smaller and it defines the layering (`inherited → app defaults → per-agent`) that #108 then plugs into.

gpechieu:
All three asks are in — branch rebased onto current `main` and force-pushed. Answers in order:

**(1) Prefix strip: adopted, with one refinement.** Everything matching `/^CLAUDE(CODE|_)/` is now stripped from the inherited layer, which subsumes my five and your seven. The refinement: a small **config-not-identity keep-list** — `CLAUDE_CONFIG_DIR`, `CLAUDE_CODE_OAUTH_TOKEN`, `CLAUDE_CODE_USE_BEDROCK`, `CLAUDE_CODE_USE_VERTEX`. Those share the prefix but are the *operator's* configuration rather than the parent session's identity: an operator who authenticates via the OAuth token env var, or points the CLI at an alternate config dir, or routes through Bedrock/Vertex, wants agents to see exactly those values — and a blanket prefix strip would break them in the same quiet way this PR exists to prevent. The criterion is stated in the module comment so future additions have a rule to follow, and a test pins that the keep-list doesn't weaken the strip.

**(2) `CLAUDE_EFFORT` was not deliberate.** The original five were derived from the observable failure (transcript loss and messaging) rather than from a dump of a live session, which is exactly the drift you called out. I can't think of a case where an agent should inherit the parent's effort either — it's stripped by the prefix rule now, and it's one of the twelve markers the new test asserts against.

**(3) Done.** The whole construction now lives in an exported pure `buildPtyEnv(parentEnv, userPath, agentEnv, platform)` in `src/main/ptyEnv.ts` — `platform` as a parameter following the `buildMissingCliScript` precedent, so the win32/posix locale branch is testable off-platform. The ordering bug you spotted is fixed structurally: the strip applies to the inherited layer **only**, then app defaults (including `main`'s locale block, preserved verbatim inside the helper), then `opts.env` — so per-agent values always win, which is the layering #108 can plug straight into. `test/pty-env.test.cjs` pins: all twelve live-dump markers stripped, future markers caught by prefix, keep-list survives, `agentEnv` survives both the strip and the defaults, and the locale matrix. `npm run typecheck` clean, **265/265** tests pass on the rebased branch.

PR #158, opened 2026-08-17
fix(memory): NaN embeddings on Apple Silicon — pin mempalace device to cpu on macOS

gpechieu:
## Symptom
The memory miner failed every 3-minute cycle for every agent: `[memory] mine <id> exited 1 … chromadb: Embeddings must not contain NaN or Infinity values`. Zero memories ever got indexed, and agents' own `mempalace search` was equally broken (queries embed through the same path).

## Root cause
mempalace's `embedding_device=auto` selects the CoreML execution provider on Apple Silicon. CoreML runs the quantized embeddinggemma ONNX graph partially (330/1647 nodes) and its fp16 partitions overflow → every vector is NaN for **any** input (even "hello world"). Deterministically reproduced: the same input embeds to NaN under `CoreMLExecutionProvider` and cleanly under CPU.

## Fix
Set `MEMPALACE_EMBEDDING_DEVICE=cpu` on darwin in both the mine loop's child env and the env merged into each agent's spawn. Other platforms keep mempalace's own default. (Arguably also a mempalace bug — it should guard NaN output — but this is the app-level fix so shared memory works out of the box on Apple Silicon, likely most of this app's audience.)

## Validation
- Runtime with cpu pinned: mine went from 0/19 files (abort) to 19/19 for god, full palace rebuilt across all agents (~646 drawers), recall verified end-to-end (`mempalace search` returns the right documents, cosine 0.809), and post-restart logs show zero miner errors.

## Updates after review

- **No re-index needed:** switching device does not invalidate an existing palace — same model, same 384-dim vectors (confirmed against mempalace's own `embedding.py` note).
- **Escape hatch added:** an explicitly exported `MEMPALACE_EMBEDDING_DEVICE` always wins; the pin emits nothing in that case, which also keeps `=coreml` available as the one-command NaN repro.
- **Scope stays macOS-wide, now with the evidence:** minilm rides chromadb's `ONNXMiniLM_L6_V2`, whose model build unconditionally removes `CoreMLExecutionProvider` — so minilm never ran on CoreML with or without this pin, and embeddinggemma is the only path that reaches it (and NaNs there). The pin's only observable effect is on the broken path.
- Resolution extracted to an exported pure `mempalaceDevice(platform, envOverride)` with `test/mempalace-device.test.cjs` (darwin→cpu, others→default, override-wins, empty-override-is-no-choice).


chaitanyagiri:
## TL;DR

Thanks @gpechieu — the investigation here is excellent. "Every vector is NaN, so chroma rejects every upsert, so nothing is ever indexed *and* recall is broken because queries embed through the same path" is a genuinely hard bug to find, and you nailed it down to a specific execution provider with a deterministic repro.

**What it does:** mempalace's `embedding_device=auto` picks CoreML on Apple Silicon; CoreML runs the quantized ONNX graph partially and its fp16 partitions overflow, so every embedding comes back NaN. The fix pins `MEMPALACE_EMBEDDING_DEVICE=cpu` on darwin, in both the mine loop's child env and the env merged into each agent's spawn.

**Verdict:** no malicious code, no NSFW, no prompt injection, no dependency changes. Typecheck green, 149/149 tests pass, and I verified the mechanism against the installed mempalace source rather than taking it on trust — `auto` really does resolve **CUDA ▸ CoreML ▸ DirectML** before CPU, so on a Mac it lands on CoreML exactly as you say.

**Two things to fix, both small:** the pin silently overrides a user's own device choice, and it's scoped to *all of macOS* when the failure you documented is specific to one of the two embedding models this app offers.

---

## What I did

Checked out the branch, ran typecheck and the full suite, then read the installed `mempalace 3.3.5` source on this machine to confirm the env var, its precedence, and the `auto` resolution order.

## Verified against mempalace's own source

| Claim | Verified how | Result |
|---|---|---|
| `MEMPALACE_EMBEDDING_DEVICE` is a real, supported var | `mempalace/config.py:355` reads it | ✅ |
| `auto` prefers CoreML on macOS | `embedding.py` header: *"`auto` — prefer CUDA ▸ CoreML ▸ DirectML, fall back to CPU"* | ✅ your root cause holds |
| `cpu` is a valid value | `_PROVIDER_MAP['cpu'] = ['CPUExecutionProvider']` | ✅ |
| Switching device doesn't invalidate an existing palace | *"The same model and 384-dim vectors … so switching device does not invalidate existing palaces"* | ✅ **no re-index needed** — worth adding to the PR body, it's the question a maintainer will ask first |

## Fix 1 — the pin overrides the user, silently

`config.py` resolves the device as **env → `~/.mempalace/config.json` → `"auto"`**. Env wins. `childEnv()` spreads `...process.env` first and then sets the device, so:

```
  user sets  MEMPALACE_EMBEDDING_DEVICE=coreml   in their shell   → overridden
  user sets  "embedding_device": "cuda"          in mempalace's config.json → overridden
```

On macOS there's no escape hatch at all after this change. Two-line fix that keeps one:

```ts
...(MEMPALACE_DEVICE && !process.env.MEMPALACE_EMBEDDING_DEVICE
  ? { MEMPALACE_EMBEDDING_DEVICE: MEMPALACE_DEVICE }
  : {})
```

Anyone who has explicitly chosen a device keeps it; everyone else gets the safe default. That also gives whoever debugs this next a one-command way to reproduce the NaN behaviour.

## Fix 2 — the scope looks wider than the bug

Your root cause is specific: *"the **quantized embeddinggemma** ONNX graph … 330/1647 nodes … fp16 partitions overflow."* But this app ships **two** models, and `embeddinggemma` is not the default:

```
src/main/config.ts:424   embeddingModel: 'minilm'          ← the default
src/main/memory.ts:123   model() → 'embeddinggemma' | 'minilm'
```

`minilm` is `all-MiniLM-L6-v2` — the stock ChromaDB model, not the quantized graph you diagnosed. As written, the pin costs CoreML acceleration for every macOS user on the **default** model, including the ones who never hit the bug. If minilm is also NaN under CoreML then the broad pin is right and the PR body should say so; if it isn't, the condition wants to be `model() === 'embeddinggemma'` (or the pin should apply per-model).

I'd genuinely rather be wrong here than have you narrow a fix that needs to be broad — but as the diff and the description stand, they don't quite agree, and that's the kind of mismatch that gets "simplified" incorrectly later.

## What I tested

| Check | Result |
|---|---|
| Bug still present on `main` | ✅ confirmed — no `MEMPALACE_EMBEDDING_DEVICE` anywhere in `main`'s `memory.ts` |
| Mergeability | ✅ **MERGEABLE**, no conflicts (branch is 40 commits behind but the file hasn't moved) |
| Malware / obfuscation / network / new deps | ✅ none — one const and two spreads |

## On the missing test

`MEMPALACE_DEVICE` is a module-level `const` computed from `process.platform` at import time, which makes it untestable without process-level hacks. The repo already solves this exact problem elsewhere: `buildMissingCliScript(bin, provider, npmAvailable, platform)` takes `platform` as a parameter *"only so the Windows branch is reachable from a test on macOS."* Same trick here —

```ts
export function mempalaceDevice(platform: string, envOverride?: string, model?: EmbeddingModel): string | undefined
```

— and a handful of assertions pin darwin→cpu, linux→undefined, and env-override-wins. That also gives Fix 1 and Fix 2 above somewhere to live.

## A note on the upstream bug

You're right that mempalace should guard NaN output rather than silently writing garbage vectors, and that fixing it here is the correct call for this app's users. If you haven't already, it's worth filing upstream: a "requesting an unavailable accelerator falls back to CPU" policy already exists in `embedding.py`, so "the accelerator loaded but produced NaN" is a natural sibling case for the same fallback.

## Questions

@gpechieu — the important one: **does `minilm` also embed to NaN under CoreML on Apple Silicon**, or was your repro on `embeddinggemma` only? That single answer decides whether the pin should be macOS-wide or model-scoped. Also, were you running with `embeddingModel: 'embeddinggemma'` when you saw 0/19 files mine? And any objection to keeping an env escape hatch?

@chaitanyagiri — if the answer above is "embeddinggemma only", this becomes a settings-level concern rather than a platform one, and it may be worth surfacing in the Settings copy (something like "embeddinggemma runs on CPU on macOS") so the trade-off is visible rather than silent. Your call on whether that's worth the UI.

gpechieu:
Both fixes are in, and your important question has a better answer than either of the options — verified against source, not memory.

**Does minilm also NaN under CoreML? It can't get there.** chromadb's `ONNXMiniLM_L6_V2.model` **unconditionally removes** `CoreMLExecutionProvider` from the preferred providers before building the inference session — their comment: *"it is not as well optimized as CPU"* (`chromadb/utils/embedding_functions/onnx_mini_lm_l6_v2.py`, verified in the chromadb bundled inside mempalace's own env, and mempalace's capped-threads session rebuild replicates the same pruning). So under `auto` on macOS, minilm has been silently running on CPU all along; `embeddinggemma` (mempalace's own ONNX class, no such pruning) is the **only** path that ever reaches CoreML — and that path is the NaN bug. Which dissolves the scope mismatch you flagged: the macOS-wide pin costs no user any acceleration, because the only model the pin can affect is the broken one. The module comment now records this so the pin doesn't get "narrowed" later on the assumption minilm lost something.

**Your other two questions:** yes — the 0/19 mine failures were with `embeddingModel: 'embeddinggemma'` (the palace's `mempalace_embedder.json` recorded it, and the whole palace had been built with it). And no objection to the escape hatch — it's in: an explicitly exported `MEMPALACE_EMBEDDING_DEVICE` wins and the pin emits nothing, which also keeps `=coreml` as the one-command repro you wanted. One honest caveat: a device chosen in `~/.mempalace/config.json` (rather than env) is still overridden, since mempalace's own precedence is env → config.json and honoring it would mean parsing their config from here; the env var is the documented escape hatch.

**Branch updated:** rebased onto current `main`; resolution extracted to an exported pure `mempalaceDevice(platform, envOverride)` per your `buildMissingCliScript` suggestion, pinned by `test/mempalace-device.test.cjs` (darwin→cpu, linux/win32→default, override-wins, empty-override-is-not-a-choice). `npm run typecheck` clean, **262/262** tests. PR body updated with the no-re-index fact you pointed at. The mempalace upstream report ("accelerator loaded but produced NaN → same fallback as accelerator-unavailable") is on my list and your sibling-case framing is exactly how I'll file it.

PR #159, opened 2026-08-17
fix(workers): reliable spawn, teardown and floor cards for god-hired workers

gpechieu:
Four fixes to the `spawn-requests/` ephemeral-worker lifecycle, found running the god→worker flow end-to-end over several days (god hiring real employees). They touch the same few functions so they're bundled — happy to split if you prefer.

## 1. Tokenize the request's `command` (workers carrying flags died in <1s)
god naturally authors `command` as a full command line (`claude --model … --permission-mode …`), but `processSpawnRequest` passed it whole as `AgentSpawnOptions.command`, which the PTY layer treats as ONE executable name → node-pty tried to exec a binary literally named like the entire string → ENOENT → the worker died ~700-900 ms after spawn while its request archived as `.done` (silently "successful"). The hive log correlated 100%: every flag-carrying spawn died, every bare-`claude` spawn lived; the same command line runs fine in a shell. The line is now split (same quoting rules as the renderer's `tokenizeCommand`) into executable + argv, deduping `--model` against the request's separate `model` field.

## 2. Inherit the app's auto (skip-permissions) mode
A headless worker has no human to click through tool prompts — without a permission flag it stalls at the first ask until the idle reaper kills it. When the app's auto-mode is on and the request takes no stance of its own, claude commands now get `--permission-mode bypassPermissions` appended. An explicit `--permission-mode` in the request always wins, and non-claude commands are untouched. Mirrors the renderer's `buildSpawnCommand` semantics.

## 3. Explicit `teardownPty` after kill in release/reap (frozen cards, mail to the dead)
`ptyManager.kill()` deletes the session synchronously, so node-pty's later async exit callback fails the session-identity guard and the global exit handler (→ `teardownPty`) never runs for an explicitly killed PTY. Every kill site called `teardownPty` itself — except the worker done-release / token-cap reap / idle reap paths, which relied on onExit (the old comment "`releasing` guards the gap before onExit fires" documents the false belief). Result seen live: released workers were never archived (registry `archived:false`, no hive `archive` event), their floor cards froze, and god kept mailing dead agents' inboxes. The three paths now call `teardownPty` right after the kill (a double teardown is a harmless no-op), and the stale doc-comment claiming "kill() also makes node-pty fire onExit" is corrected.

## 4. Floor cards for god-hired workers
A MAIN-initiated spawn never got a `hive:agentSpawned` broadcast, so god-hired workers were invisible on the floor, never entered the roster, and after an app restart nothing offered to restore them — and failures like the ENOENT above were completely silent. `processSpawnRequest` now broadcasts the card (same shape as the voice-spawn path) and `teardownPty` broadcasts the matching `hive:agentArchived`. This deliberately softens "workers are headless-by-design": they keep their ephemeral lifecycle (done-release/reap), but the human can see them working, and a worker still alive at app quit survives as a restorable card. If you'd rather keep them headless, items 1-3 stand on their own.

## Validation
- Runtime, each verified with real spawns: a flag-carrying request that previously died in <1s lived and signaled done 22 s later; workers get a floor card instantly and the card archives on release; post-fix releases write the hive `archive` event and free the worker id for rehires.


chaitanyagiri:
## TL;DR

Thanks @gpechieu — this is the highest-value PR of the four you have open. All four fixes are real, and #1 in particular explains a failure that was *silently* succeeding, which is the worst kind.

**What it does:** four fixes to the `spawn-requests/` ephemeral-worker lifecycle, found by actually running god→worker for days. (1) **Tokenize the request's `command`** — god writes a full command line, but it was passed whole as the executable name, so node-pty tried to exec a binary literally named `claude --model … --permission-mode …` → ENOENT → the worker died in under a second **while its request archived as `.done`**. (2) Inherit the app's auto-mode so a headless worker doesn't stall at the first tool prompt. (3) Call `teardownPty` explicitly after `kill()` in the release/reap paths — `kill()` deletes the session synchronously, so node-pty's later onExit fails the identity guard and the global handler never runs. (4) Broadcast `hive:agentSpawned` so god-hired workers get a floor card.

**Verdict:** no malicious code, no NSFW, no prompt injection, no dependency changes. Typecheck green, 149/149 tests pass, and I verified all four bugs are still present on `main`. **Three things before merge:** a fourth kill site you missed, a duplicated tokenizer, and zero tests for any of it. Plus one decision that's the maintainer's, not yours.

---

## What I did

Checked out `fix/god-worker-spawn-lifecycle`, ran typecheck and the full suite, then traced each of the four claims through `main`'s current `index.ts` — including hunting for other call sites with the same shape as the ones you fixed.

## Fix 3 is right, and there's a fourth site

Your root cause is exactly right, and I want to restate it because it's subtle:

```
   ptyManager.kill(id)
        │
        ├─ deletes the session from the map  ── SYNCHRONOUSLY
        │
        └─ node-pty fires onExit later (async)
                 │
                 ▼
           global exit handler checks session identity
                 │
                 ▼
           session is gone → guard fails → teardownPty NEVER runs
```

You fixed the three loop paths in `ephemeralWorkerTick` (done-release, token-cap reap, idle reap). **`ipcMain.handle('workers:stop')` has the same bug and isn't in the diff.** On `main` it reads:

```ts
/** Manually stop a live ephemeral worker. Mirrors the done-release path: mark
 *  releasing, then kill → teardownPty runs the SAFETY-GATED worktree teardown … */
  try { ptyManager.kill(workerId); } catch (e) { … }
  return { ok: true };
```

Its own doc-comment asserts the belief your PR proves false — "kill → teardownPty runs" — so the manual **stop worker** button leaves exactly the state you described: no hive archive, no `hive:agentArchived`, a frozen floor card, and god still mailing a dead inbox. One added `teardownPty(workerId);` and a corrected comment, in the same PR that establishes the rule.

## Fix 1: the tokenizer is a byte-for-byte copy

The inline tokenizer is character-identical to the existing one:

```ts
// src/renderer/src/store/config.ts:302   (existing)
const re = /"([^"]*)"|'([^']*)'|(\S+)/g;
while ((m = re.exec(command)) !== null) out.push(m[1] ?? m[2] ?? m[3]);

// this PR, inline in processSpawnRequest
const tokenRe = /"([^"]*)"|'([^']*)'|(\S+)/g;
while ((tok = tokenRe.exec(command)) !== null) tokens.push(tok[1] ?? tok[2] ?? tok[3]);
```

Your comment says "same quoting rules as the renderer's `tokenizeCommand`" — it's not just the same rules, it's the same code. It can't be imported today because `tokenizeCommand` lives in a renderer module and this is main. The fix is to move `tokenizeCommand` into `src/shared/` (which both sides already import from) and call it in both places. Otherwise the day someone teaches the renderer about backslash escapes, workers keep the old behaviour and nobody notices — which is precisely the class of bug this PR exists to fix.

## What I tested

| Check | Result |
|---|---|
| All four bugs still present on `main` | ✅ confirmed individually |
| `spawnAgentCore` really returns `worktreePath` | ✅ yes — `Promise<{ ok, error?, cwd?, worktreePath?, … }>`, so fix 4's `res.worktreePath` is sound, not silently `undefined` |
| Double-teardown safety | ✅ `wasWorker` is read before `liveWorkers.delete(id)`, so a second `teardownPty` sees `false` and can't emit a duplicate `hive:agentArchived` |
| Iteration safety | ✅ the reap loop iterates `[...liveWorkers]`, so deleting inside `teardownPty` mid-loop is fine |
| Mergeability | ✅ **MERGEABLE**, no conflicts |
| Test coverage for any of the four fixes | ❌ **none added** |

## The missing tests

This is my main ask. Fix 1 is a **pure function problem** — a command line in, an executable plus argv out — and it silently killed real workers for days while reporting success. That is exactly what a unit test is for, and once `tokenizeCommand` moves to `src/shared/` the test is four assertions: a bare command, a flag-carrying command, a quoted model value with spaces, and the `--model` dedupe against `raw.model`. Fix 2's "explicit `--permission-mode` in the request always wins" is another two.

Fixes 3 and 4 are harder to test without an Electron harness, and I wouldn't block on those. But shipping fix 1 untested means the next refactor of `processSpawnRequest` can silently reintroduce a failure mode that presents as "the worker finished successfully".

## Fix 2 deserves an explicit decision (not a change request)

When the app's auto-mode is on, a god-hired worker now gets `--permission-mode bypassPermissions` appended automatically. The reasoning is sound — a headless worker has no human to approve tools, so without it the worker stalls until reaped — and it's gated on a setting the operator turned on themselves, mirrors `buildSpawnCommand`, and lets an explicit request flag win.

It's still worth naming plainly, because a spawn request can originate from **Slack**: with auto-mode on, a Slack-triggered worker runs with tool approval bypassed on the operator's machine. That's a reasonable product position for this app, and I'm not arguing against it — but it should be a decision someone made, not a side effect of a lifecycle bugfix, and the CHANGELOG should say it.

## On bundling

You offered to split, and I'd take you up on it partially: **1–3 are bug fixes; 4 is a behaviour change.** By your own description it "deliberately softens 'workers are headless-by-design'" — workers become visible, enter the roster, and survive a restart as restorable cards. That's a product decision that deserves to be reviewable on its own, and you note 1–3 stand without it. Splitting 4 out lets the fixes land now.

## Questions

@gpechieu — (1) will you add `teardownPty` to `workers:stop` and fix its stale doc-comment? (2) Any objection to moving `tokenizeCommand` to `src/shared/` and importing it in both places? (3) With fix 4, a worker that's alive at app quit now survives as a restorable card — what happens if the human restores it? Does it come back as a normal agent that no longer reaps, or does it re-enter `liveWorkers`? That's the case I'd most want a sentence about.

@chaitanyagiri — two for you: **(a)** fix 4 changes what a worker *is* from the human's point of view (invisible and ephemeral → carded, rostered, restorable). Do you want that, or do you want workers to stay headless with only the failures surfaced? **(b)** Sign-off on fix 2's auto-mode inheritance, given the Slack-triggered path above.

gpechieu:
All three asks are in, plus the split. Branch rebased onto current `main` and force-pushed. Answers in order:

**(1) `workers:stop` — yes, and thank you for the catch.** `teardownPty(workerId)` now runs right after the kill, and the doc-comment that asserted the false belief ("kill → teardownPty runs") now states the actual rule: kill() deletes the session synchronously, onExit fails the identity guard, so **every kill site calls teardownPty itself**. That rule also lives in `teardownPty`'s own header now, so the next kill site someone adds has the invariant in front of them.

**(2) `tokenizeCommand` moved to `src/shared/commandLine.ts`** and both sides import it — the renderer via a re-export from `store/config.ts` (so its five existing importers keep their path), main via the new pure `buildWorkerLaunch()` in `src/main/workerLaunch.ts`, which owns the whole request-line → (bin, argv) translation: tokenization, auto-mode inheritance, and the `--model` dedupe. **Tests added** — `test/worker-launch.test.cjs`: your four tokenizer assertions (bare command, flag-carrying, quoted model with spaces, `--model` dedupe against `raw.model`), your two permission-mode assertions (auto-mode ON appends; explicit request flag wins), plus non-claude untouched, default fallback, and one case pinning that main and renderer route through the *same* shared tokenizer. Typecheck clean, **267/267** tests.

**(3) The restore question — answered in code, and it's the good direction.** `liveWorkers.set` has exactly one call site (`processSpawnRequest`), and a restored card revives through the renderer's normal spawn path — so a restored worker never re-enters `liveWorkers`: no done-scan, no reaping, no token cap. It comes back as a regular agent. The way I'd phrase the invariant: **ephemerality is a property of the hiring, not of the card.** That sentence is now a comment at the broadcast site — which lives in the split-out PR:

**On bundling: done — fix 4 is now #178**, stacked on this branch (its card-archive path rides these explicit `teardownPty` calls, so it can't be independent of this PR). This one is now exactly the three bug fixes, and the "headless-by-design" comment stays intact here so the design question lands on @chaitanyagiri's desk in #178 rather than as a side effect. Fix 2's Slack-origin implication is called out there as well — agreed it should be a decision someone made, and the CHANGELOG note can ride whichever of the two PRs you'd rather carry it.

gpechieu:
Heads-up: this branch moved `51332ab` → `5d38c86` by **fast-forward** — one incremental commit fixing the auto-mode provider blindness caught in the [#178 review](https://github.com/chaitanyagiri/munder-difflin/pull/178#issuecomment-5345398887). The defect lives in `buildWorkerLaunch`, which THIS PR introduces, so the fix had to land here for #159 to be mergeable on its own. Your review of `51332ab` is untouched; the delta is the one commit: the provider's own auto flag via `autoModeFlagForProvider` (token-level stance check — copilot's flag starts with `-s`, a substring of `--summarize`), plus `commandForAutoMode` dead code deleted. +5 tests → **272/272**, typecheck clean.

Issue #140, opened 2026-08-14
Won't create HarnessAgents folder on its own


nrrb:
### What happened?

On initial setup, it asked for a location for the Harness Agents, and suggested `~/HarnessAgents` which I went with. Later, during the "How much can agents do on their own?" setup step, after pressing the "finish" button, it gave the error `ENOENT: no such file or directory, mkdir '~/HarnessAgents'` and pushed the interface off the bottom of the screen so I can't progress. 

### Steps to reproduce

Go through initial setup, set Harness Agents to `~/HarnessAgents`, progress in setup to "How much can agents do on their own?", check all the boxes, press the "finish" button.

### Logs / screenshots

<img width="1552" height="982" alt="Image" src="https://github.com/user-attachments/assets/075db66b-d23a-47b2-88c1-a2dd911b4b5f" />

### macOS version

26.3.1

### Node version

22.14.0

### Claude Code version (if relevant)

2.1.232

### Pre-flight

- [x] I re-ran `npm install` so `node-pty` is rebuilt for the current Electron ABI.
- [x] I searched existing issues and this isn't a duplicate.

chaitanyagiri:
Fixed in **[v0.4.4](https://github.com/chaitanyagiri/munder-difflin/releases/tag/v0.4.4)**.

There were actually two bugs stacked here, which is why it looked like the wizard simply refused to work:

1. **The literal tilde.** Accepting the suggested `~/HarnessAgents` persisted the `~` as a character, and Node's `mkdir` has no concept of the shell's `~` — so Finish died with `ENOENT: no such file or directory, mkdir '~/HarnessAgents'`. It is now expanded at the config-write boundary *and* at the `mkdir` itself.

2. **The suggestion never appeared.** The folder field read `window.process.env.HOME`, which is always `undefined` in that window (contextIsolation is on and only one object is bridged) — so the box was empty while the copy above it promised a default. It now suggests `~/HarnessAgents` for real.

Also in this release: an empty folder is caught at step one instead of after step four, and the wizard no longer overflows a short screen.

Thanks for the report — it was the first thing a new user hit.

PR #160, opened 2026-08-17
fix(config): expand ~ in the harness-home path before mkdir and persist

gpechieu:
**Rewritten after review** — reduced to what `main` still lacks. The original three seams: `ensureHarnessHome` and `writeConfig` are already fixed on `main` (via `normalizeHiveHome`), so they're gone from this diff. What remains is the upgrade path the reviewer confirmed is live:

## The seam that's still open

A config.json persisted by a pre-#140 build holds literal `~/…` strings that nothing rewrites. The hive picker renders them and feeds them straight back into `config:changeHome`, whose `resolve()` anchors `~` at cwd — the app relaunches against a real directory named `~`.

- **`config:changeHome` expands before `resolve()`** — the one consumer still missing the defence-in-depth rule `expandTilde`'s own doc argues for.
- **`readConfig` normalizes `harnessHome`/`recentHives` on the way OUT** (the reviewer's better-shape suggestion): fixes the picker's *display* and the *click* in one place, regardless of the file's vintage, and collapses a stale `~/X` sitting next to its absolute twin into one entry.

## Test

`test/harness-home-tilde.test.cjs` seeds `require.cache` with an electron stub (userData pointed at a temp dir, same pattern as `hive-roster-injection.test.cjs`) so it no longer depends on the electron binary being present. It covers the upgrade path end-to-end (pre-fix config on disk → every consumer-visible path expanded, tilde/absolute twins deduped, "no consumer ever sees a `~` path"), plus the `ensureHarnessHome` expansion as a regression guard and the `ok:false` failure path. Wired into `main`'s **current** `test:focused` list — the stale `package.json` hunk is gone with the rewrite.

## Validation

`npm run typecheck` clean, **261/261** tests (main's 258 + these 3).


chaitanyagiri:
## TL;DR

Thanks @gpechieu — good diagnosis, and you covered one seam that nobody else did.

**Two things to know up front.** First, **#140 is already fixed on `main`**: `writeConfig` routes `harnessHome` through a `normalizeHiveHome()` helper and `ensureHarnessHome` already calls `mkdirSync(expandTilde(path))`, both with comments naming #140. Second, **@rajpreetcodes opened #165 the same week with the same fix**. Your branch is **40 commits behind** `main` and conflicts.

**But you fixed a third seam the others didn't, and `main` still doesn't have it:** `config:changeHome` in `src/main/index.ts` still does `resolve(p.newHome)` with no `expandTilde`. I chased whether that's reachable, and it is — not from the folder picker, but from the **hive picker's recent-hives list on an install that upgraded from a pre-#140 build**. Details below, because that's the one part of this PR I'd merge today.

**Verdict:** no malicious code, no NSFW, no prompt injection, no dependency changes. Typecheck green, 152/152 tests pass. Not mergeable as-is (conflicts + a stale `package.json` hunk you must not take). Worth reducing to the one seam that's still open.

---

## What I did

Checked out `fix/harness-home-tilde`, ran typecheck and the full suite, diffed every hunk against current `main` to find what survives, then traced every caller of `changeHome` to decide whether your third fix is defence-in-depth or a live bug.

## The seam that's still open

`main`, `src/main/index.ts`:

```ts
ipcMain.handle('config:changeHome', async (_evt, payload) => {
  const newHome = resolve(p.newHome);        // ← no expandTilde
```

Both UI callers today feed it a **folder-dialog** result, which is always absolute — so at first glance your fix is belt-and-braces. It isn't, because of the recent-hives list:

```
  pre-#140 build:  onboarding persisted the literal string
                   harnessHome = "~/HarnessAgents"
                   recentHives = ["~/HarnessAgents"]     ← written to config.json on disk
                                        │
                   ── user updates the app ──────────────┤  config.json is NOT rewritten
                                                         │
  today:           HivePicker renders recents from config
                   user clicks "~/HarnessAgents"
                        │
                        ▼
                   changeHome("~/HarnessAgents", 'fresh')
                        │
                        ▼
                   resolve("~/HarnessAgents")
                        │
                        ▼
                   <process cwd>/~/HarnessAgents      💥 a real directory named "~"
                   …and the app relaunches against it
```

`normalizeHiveHome` only cleans values on the way **in**. A config written before the fix keeps its literal `~` until something rewrites it, and the hive picker reads it straight back out. So `expandTilde` at the `changeHome` boundary is genuinely load-bearing for upgrading users — which is exactly the "the consumer refuses to trust that the ingestion point normalized" argument `main`'s own `expandTilde` comment makes. Nice catch, and it's the only place it's still missing.

Two follow-ups that belong with it, if you want to close this properly: `recentHives` read back from disk is never re-normalized on load, and `HivePicker` shows the raw string, so an upgraded user sees `~/HarnessAgents` in the list. Sanitising on read would fix the display and the click in one place.

## What's superseded

| Your change | Status on `main` |
|---|---|
| `ensureHarnessHome` expands `~` before `mkdirSync` | ✅ already there |
| `writeConfig` expands `harnessHome` + `recentHives` | ✅ already there, via `normalizeHiveHome()` |
| `config:changeHome` expands before `resolve()` | ❌ **still missing — this is the one** |

## What I tested

| Check | Result |
|---|---|
| `node --test test/*.test.cjs` | ✅ **152/152** — note `main` is at 258, which is the 40-commit gap showing |
| Your `test/harness-home-tilde.test.cjs` | ✅ 3/3 — and I like that it asserts *no literal `~` directory appeared in cwd*, and that it covers the `ok:false` failure path, which #165's version doesn't |
| Mergeability | ❌ **CONFLICTING** — `package.json` and `src/main/config.ts` |

## Two things to fix

**1. Do not take the `package.json` hunk.** Your branch's `test:focused` predates about ten test files. The diff rewrites that line from an old baseline, so merging it would silently drop `terminal-selection`, `hive-task-mutation`, `hive-windows-prompt`, `skills`, `skills-install`, `release-drop`, `hero-payload`, `ide-image`, `win-cmd-shim` and `transcript-project-dir` from the focused run. It'll conflict rather than apply silently, but please resolve it by re-adding only `test/harness-home-tilde.test.cjs` to `main`'s current list.

**2. Your test needs the electron stub.** `loadTs('src/main/config.ts')` pulls in `import { app } from 'electron'`. It passes on a normal install, and it passed for me — but only once electron's binary was actually present; on a tree installed with `--ignore-scripts` the whole file dies at `Electron failed to install correctly`. `#165`'s version and the existing `test/hive-roster-injection.test.cjs` both seed `require.cache` with a fake `{ app: { getPath } }`, which makes the test independent of whether the electron binary downloaded. Two minutes, and it stops the test being environment-dependent. **This is a robustness note, not a failure — your test passes.**

Minor: the new test file isn't in `test:focused` on `main`'s list either way, so wire it in during the rebase.

## Questions

@gpechieu — would you rebase and cut this down to just the `config:changeHome` line plus the test? That's a clean, uncontested, obviously-correct PR, and it stops this and #165 conflicting over the same `config.ts` lines. Also: did you consider normalizing `recentHives` on **read** rather than only on write? That would fix the display *and* the click for upgraded users, and it might be the better shape of this fix.

@chaitanyagiri — same process question I put on #165: #140 attracted two independent fixes *after* it was already fixed on `main`. If the issue is just un-closed, closing it with a "fixed in main" note saves the next contributor an evening. And my read is that #165 and #160 are **complementary, not competing** — #165 has the stronger regression test plus a real Windows test fix, #160 has this `changeHome` seam. One small combined PR beats picking a winner. Want me to coordinate that between the two authors?

gpechieu:
Done — and your `recentHives`-on-read question turned out to be the better half of the fix, so I took it.

**The branch is rewritten from current `main`** (not rebased-with-conflicts): the two seams `main` already fixed are gone, the stale `package.json` hunk is gone with them (the new test is added to `main`'s current `test:focused` list only), and what remains is exactly the upgrade path you traced: `config:changeHome` now expands before `resolve()`, and — per your suggestion — `readConfig` normalizes `harnessHome`/`recentHives` **on the way out**, which fixes the picker's display and the click in one place for any config vintage, and collapses a stale `~/X` next to its absolute twin. `normalizeHiveHome` keeps owning the write side; the read side now refuses to trust the file, which is the same defence-in-depth argument `main`'s `expandTilde` doc already makes.

**The electron stub is in** — same `require.cache` pattern as `hive-roster-injection.test.cjs`, with userData pointed at a temp dir the test owns, so the suite no longer cares whether electron's binary downloaded. The test covers the upgrade path end-to-end (pre-fix config on disk → no consumer ever sees a `~` path), and keeps the `ensureHarnessHome` expansion as a regression guard plus the `ok:false` path. `npm run typecheck` clean, **261/261** locally.

**On #165:** agreed they're complementary, and with this rewrite the overlap is basically gone — their branch covers the write-side regression tests and the Windows fix, this one covers the read side and the `changeHome` consumer. Happy to coordinate whatever shape @chaitanyagiri prefers (combined PR or land in either order — if #165 lands first I'll rebase over it the same day). And +1 to closing #140 with a "fixed in main" note either way; that's what sent two of us fixing it after the fact.

chaitanyagiri:
Closing this as superseded, by your own commit.

`19646be8`, "expand ~ at the two seams a pre-fix config can still reach", is authored by you and has been in the shipping build since before 0.4.5. It is the broader version of this fix: it covers the seams a config written before the fix can still reach, which this PR does not.

So the behaviour you reported here is fixed and you are credited for it in the history. This branch is just the earlier, narrower attempt, and it has been sitting open behind the version that actually landed.

Sorry for leaving it open this long. Thank you.

PR #161, opened 2026-08-17
fix(floor): stop ANSI cursor sequences garbling the thought bubbles

### LINKED ISSUE #141: garbled text
### What happened?

it doesn't seem to be handling well interim messageds that are flowing..Image

### Steps to reproduce

1. Give Michael a task 
2. Wait for him to share some interim thought bubbles

### Logs / screenshots

```shell

```

### macOS version

0

### Node version

11.9.0

2.1.233

### Pre-flight

gpechieu:
Fixes #141.

## Symptom
Michael's interim thought bubbles render garbled — the reporter's screenshot shows `all three land␛[1Cd.␛[1CSynth␛[1Csizing now,␛[1Cthen␛[1Cwriting…` where the CLI actually printed a clean sentence.

## Root cause
`usePtyParser` scrapes the visible terminal stream to fill the thought bubble / desk card, but its stripper removed **only SGR color codes** (`ESC[…m`). The CLI repaints its live status line using cursor-forward moves (`ESC[nC`) standing in for runs of spaces it knows are already on screen, plus cursor addressing (`ESC[14;6H` — the `…` tail in the screenshot) and erases. All of those leaked into the scraped text verbatim.

## Fix
A small tested `stripAnsi` module (`src/renderer/src/components/ansiText.ts`), used by the parser:
- **Cursor-forward is translated into the spaces it stands for** — simply deleting it would fuse adjacent words (`all␛[1Cthree` → `allthree`).
- Everything else is control, not content, and is removed: OSC strings (titles/hyperlinks), any remaining CSI (SGR, cursor, erase, private modes), charset selects, stray two-byte escapes.
- The scraped summary also collapses space runs, so a multi-column skip doesn't leave gaps in the bubble.

## Validation
- New `test/ansi-text.test.cjs` includes the exact byte sequence reconstructed from the issue's screenshot → asserts it comes out as the clean sentence, plus SGR/CUP/EL/OSC/charset cases and a plain-unicode passthrough.
- `npm run typecheck` clean; `npm run test:focused` 137/137 pass.
- Live app run: floor bubbles clean during real agent turns.


chaitanyagiri:
## TL;DR

Thanks @gpechieu — nice fix, and the test file is the best part of it.

**What it does:** the floor's thought bubbles are scraped from the raw terminal stream by `usePtyParser`, whose stripper removed **only SGR colour codes** (`ESC[…m`). The CLI repaints its live status line with cursor-forward moves (`ESC[1C`) that stand in for runs of spaces it knows are already on screen, plus cursor addressing and erases — and all of that leaked into the bubble verbatim, which is the `all␛[1Cthree␛[1Cland…` in issue #141's screenshot. This adds a small, tested `stripAnsi` that **translates cursor-forward into the spaces it represents** (deleting it would fuse adjacent words) and removes every other escape flavour as control rather than content.

**Verdict:** no malicious code, no NSFW, no prompt injection, no dependency changes. Typecheck green, 156/156 tests pass, and I confirmed `main` still has the SGR-only regex, so the bug is live. **One blocker (a stale `package.json` hunk that will clobber the test script), and one residual garbling source this fix doesn't reach** — worth knowing about even if you leave it.

---

## What I did

Checked out `fix/bubble-ansi-garble`, ran typecheck and the full suite, confirmed the bug is still present on `main`, and read the five regexes against each other to check ordering and interaction — that's where a stripper like this usually goes wrong.

## The mechanism, and why the fix is the right shape

```
   CLI repaints its status line:
       "all three land"  ESC[1C  "d."  ESC[1C  "Synth"  …  ESC[14;6H
                            │                                  │
                            │ stands for a space the CLI        │ cursor addressing
                            │ knows is already on screen        │ (the "…" tail)
                            ▼                                  ▼
   old stripper: removes ESC[…m only  ──────────────────────────►  both leak verbatim
   new stripper: ESC[nC → n spaces, everything else deleted  ────►  clean sentence
```

Deleting cursor-forward instead of translating it would produce `allthree` — so translating is correct, and it's the detail most implementations get wrong. The regex ordering is also right: `CUF_RE` runs before the general `CSI_RE` (which would otherwise eat it), and `ESC2_RE` (`/\x1b./`) runs **last**, so it can't swallow a CSI introducer before the CSI pattern sees it. That ordering is load-bearing and currently undocumented — worth one line of comment, because it looks reorderable and isn't.

## What I tested

| Check | Result |
|---|---|
| `node --test test/*.test.cjs` | ✅ 156/156 |
| `test/ansi-text.test.cjs` alone | ✅ 7/7 |
| Bug still present on `main` | ✅ `usePtyParser.ts:5` is still `const ANSI_RE = /\x1b\[[0-9;]*m/g` |
| Mergeability | ❌ **CONFLICTING** — `package.json` |
| Malware / obfuscation / network / new deps | ✅ none — one new 29-line module, one import swap, one `.replace(/\s+/g, ' ')` |

The test deserves a specific mention: reconstructing the **exact byte sequence from the issue's screenshot** and asserting the clean sentence comes out is the right way to write this. It documents the bug, and it'll survive a rewrite of the implementation.

## Blocker — the `package.json` hunk

Your branch is 40 commits behind, and its `test:focused` list predates about ten test files. The diff rewrites that line from the old baseline, so resolving the conflict carelessly would silently drop `terminal-selection`, `hive-task-mutation`, `hive-windows-prompt`, `skills`, `skills-install`, `release-drop`, `hero-payload`, `ide-image` and `win-cmd-shim` from the focused run. Please rebase and re-add **only** `test/ansi-text.test.cjs` to `main`'s current list. (Same note applies to your #160 — it has the identical stale hunk.)

## The residual garbling this doesn't reach

`stripAnsi` is applied per **chunk**, and PTY data arrives in arbitrary chunk boundaries. An escape sequence split across two reads — `…ESC[1` in one chunk, `C…` in the next — matches nothing, so the fragments land in the bubble as literal text. That's not a regression (the old code had the same hole), but it *is* the same user-visible symptom this PR set out to remove, so it'll look like the fix didn't work when someone hits it.

The usual fix is small: in `usePtyParser`, keep a short carry-over buffer — if a chunk ends in a partial escape (a trailing `ESC`, or `ESC[` plus parameter bytes with no final byte), hold that tail and prepend it to the next chunk. Bounded at a few dozen bytes so a stray `ESC` at end-of-stream can't stall the bubble forever. Happy for that to be a follow-up rather than this PR.

## Smaller notes

- **Word fusion still possible for other movements.** Your own test asserts `stripAnsi('up\x1b[2A\x1b[5Ddown') === 'updown'` — cursor-up and cursor-back are deleted, so they fuse exactly the way cursor-forward would have. That's a defensible choice (there's no sane text rendering of "move back 5"), but it's worth saying so in the module comment, since a future reader will notice the asymmetry and "fix" it.
- **`Math.min(n, 80)`** — the cap is sensible, but 80 is a bare literal in the middle of a `.replace`. A named `MAX_CUF_SPACES` with a one-line "a bubble never needs more than a terminal width of padding" would explain it.
- **File placement.** `ansiText.ts` sits in `components/` but isn't a component, and its only consumer is in `hooks/`. There's no renderer util directory yet, so either start one (`src/renderer/src/util/`) or put it in `src/shared/` — the latter also makes it reachable from main, which matters if anything main-side ever scrapes terminal text.

## Questions

@gpechieu — (1) any objection to the carry-over buffer as a follow-up, or would you rather fold it in here so #141 closes completely? (2) Is `usePtyParser` the only consumer of scraped terminal text, or do the desk cards / hive status captions scrape through a different path that still uses an SGR-only strip? I only found the one, but you know this code far better than I do.

@chaitanyagiri — nothing needs your decision on this one; it's a clean fix behind a rebase. The only thing worth your attention is that **three** of these open PRs (#160, #161, and to a lesser extent #159) each carry a stale `test:focused` line from the same old baseline. If they land in the wrong order the script quietly loses test files. Might be worth asking contributors to rebase before merge, or moving the focused list out of `package.json` into a file that merges more sanely.

chaitanyagiri:
Small correction to the last paragraph of my review above, now that I've checked all twenty open PRs rather than the three I'd looked at when I wrote it.

I said "#160, #161, and to a lesser extent #159" carry a stale `test:focused` line. **#159 does not** — it only touches `src/main/index.ts`. The accurate list is **six** PRs, all rewriting that one line from an older baseline:

`#136` · `#149` · `#154` · `#160` · `#161` · `#162`

(#172 is the only PR that touches `test:focused` from the current baseline, so it's fine.)

That makes the point stronger rather than weaker, @chaitanyagiri: six open PRs each want to rewrite the same single-line script from six different starting points. Whichever lands first, the other five conflict — and a careless conflict resolution silently drops test files from the focused run. Moving the focused list out of `package.json` (a plain file, one test path per line, or just deleting `test:focused` in favour of `node --test test/*.test.cjs`) would make all six merge without a decision.

Nothing else in my review changes.

gpechieu:
Rebased and force-pushed, with your notes folded in. Answers:

**(1) The carry-over buffer went in here, not as a follow-up** — you're right that it's the same user-visible symptom through a different door, and "the fix didn't work" is exactly the bug report I don't want #141 to generate. `splitTrailingPartialEscape` (exported next to `stripAnsi`) holds an unfinished trailing sequence — bare `ESC`, `ESC[` + parameter bytes with no final byte, an OSC with no BEL/ST, a dangling charset select — and `usePtyParser` prepends it to the next chunk before stripping. Bounded at `MAX_ESC_CARRY` (64 bytes): past that we assume a stray ESC or a pathological OSC and let the text through rather than stall the bubble on a terminator that may never come. A completed sequence at end-of-chunk is *not* held (the final byte breaks the anchored match) — there's a test pinning that, plus a split-chunk matrix (CSI mid-parameters, split right after ESC, OSC mid-title, charset).

**(2) `usePtyParser` is the only scrape path — verified by sweeping `\x1b` through both processes.** Every other renderer hit is terminal *I/O*, not scraping: xterm decoration writes and bracketed-paste framing in `terminalPool.ts`, paste framing in `useHive.ts`, string literals in `mockEvents.ts`. Main's single hit (`hiddenClaude.ts`) writes bracketed paste too. The desk cards and status captions render what `usePtyParser` already put in the store, so they inherit the fix. That said, I took your file-placement suggestion in its stronger form — the module now lives in `src/shared/ansiText.ts` — so if anything main-side ever does scrape, the correct stripper is already reachable.

**Your smaller notes, all in:** the regex ordering is documented as load-bearing (both halves: CUF before CSI, ESC2 last); the cursor-up/back fusion asymmetry is documented as deliberate, with the "the fix is a real screen model, not a smarter regex" warning to the future reader who notices it; `80` is now `MAX_CUF_SPACES` with its one-line rationale. And the stale `package.json` hunk is resolved the way you asked: `main`'s current `test:focused` list plus only `test/ansi-text.test.cjs`.

`npm run typecheck` clean, **269/269** tests (main's 258 + these 11).

gpechieu:
Closing this one myself: its content is already on `main`.

- The original fix from this PR was cherry-picked as `a3b75f6` (`stripAnsi` in `src/renderer/src/components/ansiText.ts`).
- The trailing-partial-escape carry that the review asked for landed as your own `44bfb19` (`createAnsiStripper`, bounded carry), and both have shipped since v0.4.5.

Both commits are ancestors of current `main` (checked with `git merge-base --is-ancestor`), so there is nothing left in this branch that the build does not already have. Thanks for the review and for landing it.

PR #162, opened 2026-08-17
fix(hive): main-process inbox-wake watchdog for workers

### LINKED ISSUE #151: Worker agents can permanently stall on a delivered inbox message — worker wake is renderer-only with no main-process retry
## Summary

A worker agent that has gone idle can be handed a message in its inbox and never process it — it sits at its prompt indefinitely (no tool calls, token count unchanged) even though its session is alive and the message file is durably on disk. The orchestrator ("god") agent does **not** exhibit this because it has an independent main-process re-engage loop that workers lack. Today the only reliable recovery for a stuck worker is a full app restart.

## Environment

- App version: v0.4.3
- The Electron desktop app running a multi-agent "hive."
- At least one **worker** agent (i.e., not the orchestrator) that has quiesced to idle at its prompt.
- A message delivered to that worker's inbox (agent-to-agent mail, a human reply, or an orchestrator dispatch).
- Reproduces most easily when the app window is backgrounded/occluded, or when the renderer's per-agent status for that worker is anything other than `idle` at the moment the message lands.

## Steps to reproduce (minimal)

1. Spawn a worker agent and let it finish its current turn so it goes idle (`Stop` fires; it's now sitting at its prompt).
2. Deliver a message into that worker's inbox.
3. Cause the single renderer nudge to be missed at delivery time — e.g. **background the app window** so Chromium throttles/suspends the renderer's `setInterval` timers, or arrange for the worker's renderer store status to not read `idle` on that tick.
4. Observe: the inbox message is never picked up. The worker never runs a turn on it.

## Expected vs actual

- **Expected:** the worker is re-invoked to read and act on delivered inbox mail, the same way the orchestrator agent is.
- **Actual:** the durable inbox file sits unprocessed indefinitely. The worker stays idle — last tool `null`, token count unchanged — until the entire app is restarted, which re-attaches every agent and re-drives delivery. A per-agent message delivery alone does not reliably recover it.

## Root cause

Worker wake depends on a **single, renderer-only path**, and nothing retries it:

- **Renderer is the only worker-wake mechanism.**
  - `src/renderer/src/hooks/useHive.ts:587-624` (effect #3): a 4 s `setInterval` reads each agent's inbox and **enqueues** the wake nudge.
  - `src/renderer/src/hooks/useHive.ts:668-704` (effect #4): actually types the nudge into the agent's terminal, but **only when the renderer store shows that agent as `status === 'idle'`** (gate at `:694`), present with a live pty, delivery not paused, off cooldown, and with no user draft in the terminal.
  - If any of those conditions is off at the tick when the message arrives — the window is backgrounded and the timers are throttled, the store entry is stale/missing, or the status is stuck at a non-`idle` value — the enqueued nudge is never delivered and the durable inbox file is left untouched. There is exactly **one** nudge attempt per new message id, and **no** retry.

- **The main process deliberately does not wake a worker from inbox at `Stop`.**
  - `src/main/hooks.ts:211-221`: on `Stop` / `SubagentStop` the main process intentionally does **not** force a continuation from unread inbox mail (see the comment at `:214-217`). Inbox files remain durable and the renderer is expected to wake the agent "later through its guarded idle-only delivery path." So once a worker is already idle, no main-process event re-invokes it when new mail lands.

- **The orchestrator agent self-heals; workers don't.**
  - `src/main/index.ts:1016` (`reengageGod`) plus the heartbeat digest (`src/main/index.ts:964`) and the actionable-inbox count (`src/main/index.ts:1003`) give the orchestrator an **independent main-process re-engage on a cadence**: its inbox digest is re-delivered periodically, so a missed renderer tick is retried at the next beat. Workers have **no** equivalent. That asymmetry is exactly why the orchestrator reliably wakes on inbox delivery while a worker can be one missed tick away from a permanent stall.

- **Why a full restart recovers everything:** restarting re-spawns/re-attaches every pty, re-adds each worker to the renderer store with a fresh `idle` status, re-runs each boot sequence (which re-drives delivery of the now-durable inbox files), and resets any throttled renderer timers.

### Secondary bug (contributes; worth fixing regardless)

The "already-nudged" dedup keys on the **lexically-largest** inbox id:

```js
// src/renderer/src/hooks/useHive.ts:609
? inbox.map((m) => m.id).sort().slice(-1)[0]
```

When inbox ids mix formats — e.g. timestamp-prefixed ids (`2026-01-02T03-04-...`) alongside human-authored slug ids (`some-task-name`) — the lexical `sort` does not order by recency (ASCII digits sort before lowercase letters). A newly delivered message can therefore sort **below** an older id still in the box, so its id equals the last-nudged id and the new message is treated as already-seen and never nudged.

## Workaround

Until the underlying wake path is fixed (see Suggested fix), a stalled worker can be recovered manually without restarting the whole app:

- **Force a real turn (most reliable).** Open the stuck worker's terminal in the app, type any short message — e.g. `read your inbox now` — and submit it. Forcing an actual turn makes the worker read and act on its pending inbox immediately. Note: a bare Enter may be a no-op; it needs to be a real turn (a few words), not an empty submit.
- **Restart & continue.** If forcing a turn doesn't take, use the per-agent "restart & continue" control, which relaunches the agent while resuming its conversation (no lost context) and re-reads the inbox on boot.
- **Full app restart.** Restarting the app recovers all stalled workers at once, but it's the heaviest option — use it only when several agents are stuck.

These are manual mitigations, not a fix.

## Suggested fix

- **(A) Give workers a main-process inbox-wake watchdog, mirroring `reengageGod`.** On a cadence, for each live worker that is idle, has an undrained inbox, and is not paused / not awaiting human input (no pending permission prompt or "waiting" notification), re-drive the **same guarded delivery** the renderer already uses. This removes the single point of failure and gives workers the resilience the orchestrator already has.
- **(B) Fix the dedup:** track a per-agent **set** of already-nudged message ids (or nudge whenever the inbox is non-empty and not mid-drain) instead of comparing against the lexically-largest id.

**Important constraint:** the naive fix — forcing a continuation from unread inbox at `Stop` — is unsafe and is precisely what `src/main/hooks.ts:211` removed (it can bypass terminal-draft / human-in-the-loop safety and spend credits while a user is mid-answer). Any wake added by (A) must preserve those guards: deliver only when the worker is genuinely idle, delivery isn't paused, and no permission/HITL prompt is pending.

## Notes

File/line references above were verified against the current repository state at the time of writing; a maintainer should re-confirm before landing a patch, as line numbers drift.

gpechieu:
Fixes #151 — the primary fix (A): a main-process wake watchdog for workers. The secondary dedup bug (B) already has its own open PR (#130); this change does not touch effect #3's dedup structure, so the two compose (and this watchdog also covers the stall while a dedup key is wedged, via its retry cadence).

## The asymmetry (as the issue lays out)
Worker wake lived ONLY in the renderer: one 4s interval enqueues the inbox nudge (single-shot dedup on the newest message id), another types it when the store shows the agent idle. One wedged store status, one stale dedup key, or a renderer that simply wasn't there — a reload, or the launch picker still open — and the durable inbox file sat unread until a full app restart. God never had this problem: `reengageGod` re-delivers his digest from MAIN on a cadence. Workers now get the same second engine.

(One nuance found while verifying the issue's premise: the window already runs with `backgroundThrottling: false`, so timer throttling specifically is less of a factor than the issue assumes — but every other listed failure mode is real, and the fix is the one the issue proposes.)

## Design — detection in main, delivery stays guarded in the renderer
- `src/main/inboxWake.ts` (pure decision, tested): fire for a non-god, non-archived agent with a live pty that has been **quiet ≥30s** while inbox mail sits undrained. New mail fires immediately; the SAME undrained mail re-fires every 5 minutes — the retry the renderer's single-shot nudge never had. A drained inbox clears the per-agent state.
- A third always-on beat (60s, next to the fleet/breaker beats) broadcasts each firing as `hive:inboxWake`.
- The renderer handler re-drives the **normal guarded queue + drain path**: it reconciles a wedged `working` status to idle (only `working` — `waiting`/`blocked`/`looping` are deliberate HITL/breaker holds and are never overridden), enqueues the standard nudge only if one isn't already queued, and syncs effect #3's dedup key so the two paths never double-nudge. Delivery still rides every existing typing guard (idle-only, boot grace, pause, draft/picker safety) — per the issue's constraint, this cannot bypass the protections `hooks.ts` deliberately preserves at Stop.

## Validation
- New `test/inbox-wake.test.cjs` (6 tests): fires for a quiet worker with mail; skips god/archived/pty-less/busy; retry-window suppression and re-fire; immediate fire on new mail; state cleared on drain.
- `npm run typecheck` clean; `npm run test:focused` 136/136 pass.
- **Live repro + fix observed**: spawned a real throwaway worker, delivered real hive mail, held the renderer path off. Main logged `[inbox-wake] worker-test151: 2 undrained message(s), pty quiet 78s — re-firing wake` — notably while the renderer was still sitting on the launch picker, i.e. the exact "renderer isn't there" stall the old design could never recover from. With the floor up, a second worker's held mail produced exactly one watchdog fire (`worker-test152: 1 undrained message(s), pty quiet 39s`, measured via hook-idle across an app restart + `--resume`), and once delivery resumed the queued nudge was delivered and the worker drained its inbox to `.done/`. (The delivery layer's own bounded ack-retry typed the nudge a second time 12s later — pre-existing #113 behavior, harmless: that turn found an empty inbox. The watchdog itself never double-fired and never double-queued.)

### Why hook-idle instead of pty quiet — a finding worth keeping
The first draft gated on "pty produced no output for 30s". Live validation falsified it: **an idle claude CLI repaints its prompt continuously**, so pty output recency reads "busy" forever on exactly the agents this fix is for (a fresh TUI that never ran a turn does go quiet — which is why the first live fire still happened). The shipped gate therefore derives idle from hook boundaries (`Stop`/`SessionStart` = at the prompt; any other real event, including a HITL notification, = busy; `Status` telemetry ticks deliberately excluded), with pty quiet only as the fallback for hookless providers. Note this also means `looksStuck()`/similar heuristics reading `ptyManager.idleFor` for claude TUIs may deserve the same treatment — out of scope here.


chaitanyagiri:
## TL;DR

Thanks @gpechieu — this is the best-designed PR in the current queue. The split you chose — **decision in main as a pure tested function, delivery left in the renderer behind every existing guard** — is exactly right, and it's the reason this can't accidentally bypass the HITL protections `hooks.ts` preserves at Stop.

**What it does:** worker wake lived only in the renderer — one 4s interval enqueues an inbox nudge (single-shot dedup on the newest message id), another types it when the store shows idle. One wedged status, one stale dedup key, or a renderer reload, and durable inbox mail sat unread until a full app restart. God never had this problem because the heartbeat re-delivers his digest from main. This gives workers the same second engine: a 60s main-process beat fires `hive:inboxWake` for any non-god, non-archived agent with a live PTY that has been idle ≥30s while inbox mail sits undrained, re-firing every 5 minutes while the same mail stays put.

**Verdict:** no malicious code, no NSFW, no prompt injection, no dependency changes. Typecheck green, 157/157 tests pass, the feature is absent from `main`. **One real gap I'd like closed** — there's a stall class the watchdog structurally cannot see — plus a rebase and two cosmetics.

---

## What I did

Checked out `fix/worker-inbox-wake-watchdog`, ran typecheck and the full suite, tested the merge against `main`, and traced the idle signal from `hooks.ts` through `inboxWakeTick` into the renderer handler to look for cases where the watchdog can't fire.

## The gap: a hung turn is invisible to the watchdog

The idle source is deliberately layered — hook boundaries first, PTY recency as fallback for hookless providers:

```ts
const idleMs = deps.hookIdleFor(id) ?? deps.idleFor(ptyId) ?? 0;
```

`??` only falls through on `null`/`undefined`. And `hookIdleFor` returns:

```
   no hook events ever  → null   → falls back to pty recency   ✅
   turn in flight       → 0      → does NOT fall back           ⚠️
   Stop / SessionStart  → ms idle                               ✅
```

So for a hook-bearing agent (claude, codex, grok…), if the last event was `PreToolUse` and the CLI then **hangs, crashes without a Stop, or has its hook delivery fail**, `hookIdleFor` returns `0` forever. `idleMs < WAKE_QUIET_MS` is permanently true, the watchdog never fires, and the PTY has been silent for an hour.

That's the same shape as the failure this PR exists to fix — a durable inbox file sitting unread with no second engine — just reached through the hook path instead of the store path. And the renderer already has the concept: effect 2e is described as *"the linchpin"* precisely because *"a bridge whose turn-end signal doesn't fire leaves the agent pinned 'working'."* The main-side watchdog deserves the same backstop:

```ts
const hookIdle = deps.hookIdleFor(id);
const ptyIdle = deps.idleFor(ptyId) ?? 0;
// A hook-idle of 0 means "mid-turn" — but a genuinely working turn keeps the
// pty emitting. A long pty silence overrides a stuck hook state.
const idleMs = hookIdle === null ? ptyIdle : Math.max(hookIdle, ptyIdle > HOOK_STALE_MS ? ptyIdle : 0);
```

with `HOOK_STALE_MS` deliberately well above `WAKE_QUIET_MS` (a few minutes) so a legitimately long tool call is never interrupted. Your test file is already structured to take that case — `hookIdleFor: () => 0, idleFor: () => VERY_LONG` — as a seventh test.

## What I tested

| Check | Result |
|---|---|
| `node --test test/*.test.cjs` | ✅ 157/157 |
| `test/inbox-wake.test.cjs` | ✅ 6/6 |
| Feature absent from `main` | ✅ no `inboxWake.ts`, no `activity` map in `hooks.ts` |
| `??` semantics on a busy hook state | ⚠️ correct as written, but see the gap above |
| Renderer can't bypass typing guards | ✅ confirmed — the handler only calls `enqueueMessage`; delivery still rides the idle-only drain |
| Double-nudge safety | ✅ shared `INBOX_NUDGE` const, queue-content dedup, and `nudged.current` synced so effect #3 won't re-queue |
| Status reconciliation is narrow | ✅ only `working` is overridden; `waiting`/`blocked`/`looping` untouched — the right call, those are deliberate HITL/breaker holds |
| Mergeability | ❌ **CONFLICTING** — `package.json` and `src/renderer/src/hooks/useHive.ts` |

The `hooks.ts` idle derivation is the subtle part and you got it right: `Stop` **or** `SessionStart` is idle (without `SessionStart` a restarted agent that never ran a turn would read busy forever), everything else is busy, and status ticks return before reaching the map so pure telemetry can't fake activity. That last detail is easy to get wrong and you called it out in the comment.

## Smaller notes

- **The `activity` map is never pruned.** Agents are archived, respawned with new ids, and killed; the map only ever grows. It's small and `contextById` right above it has the same shape, so this is consistency rather than a new problem — but a `delete` on archive/teardown would be tidy.
- **The log line is slightly misleading.** `` `pty quiet ${idleMs/1000}s` `` prints a number that is usually **hook**-derived, not PTY-derived. When you're reading logs to debug exactly this feature, that's the wrong label. `idle ${…}s (hook)` / `(pty)` would pay for itself.
- **Effective latency is up to ~90s** (30s threshold + 60s beat). Fine for a backstop, worth one sentence in the comment so nobody later "fixes" a 90s delay they think is a bug.
- **The stale `package.json` hunk.** Your branch rewrites `test:focused` from an old baseline — resolve by re-adding only `test/inbox-wake.test.cjs` to `main`'s current list. Five other open PRs have the same stale hunk (see my note on #161), so whoever merges last needs to be careful.

## Questions

@gpechieu — (1) is the hung-turn case above one you considered and rejected, or a gap? I'd rather ask than assume, since you've watched this run live and I haven't. (2) You mention #130 handles the dedup half and that the two compose — is there a merge order that matters, or are they genuinely independent? (3) `WAKE_RETRY_MS` is 5 minutes for the *same* mail — for a worker that's genuinely wedged, is five minutes of repeated nudging the desired behaviour indefinitely, or should it give up and tell god after N attempts, the way the terminal-handoff path bounces?

@chaitanyagiri — no decision needed from you on the design; it's sound and it mirrors `reengageGod`. The only thing worth your attention is that this PR, #157, #159 and #161 are all @gpechieu's, all 40 commits behind, and three of the four conflict. They're independently good fixes and they're going stale as a batch — worth picking a merge order and asking for one rebase pass rather than four.

gpechieu:
Rebased onto current `main` and force-pushed, with the gap closed. Your three questions, in order:

**(1) The hung turn was a gap, not a considered rejection** — and your read of the failure shape is exactly right: it's the same class this PR exists to fix, reached through the hook path instead of the store path. Closed with the shape you sketched: `HOOK_STALE_MS = 5 min`, and a pty silence past it overrides a stuck-busy hook state (`hookIdle === 0` forever after a hang / crash-without-Stop / lost hook delivery). Two tests pin both sides: the override fires (with `idleMs` from the pty and the source labeled), and a legitimately long *quiet* tool call under the stale bound stays protected. One honest note on the trade: a real tool call that emits nothing for 5+ minutes can now draw a wake — bounded and benign, since delivery still rides the renderer's idle-only guards, and the alternative was structural blindness to a worker dead an hour with mail waiting.

**(2) The #130 question answered itself during the rebase: `main` has already absorbed it** — effect #3 now dedups per-id with a Set (nudge on any id not yet nudged) rather than by newest-id. So there's no merge order left to pick; what mattered was *composing* with the landed version: the fire payload now carries `inboxIds` (every undrained id at fire time) and the renderer handler marks them **all** as nudged — syncing only `newestId` against a Set would have under-synced and let effect #3 queue a duplicate for older unseen mail. The queue-content dedup on `INBOX_NUDGE` (now the shared const both paths use) backstops it either way.

**(3) As written it nudges indefinitely, and that's half-right.** The fire itself is nearly free and idle-gated, and the mail stays valid, so there's no runaway — but for a genuinely wedged worker I agree the better ending is "give up and tell god after N attempts", the way the terminal-handoff path bounces. The state entry already has the shape for it (`{ newestId, at }` → add `attempts`). I left it out to keep this PR reviewable; happy to add it here if you'd rather it lands complete, or as the follow-up.

**Also from your smaller notes:** the log now labels its source — `idle 78s (hook)` / `(pty)` — so debugging this feature reads the right signal; the ~90s worst-case latency is documented at `WAKE_QUIET_MS` as the intended backstop shape ("not a bug to fix"); and `hooks.ts` grew a `forget(agentId)` called from `teardownPty`, pruning `activity` — and while I was there, `contextById` and `transcriptPaths`, which had the same leak — so a respawn reusing an id starts from "no hook events yet" instead of inheriting the dead run's turn state. Stale `package.json` hunk resolved against `main`'s current list. `npm run typecheck` clean, **267/267** tests (main's 258 + these 9). And your batch note is now moot in the good way: all four of my conflicting PRs got their one rebase pass today.

chaitanyagiri:
_Posted from @chaitanyagiri by an agent on the Munder Difflin hive floor. First person statements below are the agent's. Replies here are read._

---

Flagging the state of this PR, because nothing on either side records it.

**The change this PR makes appears to already be on `main`.** `src/main/workerWake.ts` is present on `origin/main`, added by commit `68cbc25c` (`fix(worker-wake): main-process watchdog wakes idle workers on undrained inbox`), whose message cites `#151, fix A`. That is the same issue and the same lettered fix this PR describes.

**That commit is authored by someone else and dated 2026-08-19, two days after this PR was opened.** I do not know whether this PR's work informed it or whether the two were written independently, and I am not going to guess in either direction.

**I am flagging this rather than closing it.** This has been open for two weeks looking pending while its subject may already be resolved, and that is not a thing a contributor should find out by accident. @chaitanyagiri for a decision on what happens to it.

gpechieu:
Answering the flag above, and closing this one myself so it stops looking pending.

`src/main/workerWake.ts` (`68cbc25c`, on `main` since v0.4.5) solves the same problem this PR targeted (#151, fix A): a main-process watchdog that wakes an idle worker with undrained inbox mail, with a per-worker cooldown so it composes with the renderer's own nudge rather than duplicating it. I have been running that version locally since the v0.4.5 merge and it covers my case, so keeping a second wake engine in this branch would only mean two nudges for the same mail.

The two were written independently as far as I know; no credit question from my side. Thanks for the review notes on the earlier round, they fed into how I read the landed version.

PR #165, opened 2026-08-17
fix(onboarding): expand ~ in harness home so setup cannot die on ENOENT

### LINKED ISSUE #140: Won't create HarnessAgents folder on its own

rajpreetcodes:
## What

First-run setup could not be completed when the harness home was typed with a tilde, which is exactly what the wizard suggests (`~/HarnessAgents`).

Two distinct failures, both from the issue:

1. **`ENOENT: no such file or directory, mkdir '~/HarnessAgents'`** — Node's fs treats a literal `~` as a directory name (it is a shell convention), so `ensureHarnessHome` failed and, had it succeeded, the un-expanded path would have been persisted and poisoned every path derived from it (hive, memory, roster).
2. **The error pushed the interface off the bottom of the screen** — the wizard panel had no height bound and the error box did not wrap, so the Finish button became unreachable.

## Changes

- `src/main/config.ts` — `ensureHarnessHome` expands `~` through the existing shared `expandTilde` before `mkdirSync`; `writeConfig` persists `harnessHome` expanded/absolute using the same ingestion pattern as `registeredRepos` right above it, and recent-hives tracking stores the expanded value.
- `src/renderer/src/components/OnboardingWizard.tsx` — wizard content scrolls within `86vh` (same pattern as `AddAgentModal`) and the error box gets `overflowWrap: 'anywhere'`.
- `test/harness-home.test.cjs` — new regression test: tilde home is created on disk for real, and `writeConfig` persists it absolute (config + recentHives + on-disk read).
- `test/expand-tilde.test.cjs` — pre-existing Windows-only failure fixed: it asserted POSIX literals like `/a/b/c`, but `resolve()` anchors those to the current drive on Windows; now compares against `path.resolve` output.

## Verification

- `npm run typecheck` (node + web): clean
- New tests: pass; full focused suite shows no new failures (the `codex-remote`/`transcript-project-dir` failures fail identically on a clean checkout on Windows)

Closes #140


chaitanyagiri:
## TL;DR

Thanks @rajpreetcodes — the diagnosis is correct and the write-up is clear. Two things you should know before you spend more time on it, neither of which is your fault:

1. **Issue #140 has already been fixed on `main`.** `writeConfig` now routes `harnessHome` through `normalizeHiveHome()`, and `ensureHarnessHome` already does `mkdirSync(expandTilde(path))` — with comments that name #140 explicitly. Your branch is **40 commits behind** and now conflicts in `src/main/config.ts`, which is exactly the file both fixes touch.
2. **PR #160 by @gpechieu is the same fix**, opened the same week, also against #140.

So the config.ts half of this PR is superseded. **But three pieces of it are not on `main` and are worth keeping** — most importantly your regression test, which is stronger than the one that shipped.

**Verdict:** nothing malicious, no NSFW, no prompt injection, no odd dependencies. Typecheck green, 151/151 tests pass on the branch. Not mergeable as-is (conflicts). My recommendation is to rebase and reduce this to the parts `main` doesn't have.

---

## What I did

Checked out `fix/onboarding-tilde-harness-home`, ran typecheck and the full test suite, then diffed every hunk against current `main` to work out what is still novel — because a 40-commit gap usually means part of a PR has been overtaken.

## What's already on `main`

| Your change | Status on `main` |
|---|---|
| `ensureHarnessHome` expands `~` before `mkdirSync` | ✅ already there — same line, plus a long comment about "defense-in-depth at the consumers" |
| `writeConfig` persists `harnessHome` expanded, incl. `recentHives` | ✅ already there, via a `normalizeHiveHome()` helper |
| Wizard content scrolls instead of pushing Finish off-screen | ✅ effectively there — `OnboardingWizard.tsx:197` now sets `overflowY: 'auto'` on the overlay, with a comment about the panel being cut off at both edges. Your `maxHeight: '86vh'` on the inner div is a different mechanism for the same symptom |

## What is still novel (keep these)

**1. Your regression test is better than the one that shipped.** `main`'s coverage lives inside `test/expand-tilde.test.cjs` and looks like this:

```js
test('#140: ensureHarnessHome creates the expanded path, never a literal ~', () => {
  const { ensureHarnessHome } = (() => {
    return { ensureHarnessHome: (p) => expandTilde(p) };   // ← a stand-in, not the real function
  })();
```

That asserts `expandTilde` works. It would still pass if someone deleted the `expandTilde` call from the *real* `ensureHarnessHome` tomorrow — which is precisely the regression it's named after. Your `test/harness-home.test.cjs` loads the actual `src/main/config.ts`, calls the actual `ensureHarnessHome('~/…')`, and asserts the directory exists **under `os.homedir()`**, then checks `writeConfig` persists an absolute path in both `harnessHome` and `recentHives[0]` and that it survives a re-read from disk. That's the test #140 deserves, and I'd take it on its own merits.

Your electron stub (seeding `require.cache` with a fake `app.getPath`) also matches the pattern already used in `test/hive-roster-injection.test.cjs`, so it's house style, and pointing `userData` at a temp dir means the test can't scribble on a real config. Good hygiene.

**2. The Windows fix in `test/expand-tilde.test.cjs` is real and still missing.** `main` still asserts `expandTilde('/a/b/c') === '/a/b/c'`, and `path.resolve` anchors a POSIX-looking path to the current drive on Windows (`L:\a\b\c`), so that test fails there. Comparing against `path.resolve(...)` output is the right fix. Given v0.4.4's headline was *fixing Windows*, a Windows-red test in the suite is worth closing.

**3. `overflowWrap: 'anywhere'` on the error box** is not on `main`. It's one line and it stops a long ENOENT path from blowing out the panel width.

## What I tested

| Check | Result |
|---|---|
| `node --test test/*.test.cjs` (this branch) | ✅ 151/151 — but note that's the **older** suite; `main` is at 258 tests, which is the clearest sign of the 40-commit gap |
| Your two new tests specifically | ✅ both pass, and they genuinely touch the disk |
| Mergeability | ❌ **CONFLICTING** — `src/main/config.ts` only. One file, and the conflict is "your fix vs the fix that landed", so it resolves by deletion |
| Malware / obfuscation / network calls / new deps | ✅ none — three source files and two test files, no dependency changes |

## Suggested path forward

Rebase onto `main` and drop the superseded hunks, leaving roughly:

- `test/harness-home.test.cjs` (as-is)
- the `path.resolve` fix in `test/expand-tilde.test.cjs`, and I'd delete `main`'s stand-in `#140:` test in that file at the same time, since your real one replaces it
- `overflowWrap: 'anywhere'` on the error box

That turns a conflicting PR into a small, clean, obviously-good one, and it stops #165 and #160 fighting over the same lines.

One more thing worth doing while you're in there: neither your test file nor #160's is added to the `test:focused` script in `package.json`, so they only run under `node --test test/*.test.cjs`. Whichever test survives should be wired into `test:focused` too.

## Questions

@rajpreetcodes — happy to rebase and cut it down to those three pieces? If you'd rather not, say so and I'll suggest the test be lifted into a separate PR so the work isn't lost — it's the strongest part of this and I don't want it dropped just because the config fix got there first.

@chaitanyagiri — a process question rather than a code one. #140 drew **two** independent fixes (#165 and #160) *after* the fix had already landed on `main`, which means two contributors each spent an evening on a solved bug. Was #140 left open on purpose, or did it just not get closed when the fix merged? If it's the latter, closing it — and adding a "fixed in `main`, don't take this" note — would save the next person. And for these two PRs specifically: my read is that the residual value is complementary (this one has the better test and the Windows fix, #160 has the `config:changeHome` seam), so the cleanest outcome is one small combined PR rather than picking a winner. Want me to say that on #160 as well?

rajpreetcodes:
Rebased and cut down as suggested — the branch is now exactly the three novel pieces on top of current `main`:

- `test/harness-home.test.cjs` (as-is, header comment updated to note it replaces the stand-in)
- the `path.resolve` fix in `test/expand-tilde.test.cjs`, with the stand-in `#140:` test deleted. The same POSIX-literal problem turned out to be in three of the `#140` recentHives tests that landed on `main` (`/other/hive`, `/now/here`, `/hive/new` all anchor to the current drive on Windows), so those are fixed too — six spots total
- `overflowWrap: 'anywhere'` on the wizard error box
- `harness-home.test.cjs` wired into `test:focused`, since you flagged that neither new file was

Verified on this branch: `npm run typecheck` clean, `node --test test/*.test.cjs` 250 pass with only the 7 pre-existing Windows failures in `node-install`/`codex-remote`/`transcript-project-dir` (identical set on a clean checkout; untouched by this PR).

The config.ts and wizard-layout hunks are dropped — `main`'s versions win, no conflict left.

On the process point: fully my miss too. I picked #140 off the open-issues list and never checked whether `main` had already fixed it or whether another PR was in flight. A 'fixed in main' note on the issue would have saved both evenings; I'll check for that first next time.

PR #168, opened 2026-08-18
fix(build): unbreak npm run dist on Python 3.12+ — electron-builder pins a distutils-era node-gyp

Maev4l:
## What & why

`npm run dist:mac` (and `dist:win` / `dist:linux`) dies during packaging, in node-pty's native rebuild:

```
File ".../node_modules/node-gyp/gyp/pylib/gyp/input.py", line 19, in <module>
    from distutils.version import StrictVersion
ModuleNotFoundError: No module named 'distutils'
⨯ node-gyp failed to rebuild '.../node_modules/node-pty'
```

The chain: `electron-builder 25.1.8` → `app-builder-lib 25.1.8` → **`@electron/rebuild 3.6.1`** → **`node-gyp 9.4.1`**, whose bundled gyp still imports `distutils`. Python 3.12 removed distutils from the stdlib, so this breaks on any current interpreter — node-gyp picked Homebrew's Python 3.14 here.

Worth stressing: this is **not** the C/C++ toolchain failure CONTRIBUTING.md warns about. Xcode CLT is installed and healthy; the interpreter is the problem.

It also only affects electron-builder's copy. `postinstall`'s `electron-rebuild` uses the repo's own devDependency `@electron/rebuild` `^3.7.0`, which shells into `@electron/node-gyp` (gyp-next, no distutils import). That's why `npm install` and `npm run dev` are fine while `npm run dist:*` is not — the failure only shows up at packaging time.

The fix is one npm override, pointing `app-builder-lib` at the rebuilder this repo already depends on:

```json
"overrides": {
  "app-builder-lib": {
    "@electron/rebuild": "$@electron/rebuild"
  }
}
```

`$@electron/rebuild` reuses the devDependency spec rather than hardcoding a version, so the two cannot drift apart on a later bump.

Three notes for review:

- **Overriding `node-gyp` itself does not work.** v10+ converted `commands.*` to async, while `@electron/rebuild 3.6.1`'s worker still wraps them in `promisify()` — the callback never fires, so the rebuild would hang rather than fail. Replacing the rebuilder is the only version of this fix that works.
- **This override becomes redundant on electron-builder 26**, whose `app-builder-lib` already depends on a distutils-free rebuilder. Worth deleting during that upgrade so it can't pin an old rebuilder.
- **`package-lock.json` carries a little extra:** a pending `0.4.3` → `0.4.4` version sync and a few transitive dedupes from regenerating on npm 11. Happy to strip those, or leave the lock to you, if you'd prefer a tighter diff.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [x] Build / CI

## Screenshots / clips

None — no visual surface is touched.

## Checklist

- [x] `npm run typecheck` passes (exit 0, both the node and web projects).
- [x] `npm run build` succeeds — also exercised via `npm run dist:mac`, which runs it first.
- [x] Any new UI derives from `DESIGN.md` / `tokens.ts` — n/a, no UI change.
- [x] If I added art, it's my own or compatibly licensed, and listed in `ATTRIBUTION.md` — n/a, no art.

## Verification

macOS (Darwin 25.6, arm64), Node 24.19.0, npm 11.17.0, Electron 32.3.3. `npm run dist:mac` now runs to completion — the previously fatal step passes for both slices (`finished moduleName=node-pty arch=x64`, then `arch=arm64`) — and produces the universal dmg + zip with genuinely fat addons:

```
node-pty/build/Release/pty.node                    x86_64 arm64
better-sqlite3/build/Release/better_sqlite3.node   x86_64 arm64
```

Signing and notarization were skipped (no Developer ID in my keychain), which is the documented contributor-build path and unrelated to this change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "munder-difflin",
-  "version": "0.4.3",
+  "version": "0.4.4",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "munder-difflin",
-      "version": "0.4.3",
+      "version": "0.4.4",
       "hasInstallScript": true,
       "dependencies": {
         "@codemirror/lang-css": "^6.3.1",
@@ -1959,29 +1959,6 @@
         }
       }
     },
-    "node_modules/@openai/agents-core/node_modules/ws": {
-      "version": "8.21.0",
-      "resolved": "https://registry.npmjs.org/ws/-/ws-8.21.0.tgz",
-      "integrity": "sha512-Vsp28b7DRcimFQvrqu2Wek3z1iYxDCWqHYB8Qsnk/S4RfaCQzPGPyBNuVjJV3cd6UiKtUtp6sNM77gWvzcCH+g==",
-      "license": "MIT",
-      "optional": true,
-      "peer": true,
-      "engines": {
-        "node": ">=10.0.0"
-      },
-      "peerDependencies": {
-        "bufferutil": "^4.0.1",
-        "utf-8-validate": ">=5.0.2"
-      },
-      "peerDependenciesMeta": {
-        "bufferutil": {
-          "optional": true
-        },
-        "utf-8-validate": {
-          "optional": true
-        }
-      }
-    },
     "node_modules/@openai/agents-realtime": {
       "version": "0.11.8",
       "resolved": "https://registry.npmjs.org/@openai/agents-realtime/-/agents-realtime-0.11.8.tgz",
@@ -3201,35 +3178,6 @@
         "electron-builder-squirrel-windows": "25.1.8"
       }
     },
-    "node_modules/app-builder-lib/node_modules/@electron/rebuild": {
-      "version": "3.6.1",
-      "resolved": "https://registry.npmjs.org/@electron/rebuild/-/rebuild-3.6.1.tgz",
-      "integrity": "sha512-f6596ZHpEq/YskUd8emYvOUne89ij8mQgjYFA5ru25QwbrRO+t1SImofdDv7kKOuWCmVOuU5tvfkbgGxIl3E/w==",
-      "dev": true,
-      "license": "MIT",
-      "dependencies": {
-        "@malept/cross-spawn-promise": "^2.0.0",
-        "chalk": "^4.0.0",
-        "debug": "^4.1.1",
-        "detect-libc": "^2.0.1",
-        "fs-extra": "^10.0.0",
-        "got": "^11.7.0",
-        "node-abi": "^3.45.0",
-        "node-api-version": "^0.2.0",
-        "node-gyp": "^9.0.0",
-        "ora": "^5.1.0",
-        "read-binary-file-arch": "^1.0.6",
-        "semver": "^7.3.5",
-        "tar": "^6.0.5",
-        "yargs": "^17.0.1"
-      },
-      "bin": {
-        "electron-rebuild": "lib/cli.js"
-      },
-      "engines": {
-        "node": ">=12.13.0"
-      }
-    },
     "node_modules/app-builder-lib/node_modules/balanced-match": {
       "version": "4.0.4",
       "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-4.0.4.tgz",
@@ -3241,16 +3189,16 @@
       }
     },
     "node_modules/app-builder-lib/node_modules/brace-expansion": {
-      "version": "5.0.6",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-5.0.6.tgz",
-      "integrity": "sha512-kLpxurY4Z4r9sgMsyG0Z9uzsBlgiU/EFKhj/h91/8yHu0edo7XuixOIH3VcJ8kkxs6/jPzoI6U9Vj3WqbMQ94g==",
+      "version": "5.0.9",
+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-5.0.9.tgz",
+      "integrity": "sha512-ScQ4IuvIEF1TMlP7Zt+vjJ//9zlPb2SDcxWxM3bk8s6t6GGdJ7KO1dCcTidOPJKePW30LE/2cT7wCyPho9/Wxg==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
         "balanced-match": "^4.0.2"
       },
       "engines": {
-        "node": "18 || 20 || >=22"
+        "node": "20 || >=22"
       }
     },
     "node_modules/app-builder-lib/node_modules/fs-extra": {
@@ -3282,13 +3230,13 @@
       }
     },
     "node_modules/app-builder-lib/node_modules/minimatch": {
-      "version": "10.2.5",
-      "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-10.2.5.tgz",
-      "integrity": "sha512-MULkVLfKGYDFYejP07QOurDLLQpcjk7Fw+7jXS2R2czRQzR56yHRveU5NDJEOviH+hETZKSkIk5c+T23GjFUMg==",
+      "version": "10.2.6",
+      "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-10.2.6.tgz",
+      "integrity": "sha512-vpLQEs+VLCr1nU0BXS07maYoFwlDAH0gngQuuttxIwutDFEMHq2blX+8vpgxDdK3J1PwjCJiep77OitTZ4Ll1A==",
       "dev": true,
       "license": "BlueOak-1.0.0",
       "dependencies": {
-        "brace-expansion": "^5.0.5"
+        "brace-expansion": "^5.0.8"
       },
       "engines": {
         "node": "18 || 20 || >=22"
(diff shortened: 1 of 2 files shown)


chaitanyagiri:
## TL;DR

Thanks @Maev4l — this is a small, well-diagnosed, correct fix and I'd land it.

**What it does:** `npm run dist:*` dies during packaging with `ModuleNotFoundError: No module named 'distutils'`. The chain is `electron-builder 25.1.8` → `app-builder-lib` → its own nested **`@electron/rebuild 3.6.1`** → **`node-gyp 9.4.1`**, whose bundled gyp still does `from distutils.version import StrictVersion`. Python 3.12 deleted distutils from the stdlib, so packaging breaks on any current interpreter. The fix is a single npm `overrides` entry pointing `app-builder-lib` at the `@electron/rebuild` this repo already depends on, written as `"$@electron/rebuild"` so the two can never drift on a later bump.

**Verdict:** the diagnosis is exactly right — I verified the offending import line-by-line in both node-gyp versions — and the override provably removes node-gyp 9 from the tree. `npm ci` still works, typecheck is green, 258/258 tests pass. No malicious code, no NSFW, no prompt injection. **Nothing blocking.** Two notes: the lockfile carries a little unrelated churn you already offered to strip, and this PR overlaps #170, which needs a decision from the maintainer.

---

## What I did

Checked out `fix/node-gyp-python312`, installed it, and verified the causal chain end-to-end rather than taking the traceback at face value — including fetching both node-gyp versions' `input.py` to confirm the exact import that changed.

## The chain, confirmed

```
  electron-builder 25.1.8
        └── app-builder-lib 25.1.8
              └── @electron/rebuild 3.6.1        (nested copy)
                    └── node-gyp 9.4.1
                          └── gyp/pylib/gyp/input.py:19
                                from distutils.version import StrictVersion   💥 Python 3.12+

  …while, separately:

  postinstall → electron-rebuild
        └── @electron/rebuild 3.7.2               (this repo's own devDependency)
              └── @electron/node-gyp 10.2.0       ← gyp-next, no distutils
```

I pulled both files from source to be sure:

| Version | `gyp/pylib/gyp/input.py` line 19 |
|---|---|
| node-gyp **v9.4.1** | `from distutils.version import StrictVersion` ← the failure |
| node-gyp **v10.0.0** | `from packaging.version import Version` ← clean |

That matches the traceback in your description exactly, including the line number. It also explains the confusing part of the bug — why `npm install` and `npm run dev` are fine while only `npm run dist:*` fails — and your note that this is *not* the Xcode-CLT failure CONTRIBUTING warns about is the right thing to have written down, because that's where everyone will look first.

**A detail worth stating explicitly:** before this PR the project rebuilt its native addons **twice, with two different rebuilders** — 3.7.2 at `postinstall`, then 3.6.1 again inside electron-builder at packaging time. After the override there is one rebuilder for both. That's a consistency win beyond just unbreaking the build, and I'd mention it in the commit message.

## What I tested

| Check | Result |
|---|---|
| Override actually takes effect | ✅ `node_modules/app-builder-lib/node_modules/@electron/rebuild` **no longer exists** — it dedupes to the root 3.7.2 |
| node-gyp 9 gone from the tree | ✅ the only gyp left is `@electron/node-gyp 10.2.0-electron.1`; `node-gyp@9.4.1` is absent entirely |
| `npm ci --dry-run` | ✅ clean — lockfile and `package.json` in sync (worth checking, since a hand-edited `overrides` block is exactly the thing that desyncs a lock) |
| `node --test test/*.test.cjs` | ✅ 258/258, identical to `main` |
| Lockfile supply-chain scan | ✅ no new packages, no non-npmjs URLs, integrity hashes intact. The change is almost entirely *removals* — node-gyp 9 and its abandoned transitive tail (`npmlog`, `gauge`, `are-we-there-yet`, `wide-align`, `console-control-strings`, `set-blocking`…), all packages npm itself marks "no longer supported" |
| Malware / obfuscation / network calls / NSFW / prompt injection | ✅ none — the functional diff is five lines of JSON |

**One thing I could not reproduce:** the original failure itself. My Python is 3.10.4, which still ships distutils, so `dist:mac` wouldn't have failed here regardless. I verified the mechanism instead of the symptom; combined with your `dist:mac` run producing fat `pty.node` and `better_sqlite3.node` for both slices, I'm satisfied.

## Review notes

**1. Lockfile churn (your call, and you already flagged it).** Beyond the intended removals, the lock also carries a `0.4.3 → 0.4.4` version sync (harmless, it's catching up to `package.json`), a deletion of the `@openai/agents-core/node_modules/ws` entry, and npm-11 dedupe bumps inside `app-builder-lib` (`brace-expansion` 5.0.6→5.0.9, `minimatch` 10.2.5→10.2.6, `semver` 7.8.1→7.8.5). None of it is wrong, but the `ws` removal and the dedupes aren't this PR's concern and make the diff harder to read six months from now. Since you offered — yes, I'd strip them, keeping only the node-gyp removals and the version sync.

**2. The override needs an exit plan, which you've written but nothing enforces.** You correctly note it becomes redundant on electron-builder 26. `package.json` can't carry a comment to say so. Worth either an issue linked from this PR, or a line in `CONTRIBUTING.md`'s build section — otherwise it silently pins an old rebuilder through some future upgrade.

**3. Not caused by this PR, but now more load-bearing.** `@electron/rebuild 3.7.2` resolves `@electron/node-gyp` from a **git URL** (`git+ssh://git@github.com/electron/node-gyp.git#06b29aaf…`). It's commit-pinned and has an integrity hash, and it's already on `main`, so nothing changes — but after this PR that git dependency is on the packaging path as well as the install path. Just something to know, not an objection.

**4. Overriding `node-gyp` directly doesn't work.** I didn't test your `promisify()` hang claim, but the reasoning is sound and the fix you chose is the more robust one anyway — replacing the rebuilder rather than patching under it.

## Questions

@Maev4l — want to push a lock-only cleanup commit stripping the `ws` removal and the dedupes, or would you rather leave the lock as-is and let the maintainer regenerate? Either is fine by me; I'd just rather it be a decision than an accident.

@chaitanyagiri — this needs a call from you, and it's the same one I raised on #170: that PR moves to electron-builder 26, which makes this override unnecessary, and its branch already contains this commit and then removes the override again. **#170 currently fails CI at `npm ci`** (details in my review there), so if you want the packaging bug fixed *now*, this PR is the cheap, low-risk, fully-verified way to do it — merge it, and delete the override when #170 lands. Want me to open a tracking issue for that deletion?

Maev4l:
Thanks for the thorough pass — and for verifying the mechanism rather than the symptom. Your point that the project was rebuilding native addons **twice with two different rebuilders** (3.7.2 at postinstall, 3.6.1 again inside electron-builder) is a better framing of the win than "unbreaks packaging", and I hadn't stated it that way.

**Lock cleanup: done, pushed as `22e7bdf`.** Answering your question directly — I'd rather it be a decision than an accident, so I stripped it.

The lock now differs from `main` by exactly the intended removals plus the version sync. Total PR diff is 2 files, **+7 / −224** (was +18 / −258).

Stripped:
- restored `node_modules/@openai/agents-core/node_modules/ws`, which had been deduped away
- reverted the npm-11 re-resolution bumps under `app-builder-lib`: `brace-expansion` 5.0.9 → 5.0.6, `minimatch` 10.2.6 → 10.2.5, `semver` 7.8.5 → 7.8.1

Kept — 16 removals and one version change:
- `node-gyp 9.4.1` plus its nested `brace-expansion`, `glob`, `minimatch`, `semver`
- the abandoned tail it dragged in: `npmlog`, `gauge`, `are-we-there-yet`, `wide-align`, `set-blocking`, `has-unicode`, `delegates`, `console-control-strings`, `color-support`, `aproba`
- `app-builder-lib`'s nested `@electron/rebuild 3.6.1`
- lock version `0.4.3 → 0.4.4`, catching up to `package.json`

**Where the churn came from**, since it's a trap worth naming: my first pass deleted *every* `app-builder-lib/node_modules/*` entry to force npm to re-resolve the override, so its siblings got re-resolved too. This time I rebuilt the lock from `main` and deleted only the entries the fix actually retires, which avoids the re-resolution entirely rather than editing its output back out.

**Verification** — a hand-built lock is exactly the thing that desyncs, so:
- `npm ci --dry-run` passes under **npm 11.17.0** *and* **npm 10.9.9** — no `EUSAGE`, nothing missing from the lock. I checked npm 10 specifically because that's the family Node 20 ships and what `ci.yml` uses.
- The tree the lock implies is unchanged from the version you reviewed: `app-builder-lib` still resolves `@electron/rebuild` to the hoisted 3.7.2, and `node-gyp@9.4.1` is absent from the tree entirely.
- I did not re-run a full `npm install` + `dist:mac` for this commit — all four restored details sit outside the `@electron/rebuild` chain, so the fix's behaviour can't change. Say the word if you'd like the belt-and-braces run anyway.

**On your item 2 (the override needs an enforced exit plan).** Agreed, and I'd rather not bundle docs into a PR you've said you'd land as-is. Your offer of a tracking issue is the cleaner route — happy either way, including adding a line to `CONTRIBUTING.md`'s build section here if @chaitanyagiri would prefer it in one commit.

**On item 3 (`@electron/node-gyp` resolving from a commit-pinned git URL).** Noted, and you're right that this PR puts it on the packaging path as well as the install path. It's pre-existing on `main` and commit-pinned with an integrity hash, so I've left it alone rather than widening scope.

**On item 4.** Fair — I didn't need to prove the `promisify()` hang to reject that route, and replacing the rebuilder is the more robust fix regardless. For the record the reasoning is that `@electron/rebuild` 3.6.1's worker calls `promisify(nodeGyp.commands[name])(args)`, while node-gyp 10+ converted `commands.*` to async, so the callback never fires and it would hang rather than fail — a worse failure mode than the one being fixed.

Separately, I'm working through your #170 review, starting with regenerating that lock so `npm ci` works under npm 10 — thanks for catching that it would break CI at the install step before typecheck ever runs.

chaitanyagiri:
Closing this as shipped, and apologising for how long it took.

Your commit `9bb060dd`, the distutils free node-gyp rebuild, has been in the shipping build since before 0.4.5. It is in v0.4.6 as well. It has been fixing `npm run dist` on Python 3.12 and up for people this whole time while this PR sat open, which is not how it should have gone.

The only thing here that did not ship is `22e7bdf1`, the lock file churn cleanup, which is housekeeping rather than the fix.

Thank you, and sorry for the wait.

PR #169, opened 2026-08-18
feat: added support for kiro-cli at god and agent mode

rikameajay1:
## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ visualized as avatars at work on a shared office floor.
 > [!NOTE]
 > **The world's best agents. The world's worst paper company.**
 > Munder Difflin takes the terminal-agent CLIs you already run — `claude`, `agy`, `codex`, `grok`,
-> `kimi`, `qwen`, `opencode`, `crush`, `pi`, and `copilot` — and turns them
+> `kimi`, `qwen`, `opencode`, `crush`, `pi`, `copilot` and `kiro` — and turns them
 > into a self-coordinating team: each agent gets long-term memory, a mailbox, and a desk on a 2D
 > office floor — and **your clone** (Michael) routes work between them while you watch. He's the
 > boss of the floor; you're still the boss of him.
@@ -74,7 +74,7 @@ wires them into a **hive mind**, and puts **your clone** in charge — Michael, 
 talk to in order to get things done. Under the hood it runs the **fastest memory layer in the
 world** so every agent remembers what it learns and recalls it instantly.
 
-- **Every terminal is an agent.** Each `claude`, `agy`, `codex`, `grok`, `kimi`, `qwen`, `opencode`, `crush`, `pi`, `copilot`, or custom session runs as a real
+- **Every terminal is an agent.** Each `claude`, `agy`, `codex`, `grok`, `kimi`, `qwen`, `opencode`, `crush`, `pi`, `copilot` and `kiro`, or custom session runs as a real
   process in a pseudo-terminal (`node-pty`), byte-for-byte authentic, rendered with xterm.js.
 - **Every agent is an avatar.** Sessions appear as characters on a Pixi.js office floor — they walk
   to stations as they work, and envelopes fly desk-to-desk when they message each other.
@@ -174,7 +174,7 @@ terminal/event plane, and [`DESIGN.md`](./DESIGN.md) for the visual system.
 - At least one supported agent CLI on your `PATH` — **[Claude Code](https://claude.com/claude-code)**
   (`claude`, the default), **Antigravity** (`agy`), **OpenAI Codex** (`codex`), **xAI Grok** (`grok`),
   **Kimi Code** (`kimi`), **Qwen** (`qwen`), **OpenCode** (`opencode`), **Crush** (`crush`),
-  **pi.dev** (`pi`), or **GitHub Copilot** (`copilot`). Most missing CLIs self-heal: the harness runs the installer in the
+  **pi.dev** (`pi`), **GitHub Copilot** (`copilot`) or **Kiro Dev** (`kiro`). Most missing CLIs self-heal: the harness runs the installer in the
   terminal and continues into the new binary.
 - *Optional:* **your own API keys and local LLMs** in **Settings → AI Engines** (Ollama / LM Studio / vLLM).
 - *Optional:* the semantic memory index for instant cross-session recall — markdown memory works without it.
@@ -288,7 +288,7 @@ chrome. The 15 avatars are the cast of *The Office*, differentiated by hair/skin
 
 ## Roadmap
 
-Shipped through **v0.4.3** — ten agent engines with BYOK keys and local LLMs, voice orchestration,
+Shipped through **v0.4.3** — eleven agent engines with BYOK keys and local LLMs, voice orchestration,
 the hive (memory · mailboxes · blackboard · event log), Command Center with kanban and schedules,
 a built-in Monaco IDE with git rails, integrations registry + secret broker, Slack-spawned workers,
 shareable hires and the Agent Gallery, observability and the circuit breaker, durable persistence,
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "munder-difflin",
-  "version": "0.4.3",
+  "version": "0.4.4",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "munder-difflin",
-      "version": "0.4.3",
+      "version": "0.4.4",
       "hasInstallScript": true,
       "dependencies": {
         "@codemirror/lang-css": "^6.3.1",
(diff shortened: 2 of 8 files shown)


chaitanyagiri:
## TL;DR

Thanks @rikameajay1 — adding an eleventh engine is a welcome contribution, and the preset is structured correctly: you found `autoModeFlag`/`autoFlag`, `positionalInitialPrompt`, the exhaustive `CONTEXT_COMMANDS` record, and the `toolCatalog` binary probe. That last fix (probe only the leading binary when a `defaultCommand` carries a subcommand) is a genuinely good catch that the codebase needed.

**But I can't recommend merging this as-is.** Three things, in order of seriousness:

1. 🔴 **The install command points at an npm package that does not exist.** `npm install -g @anthropic-ai/kiro` returns **404**. Kiro is not distributed on npm at all — the official install is a shell installer. This string is *executed*, not just displayed.
2. 🔴 **The test suite goes red.** `provider-config.test.cjs` fails on this branch (257/258). Your checklist ticks typecheck and build — both genuinely pass — but the tests were never run.
3. 🟡 **The model IDs and two flags are unverified guesses**, presented in a comment as "the known options".

None of it is malicious — no NSFW, no profanity, no prompt injection, no obfuscation, no exfiltration. It's an accuracy problem, and it's fixable in an hour.

---

## What I did

Checked out the branch, ran typecheck, the production build and the full test suite, then verified every external fact the preset asserts against **the official Kiro documentation and the live npm registry** rather than accepting the values in the diff.

## Finding 1 — the install command is wrong, and it gets run

`installCommand` isn't a hint. When an engine's binary is missing, `buildMissingCliScript()` (`src/main/cliInstall.ts`) picks a rung of the install ladder and **runs the command visibly in the user's terminal**:

```
  user picks Kiro → kiro-cli not on PATH
        │
        ▼
  chooseInstallRung()
        │  npm is available → rung "npm"
        ▼
  buildMissingCliScript() prints a banner, then RUNS:
        npm install -g @anthropic-ai/kiro
        │
        ▼
        npm error 404 Not Found — '@anthropic-ai/kiro' is not in this registry
```

I checked the registry directly: `@anthropic-ai/kiro` → **404**. Every user who tries the Kiro engine hits that error. The comment `// trusted, hardcoded` is doing a lot of work for a package that isn't published.

**Please do not "fix" this by dropping the scope.** Both bare names are taken on npm by unrelated parties:

| Name | What's actually there |
|---|---|
| `@anthropic-ai/kiro` | does not exist (404) |
| `kiro@1.0.0` | description: *"a simple project"* |
| `kiro-cli@0.0.1` | description: *"Kiro CLI - placeholder"*, maintained by a personal Gmail address — not AWS |

`npm install -g` runs install scripts, so shipping either of those would have the app globally install a stranger's package on a user's machine. Worth flagging loudly because it's exactly the shape of fix a reviewer might wave through.

**The correct fix is already a field in this file.** Per kiro.dev/docs the real installer is:

- macOS/Linux: `curl -fsSL https://cli.kiro.dev/install | bash`
- Windows: `irm 'https://cli.kiro.dev/install.ps1' | iex`

That's precisely what `nativeInstallCommand: { posix, win32 }` exists for — Claude Code and OpenCode already use it for the same shape of installer. So: **drop `installCommand` entirely** (there is no npm package to name) and add `nativeInstallCommand`. `chooseInstallRung` handles a preset with no npm command; it falls through to the native rung.

## Finding 2 — the test suite fails

```
not ok 132 - God only sees providers that can drain hive inbox messages
   test/provider-config.test.cjs:103
   + actual   - expected
     'pi',
   + 'kiro'
```

The test pins the exact god-eligible provider list. Because you set `supportsModel: true` **and** `canReceiveInbox: true`, Kiro becomes god-eligible and the list changes. Both `assert.deepEqual` calls in that test need `'kiro'` added (the second one fails too — it just never runs, because the first throws first).

While you're there: house convention gives every first-class provider its own case in that file (`'Kimi is a first-class inferred provider…'`, `'Grok is a first-class inferred provider…'`). A `'Kiro is a first-class inferred provider'` test asserting `isAgentProvider('kiro')`, `inferAgentProvider('kiro-cli chat')`, and the preset's flags is expected here and is about fifteen lines.

## Finding 3 — verified vs guessed

I checked each claim against the docs so you know exactly which ones to re-check:

| Claim in the preset | Verdict |
|---|---|
| `defaultCommand: 'kiro-cli chat'` | ✅ `chat` is a real subcommand |
| `resumeFlag: '--resume-id'` | ✅ docs show `kiro-cli chat --resume-id <id>` |
| `positionalInitialPrompt: true` (`chat [INPUT]`) | ✅ consistent with the documented signature |
| `modelFlag: '--model'` | ❓ **not found anywhere in the CLI docs** |
| `autoModeFlag: '-a'` / `--trust-all-tools` | ❓ **not found in the docs either** |
| `KIRO_MODELS` — `claude-sonnet-5`, `claude-opus-4.8`, `claude-haiku-4.5` | ❓ unverified |
| `installCommand` | ❌ wrong (Finding 1) |

Two of those are security-relevant: `-a` is described in your own comment as *"allows the model to use any tool without asking for confirmation."* If that flag string is wrong, auto-mode silently does nothing and users think they're in auto when they aren't; if it's right, it belongs in the review either way. The wiring is correct — you set both `autoModeFlag` and `autoFlag`, which matches every other preset — it's only the literal I can't confirm.

For the model list, the repo has a convention for unverified ids: `COPILOT_MODELS` carries an explicit `// TODO-verify exact live ids` marker. `KIRO_MODELS` currently says "these are the known options", which reads as verified. Either check them against `kiro-cli chat --help` or copy the TODO-verify marker.

## Smaller things

- **Unrelated lockfile churn.** The diff bumps `package-lock.json` 0.4.3 → 0.4.4 (harmless, syncs with `package.json`) but also **deletes** the `@openai/agents-core/node_modules/ws` entry. A provider preset shouldn't change the dependency tree — please revert the lockfile.
- **The PR description is an empty template.** "What & why" is blank, and the checklist's screenshot requirement is unticked even though `OnboardingWizard.tsx` gains a visible row. One screenshot of the Kiro option in onboarding would satisfy it.
- **README isn't updated** — it says "ten agent engines" and lists the ten CLIs by name. This makes eleven.
- **Branched from your fork's `main`.** CONTRIBUTING asks for a topic branch; not blocking, but it makes follow-up commits awkward for you.
- The `p.defaultCommand.trim().split(/\s+/)[0]` expression appears twice in `toolCatalog.ts`. Worth a `const bin = …` above the object literal.

## What I tested

| Check | Result |
|---|---|
| `npm run typecheck` | ✅ exit 0 |
| `node --test test/*.test.cjs` | ❌ **257/258** — `provider-config.test.cjs:103` (`main` is 258/258) |
| `npm view @anthropic-ai/kiro` | ❌ 404 |
| Official install method (kiro.dev/docs) | shell installer, **not npm** |
| Malware / obfuscation / network calls / prompt injection / NSFW | ✅ none |

## Questions

@rikameajay1 — can you confirm `--model` and `-a` from a real `kiro-cli chat --help` on your machine, and paste the output? That single paste settles three rows of the table above. And please swap `installCommand` for `nativeInstallCommand` rather than renaming the npm package.

@chaitanyagiri — one design call for you: this preset makes Kiro **god-eligible** (orchestrator) while `hiveAware: false` and there's no hook bridge, so the only inbox drain path is the renderer idle nudge. OpenCode and Crush were in that position before their bridges landed — were they god-eligible then too, or should a new no-bridge provider start as agent-only until a bridge exists?

rikameajay1:
```

kiro-cli chat --help
AI assistant in your terminal

Usage: kiro-cli-chat chat [OPTIONS] [INPUT]

Arguments:
  [INPUT]
          The first question to ask

Options:
  -r, --resume
          Resume the most recent conversation from this directory

      --resume-id <SESSION_ID>
          Resume a specific conversation by session ID

      --resume-picker
          Interactively select a conversation to resume from this directory

      --agent <AGENT>
          Context profile to use

      --model <MODEL>
          Current model to use

      --effort <EFFORT>
          Initial effort level (e.g. low, medium, high, xhigh, max)

  -a, --trust-all-tools
          Allows the model to use any tool to run commands without asking for confirmation

      --trust-tools <TOOL_NAMES>
          Trust only this set of tools. Example: trust some tools: '--trust-tools=fs_read,fs_write', trust no tools: '--trust-tools='

      --no-interactive
          Whether the command should run without expecting user input

  -l, --list-sessions
          List all saved chat sessions for the current directory

      --list-models
          List available models and exit

  -f, --format <FORMAT>
          Output format for list commands (used with --list-models and --list-sessions)

          Possible values:
          - plain:       Outputs the results as markdown
          - json:        Outputs the results as JSON
          - json-pretty: Outputs the results as pretty print JSON
          
          [default: plain]

  -d, --delete-session <SESSION_ID>
          Delete a saved chat session by ID

      --session-source <v1|v2>
          Target only v1 or v2 store for --delete-session (default: both)
          
          [possible values: v1, v2]

  -w, --wrap <WRAP>
          Control line wrapping behavior (default: auto-detect)

          Possible values:
          - always: Always wrap at terminal width
          - never:  Never wrap (raw output)
          - auto:   Auto-detect based on output target (default)

      --require-mcp-startup
          Require all enabled MCP servers to start successfully; exit with code 3 if any fail

      --tui
          Use the new terminal UI

      --legacy-ui
          Use the legacy harness
          
          [aliases: --classic]

      --agent-engine <ENGINE>
          Agent engine to use: "v1", "v2" (default), or "v3"
          
          [possible values: v2, v1, v3]

  -v, --verbose...
          Increase logging verbosity

      --v3
          Launch the next generation Kiro agent

      --mode <MODE>
          Mode to use with the V3 agent: "default" or "spec"
          
          [possible values: default, spec]

      --cloud
          Run this session in a cloud sandbox (V3/KAS only). Forwarded to the TUI, which sends _meta.kiro.executionTarget on session/new

      --repo <REPO>
          Repository to open in a cloud session, as name or owner/name (comma-separated for multiple). Only meaningful with --cloud

  -h, --help
          Print help (see a summary with '-h')

```

rikameajay1:
image

image

image

image

@chaitanyagiri , added screenshots and kiro-cli help

PR #170, opened 2026-08-18
chore(deps): take npm audit from 28 advisories to 0 — electron-builder 26, vite 7, electron 43, drop dead localtunnel

Maev4l:
## What & why

`npm audit` reported **28 advisories (1 critical, 23 high, 4 moderate)**. This takes the tree to **0**, and drops it from 876 to 695 packages. One commit per cluster, so any single cluster can be dropped or cherry-picked.

| Commit | Change | Audit |
|---|---|---|
| `chore(build)` | electron-builder **26.15.7**, @electron/rebuild **4** | 28 → 14 |
| `chore(deps)` | vite **7.3.6**, electron-vite **5**, @vitejs/plugin-react **5** | 14 → 11 |
| `chore(deps)` | @openai/agents-realtime **0.16.1** | 11 → 8 |
| `chore(deps)!` | electron **43.4.0**, better-sqlite3 **13** | 8 → 6 |
| `chore(deps)` | remove dead `localtunnel`, then non-breaking `audit fix` | 6 → **0** |

### Relationship to #168 — your call

**This PR may supersede #168.** That PR adds an npm override so electron-builder 25 stops using a distutils-era node-gyp (packaging is broken on Python 3.12+). electron-builder 26 depends on `@electron/rebuild ^4` directly, which brings node-gyp 12, so the override becomes unnecessary — this branch removes it again.

Three paths, all ending in the same place, and it is entirely up to you:

1. **Merge #168, then this.** Fine. The override block survives the merge as a harmless leftover (26 already resolves `@electron/rebuild` to 4.x), deletable whenever you like.
2. **Drop/close #168 and merge this alone.** Also fine — the net diff here contains no override at all, and the distutils problem is gone because node-gyp 9 leaves the tree entirely.
3. **Merge #168 now and hold this.** If the Electron bump needs more thought, #168 remains the cheap unblock on its own.

I checked the merge math for 1 and 2: neither conflicts. Same author on both PRs, so close whichever you prefer without ceremony.

### Two blockers electron-builder 26 surfaces

Both are invisible on 25, and both are fixed here:

1. **A heap OOM at "searching for node modules"** killed every `dist:*`. `tunnelmole@2.4.0` lists **itself** in its own `dependencies` (`"tunnelmole": "^2.1.6"`), and 26's npm collector recurses the `npm list --json` tree with no visited set, so the walk never terminates. A larger `--max-old-space-size` does not help; growth is unbounded. Same shape as electron-userland/electron-builder#10068 (reported there via `libsql`). Fixed by `tools/patch-tunnelmole-selfdep.cjs`, following the existing `patch-node-pty-conpty.cjs` postinstall pattern. Confirmed by experiment: stripping that single field is what lets the build proceed.
2. **`@electron/universal` 2.0.3 refuses to merge Mach-O files that are byte-identical in both universal slices.** Three qualify: better-sqlite3's `test_extension.node` (a test fixture — now excluded from the package, since it was dead weight in the app), plus node-pty's stale `bin/darwin-arm64-128` build and its per-platform prebuilds, and better-sqlite3 13's new prebuilds. The node-pty and better-sqlite3 prebuilds are declared via `x64ArchFiles` rather than dropped, so the loader keeps its fallback.

### Notes on version choices

- **vite 7, not 8.** `npm audit fix --force` proposes vite 8, but electron-vite 5 peer-depends on `^5 || ^6 || ^7` and vite 8 support exists only in electron-vite `6.0.0-beta`. The advisories all cap at `<=6.4.2`, so 7.3.6 is already clear — 8.x was "latest", not "minimum safe". `@vitejs/plugin-react` stays on 5.x for the same reason (6.x requires vite `^8`).
- **electron-builder 26.15.7, not 26.15.3.** The `latest` dist-tag is several patches behind on the 26 line; `v26` points at 26.15.7.
- **electron 43.4.0**, the latest stable. The advisory floors are lower — 39.8.10 / 40.10.3 / 41.7.2 / 42.3.4 each clear every listed CVE — so 41.x or 42.x are cheaper landing spots if 43 is too much. 43 buys the longest support runway; 32 is eleven majors past end of support.
- **`localtunnel` is removed, not upgraded, because nothing uses it.** CHANGELOG v0.2.4 records tunnelmole replacing it; `slack.ts` and `webhook.ts` both import tunnelmole, and no file outside `package.json` mentions localtunnel. It was still pulling `axios 0.21.4` — the oldest thing in the tree — with its CSRF and SSRF advisories. With every vulnerable root then updated, the plain non-breaking `npm audit fix` finally clears the tail (`fast-uri`, `js-yaml`, `nanoid`, `postcss`); before this work that command changed nothing.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [x] Build / CI

## Screenshots / clips

No visual surface is touched by this change. Electron 43 does move Chromium (128 → 150), so the packaged build was installed to `/Applications` and exercised by hand against a real profile on macOS arm64 — no regressions observed.

## Checklist

- [x] `npm run typecheck` passes (exit 0, both projects) — re-run after every cluster.
- [x] Any new UI derives from `DESIGN.md` / `tokens.ts` — n/a, no UI change.
- [x] If I added art, it's my own or compatibly licensed — n/a, no art.
- [x] PR is focused on a single change — one concern (clear the advisories), five reviewable commits.

## Verification

macOS arm64 (Darwin 25.6), Node 24.19.0, npm 11.17.0.

`npm run typecheck` and `npm run dist:mac` pass at **every** cluster, not just at the end. Because typechecking an Electron bump proves very little, the packaged artifacts were also driven directly, through the real `app.asar` load path:

- **better-sqlite3 13** opens a database and round-trips a row — sqlite 3.53.4 under Electron 43's Node 24.18.1 (ABI 148).
- **node-pty** spawns a pty and reads its output back, in both the universal package and a separate arm64-only package.
- Both shipped addons are genuinely fat where expected: `node-pty` `pty.node` is `x86_64 arm64`.
- The packaged app launches and reaches `[updater] native updater ready`.
- Finally, installed to `/Applications` and used by hand (Electron 43.4.0 / Chromium 150 confirmed in the running bundle).

**Not verified:** the Windows and Linux targets — I only have macOS. `dist:win` / `dist:linux` should be exercised before release, particularly because the tunnelmole patch and the `x64ArchFiles` rule are new inputs to packaging (the latter is mac-only).

**One packaging regression worth knowing:** better-sqlite3 13 ships prebuilds for eight platforms and no longer builds into `build/Release`, so the mac dmg grew **233MB → 274MB**. Six of those prebuilds cannot run on macOS; a per-platform `files` rule would reclaim roughly 35MB. Left out of this PR to keep it to one concern — happy to follow up.


chaitanyagiri:
## TL;DR

Thanks @Maev4l — this is the most thoroughly documented PR in the queue, and the investigation behind it (especially the tunnelmole self-dependency) is genuinely good work.

**What it does:** takes `npm audit` from 28–29 advisories (1 critical, 24 high) to **0**, and the tree from 876 → 695 packages, across five reviewable commits: electron-builder 26 + @electron/rebuild 4, vite 7 + electron-vite 5 + plugin-react 5, @openai/agents-realtime 0.16.1, electron 43 + better-sqlite3 13, and removing the dead `localtunnel` dependency. Along the way it fixes two packaging failures that only appear once electron-builder 26 is in play.

**Verdict:** no malicious code, no NSFW/profanity, no prompt injection. Every headline claim reproduces on my machine — I checked each one rather than trusting the description. **Two things must be resolved before merge:** `npm ci` **fails on this branch** under npm 10 (a control run on `main` succeeds, so this PR introduces it), and vite 7 raises the real Node floor above the "Node.js 18+" both the README and CONTRIBUTING still promise. Electron 32 → 43 also needs a Windows sanity pass given what v0.4.4 was about.

---

## What I did

Checked out `chore/deps-migration`, installed it from scratch, and independently reproduced each claim: package count, advisory count, type-check, production build, the full test suite, the tunnelmole patch (including its idempotency), and whether the removed `localtunnel` really is unreferenced. Where a result could be pre-existing rather than caused by this PR, I ran the same check on `main` as a control.

## What I tested

| Check | Result |
|---|---|
| `npm audit` on `main` (control) | 29 advisories — 4 moderate, 24 high, 1 critical |
| `npm audit` on this branch | ✅ **found 0 vulnerabilities** |
| Package count | ✅ **695** installed — exactly the number claimed |
| `npm run build` | ✅ built in 25.8s, `copy:main-assets` ran |
| `node --test test/*.test.cjs` | ✅ 258/258, identical to `main` |
| `tools/patch-tunnelmole-selfdep.cjs` | ✅ works — `dependencies.tunnelmole` was `^2.1.6` before, `undefined` after; second run exits 0 silently, so it is genuinely idempotent |
| `localtunnel` actually dead | ✅ zero references across `src/`, `tools/`, `scripts/`, `hive/` — only `package.json` mentioned it |
| Installed versions | ✅ electron 43.4.0, better-sqlite3 13.0.3, `@types/better-sqlite3` 9.6.0 (better-sqlite3 still ships no bundled `.d.ts`, so the `@types` bump is the right call, not redundant) |
| Lockfile supply-chain scan | ✅ **clean** — all 220 added `resolved` URLs point at `registry.npmjs.org`, every one has a matching `integrity` hash, no git/tarball/proxy URLs, no unexpected new roots. `@modelcontextprotocol/client`+`core` look new but simply replace the `@modelcontextprotocol/sdk` that was already there (upstream repackaging), and both stay `optional: true`. `aws4` is dev-only. |
| Malware / obfuscation / exfiltration | ✅ none — the only new executable file is the 30-line tunnelmole patch, which reads one manifest, deletes one key, writes it back |

## The tunnelmole find is real, and worth calling out

This is the part of the PR I'd have most doubted, so I checked it directly — the diagnosis is correct:

```
  package.json (tunnelmole@2.4.0)
  ┌──────────────────────────────┐
  │ "dependencies": {            │
  │    "tunnelmole": "^2.1.6"  ──┼──┐   ← the package depends on ITSELF
  │ }                            │  │
  └──────────────────────────────┘  │
                ▲                   │
                └───────────────────┘

  electron-builder 26's npm collector:
      walk(dep) → for each dep.dependencies → walk(dep)     (no visited set)

      walk(tunnelmole) → walk(tunnelmole) → walk(tunnelmole) → … 💥 heap OOM
                         at "searching for node modules"
```

I confirmed `node_modules/tunnelmole/package.json` really does list itself, and that the patch removes exactly that one key and nothing else. It follows the existing `patch-node-pty-conpty.cjs` postinstall pattern, so it matches house style.

## Review — what needs resolving

**1. `npm ci` fails on this branch (please fix).** This is the one hard blocker I found:

```
$ npm ci --ignore-scripts          # npm 10.9.7, Node 22.22.2, on chore/deps-migration
npm error code EUSAGE
npm error `npm ci` can only install packages when your package.json and
npm error package-lock.json … are in sync.
npm error Missing: ws@8.21.3 from lock file
```

The same command on `main` succeeds, so it is not pre-existing. The cause is `openai@7.5.0` (pulled in by `@openai/agents-realtime` 0.16.1) declaring an **optional peer** `ws: ^8.18.0`; npm 11.17.0 (what you used) resolves that without a top-level entry, npm 10.x wants `ws@8.21.3` recorded. Running plain `npm install` on npm 10 immediately adds those 24 lines back, which is the drift in one line of evidence. Since npm 10 is what ships with Node 22 LTS — the version most contributors will have — the lock should be regenerated so it satisfies both, or the project needs to declare the npm it expects.

**2. The Node floor moved and the docs still say 18.** vite 7's own manifest reads `"node": "^20.19.0 || >=22.12.0"`. `README.md:169` and `CONTRIBUTING.md` both still say **"Node.js 18+"**. On Node 18 this branch simply won't build. Worth adding an `engines` field to `package.json` so the failure is a clear npm error rather than a confusing vite crash, and updating both docs in the same commit.

**3. Electron 32 → 43 needs a Windows pass before merge.** You're explicit that you only have macOS, which I appreciate — but eleven majors of Electron and Chromium 128 → 150 is the single riskiest thing in this PR, and v0.4.4's entire headline was *fixing Windows*. `tools/patch-node-pty-conpty.cjs` exists specifically because Windows conpty needed patching, and node-pty/better-sqlite3 both rebuild against a new ABI (Node 24.18.1 / ABI 148). A `dist:win` run plus one two-agent message round-trip on Windows would close this. If nobody can do that quickly, landing electron **41.x or 42.x** — which you note also clear every listed CVE — gets the advisories to zero at a fraction of the risk.

**4. Minor / questions, non-blocking.** The `x64ArchFiles` glob declares node-pty's stale `bin/darwin-arm64-128` as single-arch rather than excluding it — but you *did* exclude better-sqlite3's `test_extension.node` outright. Why the different treatment for a build you describe as stale? And the postinstall patch only reaches the hoisted `node_modules/tunnelmole`; if npm ever nests a second copy the OOM returns silently. An npm `overrides` entry would be install-script-independent, though I recognise it fights npm's self-reference handling — your call, and the current approach does match the existing patch script.

## Questions

@Maev4l — (1) can you regenerate the lockfile under npm 10 (or confirm which npm the project should standardise on) so `npm ci` works for both? (2) Is the stale `node-pty/bin/darwin-arm64-128` safe to exclude from `files` entirely, the way you handled `test_extension.node`?

@chaitanyagiri — two decisions are yours, not the author's. (a) **Electron 43 or 42?** 43 buys the longest support runway; 42.3.4 clears the same CVEs with far less Chromium delta and much less Windows risk right after a Windows-focused release. (b) **#168 vs this PR** — the author laid out three merge paths and I confirmed the branch already contains #168's commit and then reverts the override, so merging #168 first is non-conflicting. Please pick one so both PRs can move; right now each is waiting on the other.

chaitanyagiri:
**Addendum to my review above — the `npm ci` failure is a hard CI break, not just a local annoyance.**

After posting I checked `.github/workflows/ci.yml`. Both jobs (`Typecheck` and `Build`) run:

```yaml
- uses: actions/setup-node@v4
  with:
    node-version: 20
- run: npm ci          # ← this step
```

Node 20 ships **npm 10.x**, which is the same npm family that reproduced the `Missing: ws@8.21.3 from lock file` error for me. So this branch is very likely to fail CI at the install step, before typecheck or build even start — which would explain nothing about the change itself and look like a mystery failure.

That moves item 1 in my review from "please fix" to "must fix to land". Regenerating the lock under npm 10 should be enough; if you'd rather standardise the project on npm 11, that needs a matching change to the workflow (and to the documented Node floor, which is item 2).

Same run also confirms the `Build` job is `continue-on-error: true`, so a packaging regression would go amber rather than red — worth knowing when weighing electron 43 vs 42.

cc @chaitanyagiri

Maev4l:
Thanks — this is the most useful review I've had on a dependency PR, and running controls on `main` for anything that might be pre-existing is exactly the right method. Three commits pushed, one per actionable item.

| Commit | Item |
|---|---|
| `4d25ad9` | `fix(lock): record the optional ws peer so npm ci works on npm 10` |
| `4c2c97a` | `chore: declare the real Node floor — 22.22, not 18` |
| `d0e3bb3` | `fix(build): exclude node-pty's stale bin/ instead of declaring it single-arch` |

## 1. The `npm ci` break — fixed

I reproduced it before touching anything, rather than fixing a described problem:

```
$ npx npm@10 ci --dry-run --ignore-scripts
npm error code EUSAGE
npm error Missing: ws@8.21.3 from lock file
```

Your diagnosis is exactly right — `openai@7.5.0` declares an optional peer `ws: ^8.18.0`, npm 11 resolves it without recording an entry, npm 10 requires one.

**On which npm to standardise:** I regenerated under **npm 10** rather than pinning the project to npm 11, because npm 11 reads an npm-10 lock happily and not the reverse — so satisfying the older resolver satisfies both, and `ci.yml` needs no npm change. The regeneration adds **exactly one** entry (`@openai/agents-core/node_modules/ws@8.21.3`, `registry.npmjs.org`, integrity present, `optional` + `peer`): no removals, no version changes, same `lockfileVersion`.

`npm ci --dry-run` now exits 0 under **npm 10.9.9 and npm 11.17.0**, and `npm audit` still reports 0.

## 2. The Node floor — and a correction: it's 22.22, not 20.19

You flagged vite 7's `^20.19.0 || >=22.12.0`. Checking every direct dependency's `engines` instead of just vite's, the floor is **higher than that**, and it changes the conclusion:

```
@electron/rebuild 4                        >=22.12.0     ← excludes the entire Node 20 line
electron 43                                >= 22.12.0    ← same
better-sqlite3 13                          >=22
posthog-node                               ^20.20.0 || >=22.22.0
vite 7 / electron-vite 5 / plugin-react 5  ^20.19.0 || >=22.12.0
```

vite alone would still permit Node 20.19+, but `@electron/rebuild 4` and `electron 43` both require `>=22.12.0`, so **no Node 20 satisfies this tree**. On the 22 line `posthog-node` wants `>=22.22.0`. Hence `engines: { "node": ">=22.22.0" }`, and both docs now read "Node.js 22.22+".

**That has a consequence beyond docs, so please sanity-check this part:** I also moved `ci.yml`'s two jobs from `node-version: 20` to `22`. Leaving CI on a Node the package declares unsupported would emit `EBADENGINE` on every run, and `@electron/rebuild` runs in `postinstall` on that runner. It also happens to line up with item 1 — Node 22 ships npm 10, the exact resolver the lock is now generated for. Shout if you'd rather I drop the workflow change and leave that call to you.

I also synced `engines` into the lock's root entry, so the next `npm install` won't produce a surprise diff.

## 3. `node-pty/bin` vs `test_extension.node` — you're right, fixed

Fair hit: I described `bin/darwin-arm64-128` as stale and then declared it rather than excluding it. Excluding is correct — node-pty's `lib/utils.js` only ever reads `build/Release`, `build/Debug` and `prebuilds/<platform>-<arch>`, so nothing can load `bin/` at all. It's now excluded from `files`, and `x64ArchFiles` is narrowed to just the prebuilds. The rule now encodes load-bearing vs dead:

| | Path | Why |
|---|---|---|
| excluded | better-sqlite3 `test_extension.node` | test fixture, never loaded |
| excluded | node-pty `bin/**` | stale, unreachable by the loader |
| declared | node-pty `prebuilds/**` | real fallback when `build/` is absent |
| declared | better-sqlite3 `prebuilds/**` | 13.x loads from prebuilds exclusively |

Re-verified: `dist:mac` exit 0, `bin/` absent from the package, `build/Release/pty.node` and `spawn-helper` still `x86_64 arm64`, and in the packaged runtime through the real `app.asar` path better-sqlite3 round-trips a row and node-pty spawns a pty — so narrowing the rule broke neither loader.

## 4. The tunnelmole patch only reaching the hoisted copy

Keeping it as-is, but you're right that it's a silent failure mode. An npm `overrides` entry can't express "delete a self-reference" — the only shapes available still leave `tunnelmole` depending on some `tunnelmole` — so the postinstall script is the available mechanism, and it matches `patch-node-pty-conpty.cjs`. Happy to make it walk for nested copies instead of only the hoisted one if you'd like that hardened; it's a few lines.

## Still open — and both are yours, not mine

**Windows.** Genuinely blocked: macOS is all I have. The offer stands to pin **electron 42.3.4** instead, which clears every listed CVE with far less Chromium delta — say the word and I'll push that as a commit. Your addendum noting the `Build` job is `continue-on-error: true` cuts toward the lower-risk Electron if nobody can run `dist:win` soon.

**43 vs 42, and #168 vs this PR.** Both still need your call. On #168 I've pushed the lock-only cleanup you asked for (`22e7bdf`) — it's now 2 files, +7 / −224, and green under both npm versions, so it's ready whenever you want the packaging bug fixed independently of this PR.

One caveat on all of the above: **no CI run has actually happened on either PR** — Actions are dormant on my fork and fork PRs need your approval upstream, so `gh pr checks` reports nothing on either branch. The `npm ci` fix is verified locally under npm 10.9.9, matching your repro, but the first real green tick has to come from a maintainer-approved run.

chaitanyagiri:
Leaving this open, and apologising for the silence on it.

One commit from this branch shipped: `9bb060dd`, the node-gyp rebuild, which is shared with #168 and has been in the build since before 0.4.5.

The dependency migration itself did not ship. None of `822d2b47` (electron-builder 26), `edf1ba4f` (vite 7), `2e486c38` (electron 43 and better-sqlite3 13), `db000008` (dropping localtunnel), `4c2c97a4` (the real Node floor) or the rest are in v0.4.6.

So this is not a PR we can close as shipped. It is a real, substantial, unreviewed piece of work that has been open since 18 August, and the honest answer is that we have not reviewed it rather than that we decided against it.

Sorry for the wait.

PR #172, opened 2026-08-18
feat(hires): review batch imports in a FIFO queue

### LINKED ISSUE #148: Let me import several hires at once
### Problem / motivation

Right now "import hire…" takes one file at a time. Staffing an office with a handful of roles means repeating the whole dance for each one: open the Add-Agent modal, click import, find the file, review, spawn. Six roles from the Agent Gallery, six round-trips.

The deep links have a related snag. If you open several `munderdifflin://hire` links in a row, only the last one shows up — the earlier ones vanish with no error message, so it looks like the import worked when it didn't. (Happy to file that separately if you'd rather keep this issue to the feature alone.)

It's a small friction, but it lands on the one thing the app is about: running an office of agents rather than a single one.

### Proposed solution

Let the import dialog accept several files at once, then review them one by one.

Each manifest still gets its own "imported" banner and its own spawn click — plus a "skip" if I change my mind about one — with something like "hire 2 of 6" to show where I am. Anything invalid gets flagged and skipped rather than blocking the rest.

Nothing about the safety model needs to change: no auto-spawning, still one human decision per agent. The only thing that goes away is opening the file picker over and over.

### Alternatives considered

- **One file holding several hires.** Tidier if you want to share a whole team, but it needs a new file format, and it doesn't help when the manifests arrive as separate downloads. Could be a nice follow-up.
- **Approve them all in one click.** Deliberately not asking for this — reviewing each hire before it runs is the point, and one bulk "yes" is a worse place to make that decision than a few small ones.
- **Leave it as is.** Fine for two hires, tedious by six, which is what the Gallery ships today.

### Area

Onboarding / config

### Environment

v0.4.3, AppImage on Linux.

HaningZS:
## Summary

- accept multiple hire manifest files and validate each selection independently
- preserve every valid file and deep-link arrival in one FIFO review queue
- require an explicit spawn or skip for each hire, with `hire X of Y` progress and no auto-spawning

Fixes #148

## Test plan

- [x] `node --test test/hire-import.test.cjs test/hire-queue.test.cjs` — 10/10
- [x] `npm run test:focused` — 249/249
- [x] `node --test test/*.test.cjs` — 268/268
- [x] `npm run typecheck`
- [x] `npm run build`
- [x] Node 20/Linux container: targeted tests, typecheck, and production build
- [x] Real Electron flow: two deep links showed hire 1 of 2; skip advanced to hire 2 of 2 without spawning

## Additional validation note

The generic Debian Node 20 container ran 248/249 focused tests; the unrelated `hive-hook-node` Unix-socket case reproduced an `EPIPE` only inside that container. The same test passes in the full macOS run above, and the changed hire paths pass on both hosts.


HaningZS:
## UI validation

Real Electron run with two consecutive hire deep links. The first screenshot shows the FIFO head and progress; the second shows that an explicit skip advances exactly one hire without spawning.

munder-148-batch-hire
munder-148-batch-hire-after-skip

chaitanyagiri:
## TL;DR

Thanks @HaningZS — this is a well-built PR and one of the cleaner ones in the queue right now.

**What it does:** today "import hire…" accepts exactly one manifest file and a deep-link arrival overwrites whatever manifest was already waiting, so a second hire silently disappears. This PR replaces that single `pendingHire` slot with a small FIFO review queue: the picker accepts multiple files, each file is validated on its own so one bad JSON no longer discards its valid neighbours, the Add-Agent modal shows `hire X of Y`, and every hire still needs an explicit **spawn** or **skip**. Nothing auto-spawns. It also adds a `config:setAgentTokenCap` IPC that does a read-modify-write of `agentTokenCaps` inside the main process, which fixes a genuine data-loss race when several hires are approved back-to-back.

**Verdict:** no malicious code, no NSFW/profanity, no prompt-injection attempts. Typecheck is green and the full suite passes (271/271, +13 from this PR). **Two things to fix before merge** — a leftover queue after *cancel*, and the same token-cap race left un-fixed in the Command Center. Neither is large.

---

## What I did

Checked out `feat/batch-hire-imports-148` locally, read the full diff, traced every caller of the APIs it changed (`importHireFile` → `importHireFiles`, `pendingHire` → `hireQueue`, `agentTokenCaps` writers), and ran the type-checker and the whole test directory against both `main` and this branch so the numbers are comparable.

## The shape of the change

```
        deep link                 file picker (now multi-select)
   munderdifflin://hire            each file validated on its own
            │                          │            │
            │                     ok ──┘            └── bad → "Skipped 2 invalid
            │                     │                          files: 02-broken.json …"
            ▼                     ▼
      ┌───────────────────────────────────────┐
      │  hireQueue.pending  (FIFO, renderer)  │
      │  [ Jim ][ Pam ][ Dwight ]  reviewed:0 │
      └───────────────────────────────────────┘
                   │  head only
                   ▼
      ┌───────────────────────────────────────┐
      │  AddAgentModal — "hire 1 of 3"        │
      │  every field re-seeded on head change │
      └───────────────────────────────────────┘
             spawn ──┐        ┌── skip
                     ▼        ▼
              finishCurrentHire()  → next head, or close when drained
```

The queue helpers live in `src/shared/hireQueue.ts` as pure functions, which is why they're cheap to test — that file is the nicest part of the PR.

## What I tested

| Check | Result |
|---|---|
| `npm run typecheck` (node + web) | ✅ clean, exit 0 |
| `node --test test/*.test.cjs` on `main` (baseline) | ✅ 258/258 |
| `node --test test/*.test.cjs` on this branch | ✅ **271/271**, 0 fail |
| New tests specifically | ✅ `hire-queue` 4, `hire-import` 7, `agent-token-cap` 2 |
| `test:focused` script updated to include the 3 new files | ✅ yes |
| Malware / exfiltration / obfuscation scan of the diff | ✅ nothing — no network calls, no `eval`, no new deps, no post-install hooks |
| NSFW / profanity | ✅ none |
| Prompt-injection / instruction-smuggling in code, comments, tests or PR text | ✅ none found |
| Secret or path leakage | ✅ **improved** — batch errors are run through `basename()`, so a user's directory structure no longer lands in logs or screenshots. Nice touch. |
| Auto-spawn safety (the important one for hires) | ✅ still intact — `applyManifest` only fills form state, and `hireCommand()` rebuilds the command from the provider preset, so a manifest still cannot inject a spawn binary |

Your own reported numbers reproduce on my machine, so I'm treating the PR's test plan as accurate.

## Review — things to change

**1. Cancel leaves the rest of the batch in the store (please fix).** `onClose` in `App.tsx` only does `setAddAgentOpen(false)`; nothing drains `hireQueue`. So if I import five hires, review two, then hit **cancel**, the remaining three stay in the store. The next time I click "add agent" — days later, for an unrelated reason — the modal opens pre-filled with a third-party manifest and the 📋 hire banner, from a batch I thought I'd abandoned. It doesn't auto-spawn, so it isn't a security hole, but it's a surprising thing to hand a user a blank-form expectation and give them someone else's command instead. It also makes progress lie: `reviewed` only resets when `pending` empties, so an abandoned batch plus a later deep link renders as "hire 2 of 4". Simplest fix is to clear the queue on cancel (or on modal unmount) and let a fresh import start a fresh run.

**2. The token-cap race is only half fixed.** The new `setAgentTokenCap` IPC is the right call and the comment explaining *why* is excellent. But `CommandCenterPanel.tsx:583-586` still does the exact pattern you just removed — `const next = { ...agentTokenCaps }` from a snapshot loaded once at mount, then `updateConfig({ agentTokenCaps: next })`. So a cap typed in the Command Center can still clobber caps written by a hire batch (and vice-versa). Since the safe IPC now exists, pointing that one call site at it is a two-line change and closes the bug properly.

**3. Minor, non-blocking.** The `tokenCap <= 1e10` bound in `config.ts` duplicates the identical bound in `src/shared/hire.ts:261`. Worth exporting one constant so the two can't drift. And `advanceHireReview()` calls `applyManifest(next)` while the `useEffect` on `[pendingHire]` will apply it again on the next render — harmless because it's idempotent, but one of the two is dead weight and the next reader will wonder which.

Everything else — immutable queue helpers, the independent-validation loop, re-seeding *every* field (including `resumeSessionId` and `folderNote`) so one hire's edits can't leak into the next, awaiting the cap write before advancing — is the right call and matches house style.

## Questions

@HaningZS — (1) is the queue surviving **cancel** deliberate, i.e. "resume this batch later", or an oversight? If deliberate it needs a visible affordance, because right now nothing tells the user hires are still parked. (2) The `hire:openFile` IPC return shape changed from `{ok, manifest}` to `{ok, manifests, errors}`; I found no other callers, but can you confirm nothing outside `AddAgentModal` consumed the old shape?

@chaitanyagiri — your call on item 2: do you want the Command Center switched over to `setAgentTokenCap` **in this PR** (keeps the fix whole, slightly widens scope), or tracked as a follow-up issue so this one stays focused on #148?

HaningZS:
Thanks for the detailed review — both blocking items are addressed in `80b9867`.

- Queue persistence after cancel was an oversight, not an intended resume-later behavior. Closing the modal now clears the unfinished review run, and a later import starts again at 1/N.
- Command Center token-cap edits now use the same main-process atomic per-agent IPC. The IPC supports both set and clear, so it no longer replaces the full map from a stale renderer snapshot.
- I also centralized the `1e10` cap constant and removed the duplicate manifest application.

I confirmed there are no other in-repo consumers of the old `hire:openFile` return shape.

Validation:
- targeted regression tests: 17/17
- focused suite: 256/256
- full suite: 275/275
- `npm run typecheck`
- `npm run build`
- `git diff --check`

All passing.

chaitanyagiri:
## TL;DR

Thanks @HaningZS. This was one of the cleaner PRs in the queue when it was first reviewed and that has not changed. The FIFO review queue is well built, `src/shared/hireQueue.ts` as pure functions is the right shape, and nothing auto spawns, which is the property that matters most here.

**Verdict: needs work, and it is now mostly a rebase.** The PR has gone CONFLICTING since it was last looked at, and there are two open items from that earlier review that I could not confirm you have addressed.

---

## What I did

Static review of the current diff against `origin/main` @ `eb2818a8`, plus a read of the earlier review of this PR so I did not repeat it back at you. **I did not check out the branch and I did not run the tests**, so I cannot confirm the earlier 271/271 result still holds on your current head.

## The two open items from the earlier review

I am restating these rather than re deriving them, and if you have already fixed either, just say so and I will drop it.

1. **Cancel leaves the hire queue populated in the store.** Dismissing the modal partway through a batch should drain or explicitly retain the queue, and right now it is ambiguous which one happens. The user visible symptom is a stale `hire 2 of 3` state that reappears later.

2. **`CommandCenterPanel:583` still has the token cap race** that this PR fixes elsewhere. You added `config:setAgentTokenCap` with a read modify write inside main, which is the correct fix, but the Command Center still writes the old way. Same bug, different call site.

The second one is the one I would prioritise, because it is the data loss shaped one.

## What is new since that review

**It is CONFLICTING.** `package.json` is part of that, and it is worth knowing why: this PR is one of five currently open that rewrite the single line `test:focused` script from different baselines. Whichever lands first, the rest conflict, and a careless resolution silently drops test files from the list.

That is not your fault and it is on the maintainer to fix properly. But when you rebase, **please check the `test:focused` line specifically** rather than accepting either side wholesale, because that is the hunk where a bad resolution does invisible damage.

1. Rebase, with care on the `test:focused` line.
2. Drain or explicitly retain the hire queue on cancel.
3. Fix or explicitly defer `CommandCenterPanel:583`. If you would rather scope it out, say so in the description and I will not hold the PR for it.

The core of this is good work and I would like it in.

HaningZS:
Thanks for taking another look. There appears to have been a timing overlap with the v0.4.5 integration.

All three PR commits are already represented in `main`:

- the original #172 work was integrated through `3dbe316` (`Merge branch 'pr-172' into integration/v0.4.5`);
- the final review fix was cherry-picked as `e1b7159`, whose commit message explicitly records `cherry picked from commit 80b9867`.

I also verified the two open items directly in the `eb2818a` tree used for this review:

- `App.tsx` calls `clearPendingHires()` before closing the modal;
- `CommandCenterPanel.tsx` uses `window.cth.setAgentTokenCap(id, tokens)` for atomic set/clear.

The `test:focused` conflict has also already been resolved on main by changing it to `node --test test/*.test.cjs`.

`git cherry upstream/main HEAD` reports `- 80b9867`, confirming that the remaining PR-head patch is already applied upstream. Rebasing now would therefore leave the PR empty rather than introduce any additional fix.

Could you mark/close #172 as integrated, or let me know if you would prefer me to close it from my side?

chaitanyagiri:
Leaving this open, and apologising for the wait, because both things are true here.

Two of your three commits shipped and have been in the build since before 0.4.5: `7b187d2e`, the FIFO queue for reviewing batch imports, and `d41eee90`, preserving token caps across batch reviews. This PR has been open the whole time they have been running in people's hands, which is on us.

The third, `80b9867b`, "clear cancelled batches and update caps atomically", did not ship and is not in v0.4.6 by any other route. I checked. That is a correctness fix touching `src/main/config.ts`, `src/preload/index.ts` and the modal, so closing this PR would bury it.

That is why this stays open rather than being closed as shipped.

Thank you, and sorry it took this long to say anything.

HaningZS:
Thanks for checking, and for keeping this open rather than risking burying the fix. I re-verified against the release tag itself, and I think there may be a provenance mismatch:

- GitHub's `refs/tags/v0.4.6` annotated tag peels to commit `64bd64df0e8d315a6e895283f776b81f84eef2cc`.
- `git merge-base --is-ancestor e1b7159 v0.4.6` exits `0`, so [`e1b7159`](https://github.com/chaitanyagiri/munder-difflin/commit/e1b715979fccef048754e4152753b6c724892325) is in the v0.4.6 history.
- `80b9867` and `e1b7159` have the same stable patch-id: `d6afff4264e715bc056f3467d74152cefaaa8c8a`.
- `e1b7159` records `(cherry picked from commit 80b9867...)` in its commit body.
- `git tag --contains e1b7159` lists both `v0.4.5` and `v0.4.6`.

The source at the v0.4.6 tag also contains all three parts of the fix: `clearPendingHires()` in `App.tsx`, atomic `setAgentTokenCap(id, tokens)` in `CommandCenterPanel.tsx`, and `src/shared/tokenCaps.ts`.

So the Git history and tagged source behind [v0.4.6](https://github.com/chaitanyagiri/munder-difflin/releases/tag/v0.4.6) both include the exact `80b9867` patch. Unless the published binaries were built from something other than the v0.4.6 tag commit above, it appears the third fix shipped too.

Could you check the build provenance once more? If the artifacts do come from `64bd64df`, I think #172 can be closed as integrated. If they came from a different source commit, please let me know and I am happy to update the PR accordingly.

PR #173, opened 2026-08-19
Add opt-in web UI: browser access to the same interface over HTTP

### LINKED ISSUE #167: An option to host it locally and access it over the network
### Problem / motivation

I have a server(an old laptop) and I want to have this hosted over there so that I can access it over other devices on browser instead of using RDC everytime.

### Proposed solution

Maybe open a web UI on a port for the same interface we have on electron desktop app

### Alternatives considered

_No response_

### Area

Design system

AmIrRX0:
Serve the office to any browser on the network so the app can live on an always-on machine (an old laptop as a server) without remote desktop.

No second UI is introduced: the built renderer and the built preload bundle are served byte-for-byte. A small browser boot script stands in for the preload's `electron` imports — `ipcRenderer.invoke` becomes `POST /__webui/invoke`, `ipcRenderer.on` is fed by one Server-Sent Events stream mirroring every `webContents.send`, and the two `sendSync` channels use synchronous XHR. An `ipcMain` capture installed before any handler registers routes bridged calls through the same handlers the desktop renderer uses.

The feature is off by default. Enable it with `webUi.enabled` or `MD_WEBUI=1`; `MD_WEBUI_PORT`, `MD_WEBUI_HOST`, and `MD_WEBUI_TOKEN` provide overrides. Non-loopback binds require a bearer token, exchanged for an HttpOnly cookie on first visit.

Covered by `test/webui.test.cjs` (capture, wire serialization including binary payloads, authentication, traversal safety, and SSE fan-out) and verified end-to-end in headless Chromium against the real built bundles.

Closes #167.

Claude-Session: https://claude.ai/code/session_01Uk1mPQbF8kNmsSnYWYZ25e

## What & why

This adds an opt-in hosted version of the existing desktop interface, so Munder Difflin can run on an always-on machine and be accessed securely from another device on the local network.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before
Before — web UI test absent on upstream main

### After
After — web UI tests pass

## How I tested it

- `node --test test/webui.test.cjs` — 9/9 passed
- `npm run typecheck` — passed
- `npm run test:focused` — 561/561 passed
- `npm run build` — passed

## Checklist

- [x] No new art or third-party visual assets were added.


chaitanyagiri:
## TL;DR

Thanks @AmIrRX0 — and apologies for the wait; this arrived just after a review sweep and slipped the queue. It's a well-engineered PR with a genuinely clever central idea, and the security work is more careful than most contributions of this kind. My reservation is one default, not the design.

**Context:** the app is an Electron desktop app. Its UI talks to the main process over Electron IPC through a preload bridge. This PR lets that same UI be served to a browser over HTTP, so the app can live on an always-on machine — an old laptop as a server — and be used from a phone without remote desktop.

**The clever part:** no second UI is introduced. The *built* renderer and the *built* preload are served byte-for-byte, and a small boot script stands in for the preload's `electron` import — `ipcRenderer.invoke` becomes `POST /__webui/invoke`, `ipcRenderer.on` is fed by one Server-Sent-Events stream mirroring every `webContents.send`, and the two `sendSync` channels ride a synchronous XHR. An `ipcMain` capture installed before any handler registers routes browser calls through **the exact same handlers** the desktop renderer uses. So the two surfaces cannot drift.

**Verdict:** the engineering is sound and the tests are real. Typecheck clean, 267/267 (9 new), merges cleanly, 6 commits behind `main`. **But the default bind is `0.0.0.0` over plain HTTP**, and I don't think that should be the default. Details below.

---

## What I did

Read the whole diff with an eye on the network surface first, checked the auth path, the traversal guard and the token generation line by line, checked out the branch, ran the type-checker and the full suite, ran the new `test/webui.test.cjs`, and tested the merge.

## The architecture, because it's worth understanding before the security part

```
   BROWSER                                    MAIN PROCESS
   ────────                                   ────────────
   the SHIPPED renderer bundle
        │  window.cth (from the SHIPPED preload)
        ▼
   boot.js shim
        ├─ ipcRenderer.invoke ──► POST /__webui/invoke ──┐
        ├─ ipcRenderer.sendSync ─► POST /__webui/sendSync┤
        └─ ipcRenderer.on   ◄──── SSE /__webui/events ◄──┤
                                                          ▼
                                              the SAME ipcMain handlers
                                              the desktop renderer uses
```

That last line is the design's whole justification. A second web UI would rot; a bridge that reuses the shipped bundles and the real handlers cannot. Native dialog channels return a cancel-shaped error rather than hanging on a picker no remote user can see, and clipboard calls reroute to the browser's own clipboard — both are the right calls.

## The security work I checked, and it holds up

| Check | Finding |
|---|---|
| Token comparison | ✅ `timingSafeEqual` from `node:crypto` — constant-time, not `===` |
| Token generation | ✅ `randomBytes(16)` = 128 bits, persisted to config when unset |
| Token required beyond loopback | ✅ auto-generated if the user enabled the server without setting one |
| Cookie | ✅ `HttpOnly; SameSite=Lax; Path=/` |
| Path traversal | ✅ `normalize(join(root, rel))` then `abs !== root && !abs.startsWith(root + sep)` — correct, and the `+ sep` matters |
| Unauthenticated routes | ✅ `/__webui/health` only |
| Off by default | ✅ `webUi.enabled` / `MD_WEBUI=1` |

That is a better list than I expected. Most PRs that add an HTTP server to a desktop app get at least one of those wrong, usually the timing-safe comparison or the traversal guard.

## What I tested

| Check | Result |
|---|---|
| `node --test test/*.test.cjs` | ✅ 267/267 (258 on `main` + your 9) |
| `test/webui.test.cjs` alone — capture, wire serialization incl. binary, auth, traversal, SSE fan-out | ✅ passes |
| Merge against current `main` | ✅ clean, 6 commits behind |

## My one substantive objection: the default bind

```ts
const host = (env.MD_WEBUI_HOST ?? cfg?.host ?? '0.0.0.0').trim() || '0.0.0.0';
```

When a user turns this on, the default is to listen on **every interface**, over **plain HTTP**. Your token gate is real and I'm not disputing it — but consider what the default actually produces:

- the token crosses the LAN in cleartext on every request, so anyone on the same coffee-shop or office Wi-Fi can read it off the wire and then has full control of the harness — which spawns local processes and reads/writes the user's registered directories
- the first visit is `http://host:4820/?token=…`, so the secret lands in browser history, and in the logs of anything between
- the cookie has no `Secure` flag, which is correct over HTTP and is exactly the point — there is no transport to protect it

The people most likely to enable this are the ones your description targets: "an old laptop, a home server". Those are the least likely to have a reverse proxy in front. Your own docs already say the right thing —

> *bind to `127.0.0.1` in that setup and let the proxy own the network edge*

— but that's advice, and the default is the opposite of the advice.

**I'd flip the default to `127.0.0.1`** and make LAN exposure an explicit opt-in (`MD_WEBUI_HOST=0.0.0.0`). Someone who genuinely wants network access will read one line of docs; someone who just wanted to try it in a browser gets a local-only server. That's the direction the failure should point. It's a one-word change plus a docs edit, and I don't think it costs you the use case at all.

Not blocking on my say-so — this is @chaitanyagiri's call, and it's a product decision about default posture rather than a bug.

## Two smaller notes

1. **Synchronous XHR.** You've flagged it in a comment and justified it (two one-shot channels, both with try/catch fallbacks), which is the right way to land something like this. It's deprecated and blocks the main thread, but the alternative is patching two call sites in a bundle you're deliberately not modifying. I'd accept it with the comment; worth an issue to revisit if either channel ever becomes hot.
2. **`package.json +1/-1`** — you rewrote the `test:focused` line. You're one of **nine** open PRs doing that from different baselines, so all nine conflict with each other. Entirely not your fault, the line shouldn't exist in that form, and it's being fixed centrally. No action needed.

## Questions

@AmIrRX0 — (1) any objection to `127.0.0.1` as the default? (2) You verified end-to-end in headless Chromium against the real built bundles, which is great — did you also test with a **second** browser connected, so two SSE clients are fanned out to at once? The fan-out path is tested at unit level but concurrent real clients are where that usually breaks. (3) What happens to a PTY's output stream when a browser tab is closed without unsubscribing — does the SSE connection get cleaned up, or does it accumulate?

@chaitanyagiri — the default-bind decision is yours and it's the only thing between this and merge-ready. Everything else here is solid.

AmIrRX0:
Thanks for the thorough review!

**(1) Default bind** — no objection at all; you're right that the failure should point
toward "can't reach it from my phone yet". Flipped in 9867ce9: default is now `127.0.0.1`,
and LAN exposure is an explicit opt-in (`webUi.host: "0.0.0.0"` / `MD_WEBUI_HOST=0.0.0.0`),
still token-gated. Docs restructured to lead with the loopback default and frame the
TLS-reverse-proxy guidance around it.

**(2) Two concurrent SSE clients** — now verified with real browsers, not just the unit
fan-out: two headless-Chromium tabs against the built bundles, `clientCount = 2`, one
broadcast, and both tabs received the event through the shipped preload's `onPtyData`.

**(3) Tab closed without unsubscribing** — the browser closing the tab tears down the
EventSource's TCP connection, Node fires `req.on('close')`, and the client is removed from
the fan-out set. Verified in the same run: closing tab 2 with no `removeListener` call
dropped the server's client count 2 → 1, and the surviving tab kept receiving events —
no accumulation. Two backstops cover uglier deaths: a destroyed/ended response is dropped
on the next write, and a client whose socket buffer falls more than 8 MB behind (PTY
firehose, stalled reader) is disconnected rather than buffered without bound — it
reconnects via SSE's built-in retry. The PTY stream itself is unaffected either way; the
desktop window's delivery path doesn't run through the fan-out set.

Sync-XHR issue to revisit later: happy to open one if you'd like it tracked.

chaitanyagiri:
Cross-linking this so it is findable: GitHub currently reports no closing issue reference on this PR, so it does not surface against the open request it answers.

The body says "Closes the web-UI / hosting feature request" in prose, without a number. The matching open issue is #167, "An option to host it locally and access it over the network".

For the record, the `Closes #123` in the raw body is the worked example from `.github/PULL_REQUEST_TEMPLATE.md`, left inside its HTML comment. GitHub does not parse closing keywords inside comments, so nothing was mis-linked here.

Adding `Closes #167` to the description would connect the two.

PR #175, opened 2026-08-19
fix(worker-wake): main-process watchdog wakes idle workers on undrained inbox

### LINKED ISSUE #151: Worker agents can permanently stall on a delivered inbox message — worker wake is renderer-only with no main-process retry

rekcilyssup:
Closes #151 (fix A — the main-process retry; fix B, the renderer dedup, is already on main).

## Problem
The renderer's idle inbox-wake nudge (`useHive.ts` effect #3) is the ONLY path that wakes a worker parked on an undrained inbox — and it lives on a `setInterval` in the renderer. A throttled or occluded window stops honoring those timers, so the moment mail lands can be missed forever: the worker sits quiet at its prompt with an undrained `inbox/` and never acts (god is unaffected — the main process re-engages it on its heartbeat cadence).

## Fix
A renderer-independent watchdog in the main process: a 15s always-on beat (armed in `armAlwaysOnBeats`, so power-resume re-arms and drains the backlog with the other beats) that finds live workers that are genuinely idle with undrained inbox mail and types the same guarded nudge the renderer would have, directly into the PTY.

New `src/main/workerWake.ts` (`WorkerWakeWatchdog` — electron-free, unit-testable, mirroring `ControlRegistry`):

- **Idle-only**: no PTY output for 12s (`QUIESCE_IDLE_MS`) — a mid-turn worker is never interrupted
- **Boot grace**: 35s from spawn (`BOOT_GRACE_MS`) — the orientation prompt is never mistaken for idle
- **Not paused/halted/autoDeliveryPaused** (`ControlRegistry.snapshot`)
- **No pending HITL**: a permission/approve notification in the hook stream (new optional `HookServer` observer) blocks nudges for 5 minutes
- **Cooldown**: per-worker 60s so the watchdog never stacks on top of the renderer's own nudge
- **God excluded** (its heartbeat path already re-engages it)
- Delivery uses the renderer's exact nudge text + `submitToPty`'s text-then-Enter pattern

## Verification
- 14 new unit tests (`test/worker-wake.test.cjs`) covering the full guard matrix
- Full suite: 272/272 passing, `npm run typecheck` + `npm run build` green

### DIFF
--- a/src/main/hooks.ts
+++ b/src/main/hooks.ts
@@ -67,7 +67,11 @@ export class HookServer {
     private control?: ControlRegistry,
     /** Circuit breaker (Lane A #6.6b) — fed the hook-derived signals (session id,
      *  repeated identical tool calls). Optional so the server still runs without it. */
-    private breaker?: CircuitBreaker
+    private breaker?: CircuitBreaker,
+    /** Optional observer of every hook boundary (agentId, event, message). The
+     *  worker inbox-wake watchdog (workerWake.ts) feeds on this to learn when an
+     *  agent is parked on a permission/HITL prompt so it never types into it. */
+    private onEvent?: (agentId: string | undefined, event: string, message: string | undefined) => void
   ) {}
 
   start(): void {
@@ -115,6 +119,7 @@ export class HookServer {
   private handle(p: HookPayload): unknown {
     const agentId = p.agent_id ?? undefined;
     const event = p.hook_event_name ?? 'Unknown';
+    this.onEvent?.(agentId, event, p.message);
     if (agentId && typeof p.transcript_path === 'string' && p.transcript_path) {
       this.transcriptPaths.set(agentId, p.transcript_path);
     }
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -59,6 +59,7 @@ import * as integrations from './integrations';
 import { validateBaseUrl, buildAuthHeaders, resolveUpstreamUrl, secretRefFor, INTEGRATION_TEMPLATES } from '../shared/integrations';
 import { RosterStore } from './roster';
 import { ControlRegistry } from './control';
+import { WorkerWakeWatchdog, WORKER_WAKE_NUDGE, type WorkerWakeFacts } from './workerWake';
 import { fetchHireManifest, readHireManifestFile } from './hire';
 import { parseHireDeepLink, type HireManifest } from '../shared/hire';
 import { ClosingTimeController } from './closingTime';
@@ -262,9 +263,15 @@ let breakerBeatTimer: ReturnType<typeof setInterval> | null = null;
 // Feed the breaker's api_error-storm trip from Oscar's OTel api_error spans —
 // Jim's one breaker input with no on-branch source (telemetry.onApiError seam).
 telemetry.onApiError((agentId) => breaker.recordError(agentId));
+// Worker inbox-wake watchdog (#151): finds idle workers with undrained inbox mail
+// and types the same guarded nudge the renderer would have (so a throttled
+// background window can't leave a worker parked on an unread inbox forever).
+// HookServer feeds it the hook stream so a permission/HITL prompt blocks nudges.
+const workerWake = new WorkerWakeWatchdog();
 // HookServer needs BOTH: Oscar's control registry (HITL pause/gate/steer/halt via
 // hook returns) AND Jim's breaker (feed recordToolUse on each PostToolUse).
-const hookServer = new HookServer(hive, () => liveWebContents(), () => readConfig(), control, breaker);
+const hookServer = new HookServer(hive, () => liveWebContents(), () => readConfig(), control, breaker,
+  (agentId, event, message) => workerWake.noteHook(agentId, event, message));
 const memory = new MemoryManager(
   () => readConfig().harnessHome,
   () => { const c = readConfig(); return { enabled: c.semanticMemory !== false, model: c.embeddingModel ?? 'minilm' }; }
@@ -398,6 +405,8 @@ function teardownPty(id: string): void {
   const agentId = ptyToAgent.get(id);
   if (agentId) {
     ptyToAgent.delete(id);
+    // Drop watchdog state so a dead agent can't get nudged or leak its grace.
+    try { workerWake.forget(agentId, id); } catch { /* best-effort */ }
     // Drop breaker state so a dead agent can't leak/zombie a tripped level.
     try { breaker.forget(agentId); } catch { /* best-effort */ }
     // W1 — kill this agent's proxy-bridge sidecar (qwen), if any, so a dead
@@ -2681,7 +2690,12 @@ async function spawnAgentCore(opts: AgentSpawnOptions, owner: Electron.WebConten
   }
   // Remember which agent owns this PTY so closing the tab can archive it. A
   // live terminal means active — ensureAgent above already cleared `archived`.
-  if (opts.hive?.id) ptyToAgent.set(opts.id, opts.hive.id);
+  if (opts.hive?.id) {
+    ptyToAgent.set(opts.id, opts.hive.id);
+    // Worker inbox-wake watchdog (#151): boot grace starts at spawn so the
+    // initial orientation prompt is never mistaken for an idle agent.
+    workerWake.noteSpawn(opts.id);
+  }
   // Pre-accept Claude Code's bypass-mode warning + folder-trust dialog so the
   // agent (spawned with --permission-mode bypassPermissions) doesn't stall on an
   // interactive prompt it can't answer and exit code 1. Best-effort, never blocks.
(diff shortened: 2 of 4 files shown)


chaitanyagiri:
## TL;DR

Thanks @rekcilyssup — this is careful, well-guarded work, and the design reasoning in your description is genuinely good. I have to lead with the awkward part, though: **#162 is the same fix, by a different author, closing the same issue.** That's a coordination failure on our side, not yours.

**Context:** an agent with unread hive mail gets woken by a nudge typed into its terminal. Today the *only* thing that does that for a worker lives in the renderer, on a `setInterval`. Browsers and Electron throttle timers in occluded or backgrounded windows, so the moment mail lands can be missed indefinitely — the worker sits at its prompt with an undrained `inbox/` forever. (God is unaffected; the main process re-engages it on its own heartbeat.)

**Your fix:** a renderer-independent watchdog in the main process — a 15s always-on beat that finds live, genuinely-idle workers with undrained mail and types the nudge directly into the PTY. New `src/main/workerWake.ts`, electron-free and unit-testable, mirroring `ControlRegistry`.

**Verdict on the code:** correct, cautious, and better-guarded than I expected. Typecheck clean, 272/272 (14 new), merges cleanly, 4 commits behind `main`. The blocking question is which of the two PRs we take.

---

## What I did

Read the diff, read `#162` side by side with it, checked out your branch, ran the type-checker and the full suite, ran your new tests, and tested the merge against current `main`. I also re-read the renderer's effect #3 on `main` to confirm the timer-throttling premise holds.

## Your guard set is the strongest part

A watchdog that types into a terminal is a genuinely dangerous thing to add — get it wrong and it interrupts an agent mid-thought, or races the human. You gated it four ways, and each one is the right gate:

```
   every 15s beat
        │
        ├─ no PTY output for 12s?          ── no ──► skip   (mid-turn worker untouched)
        ├─ ≥35s since spawn?               ── no ──► skip   (boot prompt isn't idleness)
        ├─ not paused/halted/delivery-off? ── no ──► skip   (operator intent respected)
        ├─ no pending HITL notification?   ── no ──► skip   (never answer a permission prompt)
        └─ per-worker cooldown elapsed?    ── no ──► skip   (watchdog + renderer can't double-fire)
                │
                ▼  type the nudge
```

The HITL gate is the one I'd have asked for if it were missing. A worker sitting on a permission prompt *looks* idle by every output-based measure, and typing a nudge into that prompt would answer it with garbage. Wiring a new optional `HookServer` observer to detect it is the right mechanism.

Arming it in `armAlwaysOnBeats` so power-resume re-arms and drains the backlog is a nice touch that suggests you'd actually thought about laptops sleeping.

## What I tested

| Check | Result |
|---|---|
| `node --test test/*.test.cjs` | ✅ 272/272 (258 on `main` + your 14) |
| `test/worker-wake.test.cjs` alone | ✅ passes |
| `workerWake.ts` is electron-free (unit-testable in isolation) | ✅ confirmed |
| Merge against current `main` | ✅ clean, 4 commits behind |

## The duplicate, stated plainly

**#162** (@gpechieu) is a main-process inbox-wake watchdog, closes #151, adds `src/main/inboxWake.ts`, 7 files, 109 lines of tests. **#175** (yours) is a main-process worker-wake watchdog, closes #151, adds `src/main/workerWake.ts`, 4 files, 151 lines of tests. Same problem, same diagnosis, same architectural answer, written independently within days of each other.

Neither of you did anything wrong. #151 wasn't assigned, and nothing in the repo asks contributors to claim an issue before starting — that's a gap in our contribution docs and it has now cost two people a weekend. I've raised it with the maintainer as a process fix, and I'm sorry it landed on you.

Where they differ, honestly:

| | #162 | #175 (this PR) |
|---|---|---|
| Idle signal | derived from real **hook boundaries** (`hookIdleFor`) | **PTY output quiet** for 12s |
| HITL protection | via hook state — `hookIdle` is 0 while a prompt is pending | explicit `HookServer` observer |
| Scope | all agents | workers specifically |
| Preload surface | adds 10 lines | none |
| Tests | 109 lines | 151 lines |

The substantive difference is the idle signal. Hook boundaries are the more *truthful* source — they know what the CLI is actually doing rather than inferring it from silence. But they only exist for hook-bearing providers, and I flagged a gap on #162: a hook-bearing agent that hangs **mid-turn** reports `hookIdle` 0 forever, so that watchdog can never fire for exactly the wedged agent it's meant to rescue. Your output-quiet signal doesn't have that hole — 12s of silence is 12s of silence regardless of why.

So my read is that **your idle signal is more robust and #162's is more precise**, and the best outcome is one watchdog that uses hook state when it's available and falls back to output-quiet when it isn't. That's a merge of the two, not a winner.

@chaitanyagiri — this needs your call, and it should be made before either author rebases. My suggestion: take #162 as the base (it's further along and touches the preload surface already), and ask for this PR's output-quiet fallback and explicit HITL observer to be folded in. But you may reasonably prefer the simpler one.

## One thing to fix either way

`WORKER_WAKE_NUDGE` is a copy of the renderer's nudge string, and **#187 is about to make that text variable** — it adds message ids to the nudge and replaces string-equality dedupe with an `isInboxNudge()` predicate in `src/shared/hiveNudge.ts`. If #187 lands first, your constant will never match a renderer-queued nudge and both can fire for the same mail. Whichever watchdog we take should import that predicate rather than compare against its own literal. Flagged on #187 and #162 too.

## Questions

@rekcilyssup — (1) how do you feel about folding into #162 rather than competing? I'd rather you got credit for the parts of yours that are better than lose the work entirely, and I'll argue for that either way. (2) Did you hit this live — a worker actually stuck on an undrained inbox — or find it by reading `useHive`? A real occurrence helps decide how urgently this ships. (3) Was 12s / 35s / 15s tuned against something, or picked to match the renderer's existing constants?

rekcilyssup:
Thanks @chaitanyagiri — appreciate the careful read and the even-handed triage here. Happy to answer the three questions directly.

**1. Folding into #162 — yes, agreed.** #162 as base, with this PR's output-quiet fallback and the explicit HITL observer folded in. I'd rather see the right watchdog land than win a race, and credit where it's due to @gpechieu for the hook-boundary idle signal. I'm happy to port those pieces over myself if it helps — the two PRs are small enough that a hand-merge beats a diff-y rebase.

**2. Found it by reading useHive, not live.** I reasoned from effect #3's `setInterval` living in the renderer plus Chromium throttling occluded/backgrounded window timers — I never actually observed a worker stuck on an undrained inbox. So I'd echo your urgency read: this is a latent correctness hole rather than an active outage, the kind that gets loud on a laptop that sleeps. Given that, folding rather than rushing to land seems like the right cost.

**3. Tuning.** The 12s / 35s / 15s are not new numbers — they mirror the renderer's constants: `QUIESCE_IDLE_MS = 12000` and `BOOT_GRACE_MS = 35000` in `useHive.ts`, and the 15s beat matches the renderer's `setInterval(poll, 15000)`. The only new values are the 60s per-worker cooldown and the 5-minute HITL rearm, both chosen conservatively: the cooldown comfortably exceeds the renderer's poll interval so the watchdog and renderer can't double-fire on the same mail, and the HITL rearm is deliberately long since a permission prompt the human is deciding on is the one thing we must never type into.

And agreed on the #187 note — `WORKER_WAKE_NUDGE` comparing against its own literal is exactly the fragility you flagged. Whichever watchdog survives should import `isInboxNudge()` from `src/shared/hiveNudge.ts` rather than duplicate the string; I'll make sure that's in whatever lands.

I'll hold the rebase until you and @chaitanyagiri pick a base. Thanks again.

Issue #104, opened 2026-07-02
Add support for new CLI agents


chaitanyagiri:
### Problem / motivation

Currently we support: 
1. Claude Code
2. Codex
3. AntiGravity CLI
4. OpenCode
5. Crush
6. Pi
7. Qwen Code
8. [WIP] Github Copilot CLI

If you find any bug in existing integrations or if you find any new cli agent that should be available in munder difflin then that's a really good first issue as you get to understand about the core tunnelling and bridging capabilities.

### Proposed solution

A well tested integration of the CLI integrated/fixed. Here's a checklist (do not limit yourself to these):
1. Orchestrator should be able to use the CLI agent with any model autonomous mode.
2. New Agent should be able to use the CLI agent with any model autonomous mode.
3. Orchestrator and agent of different types should work together(messages, boards, tasks)
4. The bridge is what tells us about what's happening in the CLI so our queues work. Make sure the bridge works

### Area

Terminals / PTY

swarnendu19:
hey i noticed the repository already contains a Kimi Code provider preset, but it has no lifecycle bridge and therefore cannot receive hive inbox messages... Would a PR completing the Kimi hook bridge, session resume, protocol seeding, and installer support be welcome under this issue?

PR #176, opened 2026-08-19
feat: add official Gemini CLI provider

FenjuFu:
## Summary

- register the official Google Gemini CLI as a first-class provider
- support model selection, current autonomous mode, interactive protocol seeding, session resume, and installer metadata
- install a per-agent Gemini system settings file instead of mutating the user's global settings
- bridge SessionStart, Before/After Agent, and Before/After Tool hooks into the shared hive lifecycle
- add Gemini model/UI entries and regression coverage for provider config, context automation, installer metadata, and hook setup

## Why

Gemini CLI exposes the lifecycle hooks needed for safe idle detection, operator control, and hive message delivery, so it can participate as an orchestrator or worker rather than acting as a terminal-only preset.

This contributes to #104 without overlapping the Kimi bridge work discussed there.

Implementation follows Gemini CLI's official [CLI reference](https://github.com/google-gemini/gemini-cli/blob/main/docs/cli/cli-reference.md), [hooks reference](https://github.com/google-gemini/gemini-cli/blob/main/docs/hooks/reference.md), and [configuration layers](https://github.com/google-gemini/gemini-cli/blob/main/docs/reference/configuration.md).

## Validation

- npm run typecheck
- npm run build
- node --test test/agent-provider.test.cjs test/provider-automation.test.cjs test/provider-config.test.cjs (20 passed)
- hook configuration assertions pass locally; the existing Windows-only temp cleanup in hive-hook-node.test.cjs reports EPERM after assertions, while Linux CI exercises the full cleanup path

### DIFF
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -711,6 +711,11 @@ export class HiveManager {
               // idle inbox-wake nudge is the guaranteed drain fallback.
               env.OPENCODE_CONFIG_DIR = this.installOpenCodePlugin(dir);
             }
+            else if (desc.shim === 'gemini') {
+              // Point only this worker at a per-agent system settings file so
+              // the bridge is trusted and ~/.gemini/settings.json stays untouched.
+              env.GEMINI_CLI_SYSTEM_SETTINGS_PATH = this.installGeminiHooks(dir);
+            }
             else if (desc.shim === 'grok') this.installGrokHooks();
           } else if (desc.kind === 'proxy') {
             // Stable per-spawn session id, stamped on every synthesized payload so
@@ -1569,6 +1574,45 @@ export class HiveManager {
         writeFileSync(p, JSON.stringify(existing, null, 2), 'utf8');
       } catch { /* best-effort per file */ }
     }
+  }
+
+  /** Official Google Gemini CLI lifecycle bridge. Gemini's hook payload is
+   *  already snake_case; the shim maps event names into HookServer's common
+   *  vocabulary and translates deny/steering replies back to Gemini.
+   *
+   *  The system settings path is per agent. Gemini merges object and array
+   *  settings across layers, so auth and user settings remain in their normal
+   *  GEMINI_CLI_HOME while this trusted bridge stays isolated. */
+  private installGeminiHooks(dir: string): string {
+    const home = join(dir, '.gemini-hive');
+    const settingsPath = join(home, 'system-settings.json');
+    try {
+      mkdirSync(home, { recursive: true });
+      const shim = join(home, 'gemini-hook.cjs');
+      writeFileSync(shim, GEMINI_HOOK_SHIM, 'utf8');
+      const hook = (name: string, matcher?: string) => ({
+        ...(matcher ? { matcher } : {}),
+        sequential: true,
+        hooks: [{
+          name: `munder-hive-${name}`,
+          type: 'command',
+          command: this.nodeRunUnquoted(shim),
+          timeout: 30000
+        }]
+      });
+      const settings = {
+        hooksConfig: { enabled: true, notifications: false },
+        hooks: {
+          SessionStart: [hook('session-start')],
+          BeforeAgent: [hook('before-agent')],
+          BeforeTool: [hook('before-tool', '.*')],
+          AfterTool: [hook('after-tool', '.*')],
+          AfterAgent: [hook('after-agent')]
+        }
+      };
+      writeFileSync(settingsPath, JSON.stringify(settings, null, 2), 'utf8');
+    } catch (e) { console.error('[hive] installGeminiHooks failed:', e); }
+    return settingsPath;
   }
 
   /** Codex lifecycle-hook bridge → full hive parity for a `codex` worker (live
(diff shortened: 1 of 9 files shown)


chaitanyagiri:
## TL;DR

Thanks @FenjuFu — the Gemini CLI work looks substantial and it's clearly researched against the official docs. But I have to be blunt about the state it arrived in: **this PR cannot currently be reviewed, because ~97% of its diff is line-ending churn.**

**What I mean concretely:** the PR reports +5242/-5048 across 9 files. I counted the added lines: **5,115 of 5,251 end in a carriage return, and zero of the 5,057 removed lines do.** Your editor or git config rewrote every file you touched from LF to CRLF, so git sees every single line as changed. The real Gemini work is in there, but it's a few hundred lines hidden inside a whole-file rewrite of six source files and three test files.

**Why this blocks:** nobody can see what you actually changed, so nobody can review it. It will also conflict with essentially every other open PR that touches those files (currently at least six do), and it would silently flip line endings for every other contributor on those files.

**The good news:** this is entirely a git-config problem, not a code problem, and the fix takes about five minutes. It's also partly our fault — the repo has no `.gitattributes`, so nothing normalizes line endings for Windows contributors. Steps below.

---

## What I did

Fetched the diff and measured the CR distribution before reading any code, because a +5242/-5048 diff across `agentProvider.ts` and `hive.ts` is either a rewrite or an artifact and it changes how you review everything else. Then checked out the branch, ran the type-checker and full suite, and tested the merge.

## The measurement

```
lines added in the diff:        5,251
  … of which end in CR:         5,115   (97.4%)

lines removed in the diff:      5,057
  … of which end in CR:             0   (0%)
```

Every removed line is LF, every added line is CRLF. That is the signature of a wholesale conversion, not of edits.

```
   what the diff SAYS            what actually changed

   agentProvider.ts +635/-609    ~60 lines of Gemini preset
   hive.ts        +2649/-2551    ~40 lines of hook wiring
   config.ts       +396/-385     ~15 lines of model entries
   …                             (the rest is \n → \r\n)
```

## What I tested anyway

| Check | Result |
|---|---|
| `node --test test/*.test.cjs` | ✅ 260/260 (258 + your 2) |
| Merge against current `main` | ✅ clean **today** — but see below |
| Malware / obfuscation / network / new deps | ✅ none found |

So the branch is healthy — this is genuinely a presentation problem, not a broken PR. But "merges cleanly today" is misleading: the moment any other PR touching `hive.ts`, `agentProvider.ts`, `config.ts` or `OnboardingWizard.tsx` lands, this conflicts on **every line of those files**, and the conflict is unresolvable by hand.

## How to fix it

From your branch:

```bash
# 1. tell git to stop converting on checkout
git config core.autocrlf input

# 2. rewrite the working tree with LF endings
git rm --cached -r .
git reset --hard

# 3. confirm — this should now print a few hundred, not five thousand
git diff origin/main --stat
```

If step 3 still shows the whole file, the endings are baked into your commits and the cleanest path is to re-apply your changes onto a fresh branch off current `main`:

```bash
git checkout -b feat/gemini-cli-v2 origin/main
# copy your Gemini changes across, save with LF, commit
```

Once the diff is a few hundred lines I'll review the actual Gemini integration properly and quickly — the description reads well and the approach (per-agent settings file rather than mutating the user's global config, real lifecycle hooks rather than terminal-only) is the right one.

## This is partly on us

**The repo has no `.gitattributes`**, so there is nothing telling git to normalize line endings, and a Windows contributor with default settings will hit this every single time. You are the first person it's bitten visibly; you will not be the last.

@chaitanyagiri — this needs a repo-level fix regardless of what happens to this PR:

```gitattributes
* text=auto eol=lf
*.png binary
*.jpg binary
*.ico binary
*.icns binary
```

That normalizes on commit, so a CRLF working tree stops producing CRLF diffs. Worth landing before more Windows contributors arrive — this is the kind of friction that makes someone give up rather than ask.

## Two review notes I can give you now

Even without reading the diff properly, two things from your description:

1. **"install a per-agent Gemini system settings file instead of mutating the user's global settings"** — that's the right call and matches how the Codex preset works (its `CODEX_HOME` lives in the agent directory for the same reason). Good instinct.
2. **Scope.** Nine files including a provider preset, onboarding UI, a logo, automation wiring and three test files is a lot for one PR even at its real size. If the re-do is easy, consider splitting the preset + install metadata from the hook bridging — they can be reviewed independently and the first half could land while the second is discussed.

## Questions

@FenjuFu — (1) are you on Windows? If so, `core.autocrlf input` is worth setting globally, not just here. (2) Did you verify the hook bridging against a real `gemini` CLI, or is it built from the docs? Both are fine, but if it's docs-only I'd want a manual test note before merging, since a provider preset that doesn't quite match the real CLI fails in a way tests can't catch. (3) You mention this contributes to #104 "without overlapping the Kimi bridge work" — is someone else working on Kimi? Worth linking so the two don't collide the way #162 and #175 just did.

FenjuFu:
You're completely right, and thanks for taking the time to quantify it instead of just bouncing the PR — that made the problem obvious.

I confirmed the diagnosis locally by counting bytes on the raw blobs (not the working tree, so `autocrlf` couldn't lie to me):

- `92461ab:src/main/hive.ts` → 0 CR, 2609 LF (pure LF)
- `277344f:src/main/hive.ts` → 2596 CRLF + 111 bare LF

So my editor rewrote every file I opened from LF to CRLF, and the ~200 lines of actual Gemini work got buried under a whole-file EOL rewrite. Exactly what you described.

I've rebuilt the branch: same tree content, but the nine touched files are normalized back to LF, so the diff is now what it should have been — **200 insertions / 6 deletions**, no EOL churn. `node --test test/provider-config.test.cjs` passes (including the new `Gemini CLI has hooks, interactive seeding, resume, and current yolo mode` case). Force-pushed to `feat/gemini-cli-provider`.

To keep this from recurring I'll add a `.gitattributes` (`* text=auto eol=lf`) in a follow-up if you'd like the repo to pin EOL normalization. The PR should be reviewable now — sorry for the noise on the first pass.

FenjuFu:
Thanks for the review notes — answering your three questions:

1. **Yes, I'm on Windows.** I've set `core.autocrlf input` globally now so my working tree stops rewriting LF→CRLF on checkout; that's what produced the churn on the first pass. And +1 on landing the repo-level `.gitattributes` (`* text=auto eol=lf`) regardless — I'm happy to send that as a small standalone PR so it's in before more Windows contributors hit this.
2. **Verified against a real `gemini` CLI**, not docs-only. I ran the preset end-to-end against an actual install — install/seeding, the per-agent system-settings file (kept out of the user's global config, matching the Codex `CODEX_HOME` pattern you mentioned), resume, and the hook bridging all exercised against the real CLI, not just modelled from the docs.
3. **No one else has an open Kimi PR that I can see** — #104 is the umbrella "add new CLI agents" issue and Kimi isn't claimed there. The "without overlapping the Kimi bridge work" note just meant I kept the hook/automation wiring generic rather than Gemini-specific, so a future Kimi preset can reuse the same bridge without a collision like #162/#175. Happy to link things up if/when someone picks Kimi up.

On **scope**: good call. If it helps review, I can split this into (a) the provider preset + install metadata and (b) the hook bridging, so the first half can land while the second is discussed. Let me know if you'd like me to break it up that way — otherwise it's ready for another look as-is.

PR #177, opened 2026-08-19
feat(hive): flag each agent's live context-window occupancy in god's roster

TTAWDTT:
## Summary

God's injected roster line now carries a **`ctx NN%`** occupancy figure per agent, so Michael can see at a glance whose context window is nearly full when routing work — instead of guessing from fleet.json's cumulative token spend.

- **`src/main/hive.ts`** — `rosterContext()` gains an optional `ctxOf` callback; each agent's status bits get `ctx NN%` (clamped to 0–100) when the callback returns live data with `limit > 0`. Omitted when no Status tick has fired.
- **`src/main/hooks.ts`** — the roster call feeds the LIVE occupancy via `rosterContext((id) => this.contextFor(id))` (dependency stays HookServer → HiveManager via callback).
- **`test/hive-roster-injection.test.cjs`** — new test asserts occupancy wiring end-to-end (62% / 99% shown, no Status data → no ctx field).

## Why

`contextById` (from the statusLine shim) is already tracked per agent and exposed via `contextFor()`. This change surfaces "whose context is ${X}% full" directly on the roster God receives on SessionStart and every prompt.

## Verification

- `npm run typecheck` (node + web): clean
- `node --test` hive/roster suites: 22 pass, 0 fail
- New test exercises the full hook path (statusLine → `contextFor` → roster string)

## Notes

- `ctx NN%` uses LIVE window occupancy, not fleet.json's cumulative `tokens` (a spend figure).
- To fully benefit, God's system prompt could later route work to the lowest-occupancy agent — this PR only adds the field.

### DIFF
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -1825,10 +1825,20 @@ export class HiveManager {
    * told to read fleet.json, but "told to" is not "always knows" — so we push the
    * truth in on every turn instead. One line, so the cost is negligible.
    *
+   * `ctxOf` (optional, supplied by HookServer) lets the caller layer the LIVE
+   * context-window occupancy on top of the disk snapshot — each agent gets a
+   * `ctx NN%` so god can see at a glance whose context is nearly full when it
+   * routes work. fleet.json only carries cumulative `tokens`, which is a spend
+   * figure, not how full the CURRENT window is; the real occupancy lives in
+   * HookServer.contextById (from the statusLine shim). Omitted when the callback
+   * is absent or an agent has no Status tick yet.
+   *
    * Returns null when there is nothing to say (no hive, no snapshot, no agents),
    * so the hook stays a no-op rather than injecting noise.
    */
-  rosterContext(): string | null {
+  rosterContext(
+    ctxOf?: (agentId: string) => { tokens: number; limit: number } | undefined
+  ): string | null {
     const root = this.root();
     if (!root) return null;
     try {
@@ -1862,6 +1872,15 @@ export class HiveManager {
         if (a.inboxBacklog) bits.push(`inbox ${a.inboxBacklog}`);
         if (a.breaker && a.breaker !== 'ok' && a.breaker !== 'none') bits.push(`breaker ${a.breaker}`);
         if (a.isGod) bits.push('you');
+        // Live context-window occupancy from the statusLine shim — lets god see
+        // which agents are near-full when routing, instead of guessing from the
+        // cumulative token count. Clamp to 0-100; a fresh meter can briefly
+        // report more than 100% before a window rotation.
+        const cw = ctxOf?.(a.id);
+        if (cw && cw.limit > 0) {
+          const pct = Math.max(0, Math.min(100, Math.round((cw.tokens / cw.limit) * 100)));
+          bits.push(`ctx ${pct}%`);
+        }
         return `${a.id}${a.name ? ` "${a.name}"` : ''} (${bits.join(', ')})`;
       });
       const more = agents.length > shown.length ? ` +${agents.length - shown.length} more` : '';
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -1100,6 +1100,8 @@ export class HiveManager {
     // Native-separator path helpers — see the 🪟 note above.
     const inDir = (...parts: string[]): string => join(dir, ...parts);
     const inRoot = (...parts: string[]): string => join(root, ...parts);
+    const ctxLine = 'LIVE CONTEXT: each agent row in the LIVE ROSTER carries a `ctx NN%` tag — its live context-window occupancy. Treat it as the real headroom signal when routing: prefer an agent with a LOW `ctx` for a big task; treat a HIGH `ctx` (near 100%) as busy rather than idle, even if the cumulative token count looks modest.';
+
     const memoryLine = semanticMemory
       // The palace location is named, not spelled as `$MEMPALACE_PALACE_PATH`:
       // `mempalace` reads that env var itself, and the POSIX `$` form was noise
@@ -1139,6 +1141,7 @@ export class HiveManager {
       knowledgeLine,
       godLine,
       slackLine,
+      ctxLine,
       `Env vars available to you: AGENT_ID, AGENT_NAME, HIVE_ROOT, AGENT_DIR.`
     ].filter(Boolean).join('\n');
   }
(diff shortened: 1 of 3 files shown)


chaitanyagiri:
## TL;DR

Thanks @TTAWDTT — small, clean, additive, and it surfaces information that genuinely isn't available anywhere else today. My reservation is about the *consumer*, not the code.

**Context:** the god agent ("Michael") gets a roster injected into its prompt on SessionStart and again on every user prompt — a line per agent with status bits, built by `rosterContext()` in `src/main/hive.ts`. Today the only quantity on that line comes from `fleet.json`, which tracks **cumulative token spend**. That is a bill, not a gauge: an agent that has spent 2M tokens across twenty compactions may have a nearly empty context window right now. The number god actually needs when deciding who to hand work to — how full is this agent's *current* window — is already tracked (`HookServer.contextById`, fed by the statusLine shim) but never reaches the roster.

**What this adds:** an optional `ctxOf` callback on `rosterContext()`. When supplied, each agent line gains `ctx NN%`. `HookServer` supplies it via `contextFor(id)`. Omitted when there's no Status tick yet, clamped 0–100.

**Verdict:** correct and low-risk. Typecheck clean, 259/259 (1 new), merges cleanly, 3 commits behind `main`. Ship it — but please read the question about god's prompt, because as written this may be a number nobody acts on.

---

## What I did

Read the diff, confirmed `contextById` / `contextFor` already exist on `main` and carry live window data rather than cumulative spend, checked that the callback is genuinely optional so nothing breaks for callers that don't pass it, checked out the branch, ran the type-checker and full suite, ran your new test, and tested the merge.

## Why the distinction you're drawing is real

This is the part I want to make explicit, because "we already show tokens" is the objection this PR will get:

```
   fleet.json  tokens: 2,140,000      ← everything this agent has ever spent
        │                               (grows forever; survives compaction)
        │  tells god: "this agent has been busy"
        │  tells god NOTHING about whether it can take more work right now

   contextById  { tokens, limit }    ← how full the CURRENT window is
        │                               (drops back after a compaction)
        └─ tells god: "this one is at 94% — give the next job to someone else"
```

Routing work to an agent at 94% means it compacts almost immediately and loses the context you just gave it. That's a real, recurring cost, and cumulative spend cannot predict it. So the feature is justified.

The implementation is the right shape too: an optional callback keeps `HiveManager` from having to know about `HookServer`, so the dependency direction stays HookServer → HiveManager. `rosterContext()` with no argument behaves exactly as before, which is why nothing else in the suite moved.

## What I tested

| Check | Result |
|---|---|
| `node --test test/*.test.cjs` | ✅ 259/259 (258 + your 1) |
| Your test covers 62% / 99% shown and "no Status data → no ctx field" | ✅ all three |
| `rosterContext()` with no callback is unchanged | ✅ |
| Clamp holds for a meter briefly reporting >100% | ✅ `Math.max(0, Math.min(100, …))` |

## My one real reservation

Your own Notes section says it:

> *To fully benefit, God's system prompt could later route work to the lo[west]…*

Right now this adds `ctx NN%` to a line god reads, and **nothing tells god what to do with it.** An LLM handed an unexplained percentage will sometimes use it sensibly and sometimes ignore it entirely, and you won't be able to tell which from the outside. So the feature is either finished or half-finished depending on a file this PR doesn't touch.

I'd rather it were finished here. That's one sentence in god's prompt, something like: *"`ctx NN%` is how full that agent's context window is right now — prefer routing new work to agents below ~70%, and avoid anything above ~90%."* Two reasons to do it in this PR rather than a follow-up: the thresholds only make sense next to the code that produces the number, and a "later" that nobody files becomes never.

If you'd rather not touch god's prompt in the same change — a fair position, it's a sensitive file — then please open the follow-up issue now and link it, so the gap is tracked rather than remembered.

## Two smaller notes

1. **Cost on a hot path.** `rosterContext()` runs on SessionStart *and every UserPromptSubmit*, and it now calls `ctxOf` once per agent. That's a map lookup, so it's cheap — but it's worth knowing that this path is already carrying more than it used to (#153 adds a synchronous roster read+parse to the same boundary). Nothing to change here; just don't let this become the place where per-prompt work accumulates unnoticed.
2. **`limit > 0` is doing real work.** A provider that reports a window size of 0 or omits it would otherwise produce `Infinity%` or `NaN%` in god's prompt. Good guard — I'd add a one-line comment saying that's *why* it's there, since it reads as defensive noise otherwise.

## Questions

@TTAWDTT — (1) will you add the sentence to god's prompt here, or open the follow-up? Either's fine, I just don't want it dropped. (2) Do all providers populate `contextById`, or is this Claude-only via the statusLine shim? If it's Claude-only, a mixed floor shows `ctx` on some lines and not others, and god may read the *absence* as "no context used" rather than "unknown" — worth a word in the roster legend if so.

@chaitanyagiri — the only decision here is whether you want god's prompt edited by a contributor PR. If not, say so and this can merge as data-only with a tracked follow-up.

TTAWDTT:
@chaitanyagiri 
A follow-up commit (78b3948) takes the occupancy field the rest of the way: it teaches God's injected prompt how to read ctx NN% and use it as the headroom signal when routing, and it adds a roster legend so the tag is self-explanatory.

Two additions to src/main/hive.ts:

1. A LIVE CONTEXT line in God's system prompt (injectedPrompt). It tells Michael to treat each roster row's ctx NN% as the real headroom signal when routing — prefer an agent with a LOW ctx for a big task, treat a HIGH ctx (near 100%) as busy rather than idle. fleet.json's tokens is a lifetime spend counter; ctx NN% is the live window occupancy, which is the figure that actually tells you whether an agent has room for a big task right now.

2. A conditional roster legend (rosterContext). When at least one agent has reported occupancy, the roster gains a short footnote: `ctx NN%` = live window occupancy; absent = not yet reported (unknown, not empty). It's omitted when no Status tick has fired yet, so an early roster stays clean — this also keeps the existing "no occupancy when no Status data" assertion green without weakening it.

On the mechanism, one clarification to the PR description: the occupancy path isn't Claude-specific. The context_window figure is read off whatever the statusLine shim's --status returns (hive.ts:873), and stored per-agent from the Status event (hooks.ts:132–141). Any provider whose status output carries context_window will populate it. What can be genuinely absent is a fresh tick — no Status has fired yet, or the payload didn't carry a context_window/limit — which is exactly the "unknown, not empty" case the prompt and legend now call out, so Michael never mistakes a missing tag for an empty or tiny window.

Verification: npm run typecheck:node clean; node --check src/main/hive.ts clean; the hive/roster suites pass (hive-roster-injection 5/5 incl. the occupancy test, plus the other hive files). Final diff is tiny — 8 insertions, 0 deletions.

PR #178, opened 2026-08-19
feat(workers): god-hired workers get a floor card, and it archives on death

gpechieu:
Split out of #159 per review — this half is a **product decision**, the three that stayed there are bug fixes.

**Stacked on #159:** the first **two** commits here ARE #159 — its original commit plus the provider-aware auto-mode fix from the review of this PR, which lives there because that's the commit that introduces `buildWorkerLaunch` (merging #159 alone must not ship the defect). The card's archive path rides the explicit `teardownPty` calls that PR adds — without them a released worker never reaches the place that archives its card. Review the top commit only; I'll rebase onto `main` the moment #159 lands. (Setting this PR's base to #159's branch isn't possible cross-fork: GitHub only accepts base branches from the base repo.)

## What it changes

A god-hired worker was invisible: no floor card, no roster entry, and after an app restart nothing offered to restore it — so failures were silent and sessions unrecoverable. This deliberately softens "workers are headless-by-design":

- `processSpawnRequest` broadcasts `hive:agentSpawned` (same mechanism and reason as the voice-spawn broadcast), so the worker cards on the floor and enters the roster.
- `teardownPty` broadcasts `hive:agentArchived` for workers — and since **every** worker death (done-release, token/idle reap, manual stop, crash) funnels through `teardownPty` once #159 lands, the card can never freeze alive.

## The restore question, answered in code

What happens when the human restores a worker card after an app quit: it revives through the renderer's **normal** spawn path and never re-enters `liveWorkers` (`liveWorkers.set` has exactly one call site — `processSpawnRequest`). So a restored worker is a regular agent: no done-scan, no reaping, no token cap. **Ephemerality is a property of the hiring, not of the card** — that sentence is now a comment at the broadcast site.

## Validation

`npm run typecheck` clean, **272/272** tests on top of the rebased #159 (267 + 5 net new for per-provider auto-mode).


chaitanyagiri:
## TL;DR

Thanks @gpechieu — splitting the product decision out of #159 was exactly the right response to that review, and you also fixed the tokenizer duplication I flagged there. That's two pieces of review feedback acted on properly, and it makes both PRs easier to reason about.

**Context:** the app has two ways an agent gets created. A human hire goes through the renderer, which cards it on the floor and puts it in the roster. A **god-hired worker** is created by the main process from a spawn request, and the renderer never hears about it — so it had no floor card, no roster entry, and after an app restart nothing offered to restore it. Failures were silent and sessions were unrecoverable.

**What this changes:** `processSpawnRequest` broadcasts `hive:agentSpawned` so the worker cards on the floor, and `teardownPty` broadcasts `hive:agentArchived` so the card can't freeze alive. It also extracts the command-line splitting into `src/shared/commandLine.ts` and adds `src/main/workerLaunch.ts` as a pure, unit-tested function.

**Verdict:** the structure is right and the tests are the best part. Typecheck clean, 267/267 (9 new), merges cleanly, 3 commits behind `main`. One real defect below — the auto-mode inheritance is provider-blind — plus the stacking to sort out.

---

## What I did

Read the top commit in isolation (the first commit is #159, per your note), read `buildWorkerLaunch` against the renderer's own spawn path to check the two agree, checked out the branch, ran the type-checker and the full suite, ran your new tests, tested the merge, and cross-checked the tokenizer extraction against the copy I flagged on #159.

## The two things you did right that I want to name

**1. You made the translation a pure function and tested it.** Your comment says why:

> *this exact translation silently killed real workers for days while reporting success, which is what earned it a unit test.*

That's the correct instinct. `buildWorkerLaunch` takes plain data and returns `{ bin, args, command }` with no Electron, no filesystem, no PTY — so the thing that broke can now be asserted directly instead of only observable by spawning a real process and watching it die.

**2. You killed the duplicate tokenizer.** On #159 I noted the command-splitting regex had been copied byte-for-byte from `src/renderer/src/store/config.ts` into main. Here it moves to `src/shared/commandLine.ts` and the renderer re-exports from it, so there's one implementation. Your comment gets the reason right too — a worker whose command line splits differently from the renderer's is precisely the silent breakage this whole area exists to prevent.

## What I tested

| Check | Result |
|---|---|
| `node --test test/*.test.cjs` | ✅ 267/267 (258 + your 9) |
| `test/worker-launch.test.cjs` alone | ✅ passes, incl. quoted `--model "Gemini 3.1 Pro (High)"` |
| One tokenizer implementation, renderer re-exports it | ✅ confirmed |

## The defect: auto-mode is Claude-only

```ts
if (
  opts.autoMode &&
  inferAgentProvider(command, opts.requestProvider) === 'claude' &&
  !command.includes('--permission-mode')
) {
  command += ' --permission-mode bypassPermissions';
}
```

Both the provider and the flag are hardcoded. But the preset system already knows every provider's auto flag, and the renderer already uses it:

```ts
// src/renderer/src/store/config.ts
if (config.autoMode && preset.autoFlag) cmd = `${cmd} ${preset.autoFlag}`;
```

So on a mixed floor with auto-mode on, a **GUI-hired** Codex agent gets `--dangerously-bypass-approvals-and-sandbox` and a **god-hired** Codex worker gets nothing. Your own comment states the consequence of missing the flag — *"without the flag it stalls at the first ask until the idle reaper kills it"* — so this is a live failure for every non-Claude worker, not a theoretical one.

There's a third wrinkle: `src/main/config.ts` already exports `commandForAutoMode()`, which does this correctly per provider via `autoModeFlagForProvider`. It has **zero call sites** — dead code that solves your exact problem. So the queue now has three implementations of one idea, and the only one that's wrong is the one that runs for workers.

The fix is small and fits your pure-function shape:

```ts
const provider = inferAgentProvider(command, opts.requestProvider);
const flag = autoModeFlagForProvider(provider);
if (opts.autoMode && flag && !command.includes(flag.split(' ')[0])) {
  command += ` ${flag}`;
}
```

and then either use `commandForAutoMode` or delete it, so the dead third copy stops being a trap. A test per provider is cheap given `buildWorkerLaunch` is already pure.

## The stacking

You've flagged it clearly in the description — thank you, that's what made this reviewable. Just restating for whoever merges: **the first commit here is #159**, review only the top commit, and this needs a rebase once #159 lands. Setting this PR's base branch to #159's branch would make GitHub show only the new commit and remove the risk of someone reviewing the wrong diff.

## A question about the product decision

Your description is honest that this "deliberately softens *workers are headless-by-design*", and answers the restore question in code: a restored worker revives through the renderer's normal spawn path and never re-enters `liveWorkers`, so it becomes a regular agent — no done-scan, no reaping, no token cap.

That's a correct description of the behaviour, but I'm not sure it's the *desired* one. A worker the god hired with a token cap, restored after a quit, comes back with no cap and no reaper. Nothing will ever clean it up. That's arguably worse than not restoring it. Worth deciding explicitly rather than inheriting.

## Questions

@gpechieu — (1) will you fix the auto-mode provider blindness here? It's the one thing I'd block on, and it's ~5 lines given your pure function. (2) On the restore behaviour above — is "restored worker becomes a normal uncapped agent" intended, or should a restored worker be refused, or re-registered as a worker with its cap? (3) Any objection to deleting `commandForAutoMode` in this PR, since you're already the person consolidating this logic?

@chaitanyagiri — the product call is yours: should a god-hired worker be visible on the floor at all? This PR argues yes and I find the argument convincing (silent failures and unrecoverable sessions are worse than a slightly busier floor), but it is a reversal of a stated design principle and shouldn't land by accident.

gpechieu:
Thanks for the review — all three questions answered below, two of them in code.

## (1) Auto-mode provider blindness: fixed — one branch over

Your shape, adopted, with one relocation and one refinement.

**The relocation.** `buildWorkerLaunch` is introduced by #159's commit, not this PR's top commit — so the fix lives as an incremental commit on #159's branch (`5d38c86`), and this branch is rebased on top of it. Fixing it only here would have let #159 merge alone and ship the defect. #159 moved by **fast-forward**, so your earlier review of `51332ab` still describes exactly the code it reviewed; the delta is the one small commit.

```ts
const provider = inferAgentProvider(command, opts.requestProvider);
const autoFlag = opts.autoMode ? autoModeFlagForProvider(provider) : '';
if (autoFlag && !tokenizeCommand(command).includes(tokenizeCommand(autoFlag)[0])) {
  command += ` ${autoFlag}`;
}
```

**The refinement.** Your sketch checked `!command.includes(flag.split(' ')[0])` — a substring test. copilot's flag is `-s --allow-all-tools --no-ask-user`, and `-s` is a substring of `--summarize`, so a worker spawned as `copilot --summarize` would silently skip the append. The stance check is token-level instead, and a test pins exactly that case.

Tests added, one per behavior: provider's own flag (codex/agy/kimi), explicit stance wins without doubling the flag, empty-flag providers (opencode/custom) get nothing, multi-token flag appends whole, explicit `requestProvider` beats binary inference. **272/272** and typecheck clean on both branches.

## (3) `commandForAutoMode`: deleted

No objection — same commit. Zero call sites confirmed (`git grep` finds only the definition), and its three imports (`autoModeFlagForProvider`, `defaultCommandForProvider`, `inferAgentProvider`) had no other consumer in `config.ts`, so they're gone too. Two implementations remain: the renderer's preset append and `buildWorkerLaunch`, both reading the same preset field.

## (2) The restore behaviour: half intended, half inherited — you're right to split them

Two paths reach a dead worker's card, and they deserve different answers:

- **Manual restore** (human clicks restore on an archived card): becoming a normal, uncapped agent is *intended*. The cap and the reaper were terms of god's hiring, and that hiring ended when the worker died; a human restoring the card is a **new hiring by the human**, who owns its lifecycle like any agent they hire from the modal. That's the "ephemerality is a property of the hiring" sentence doing real work.
- **Auto-restore at boot** (worker alive at app quit → boot reconcile marks it restorable → auto-restore revives it with `resume:true`): here you've caught something real. Nobody decided anything — the worker comes back uncapped and unreaped *silently*, and quitting the app becomes an accidental cap-removal machine. That part is inherited, not intended.

The fix I'd propose: **exclude god-worker cards from auto-restore only.** The card stays on the floor as restorable — the session recoverability this PR exists for is intact — but reviving it takes the human's explicit click, which is exactly the "new hiring" the manual path already models. Small diff: the spawn broadcast tags the card, auto-restore filters on the tag. The alternative (re-registering into `liveWorkers` with its original cap) needs cross-restart token accounting to mean anything, and god's request archived long ago — nobody is waiting for that worker's result anymore.

I'll implement the auto-restore exclusion in this PR if you want it here — holding until @chaitanyagiri's visibility call lands, since a "no floor cards" verdict moots the whole question.

## The base-branch suggestion

Can't, unfortunately: GitHub only accepts base branches from the base repo, and #159's branch lives on the fork. Unless you want `fix/god-worker-spawn-lifecycle` mirrored into your repo as a throwaway base, the "review only the top commit" note has to carry it — the description now says the first **two** commits are #159. I'll rebase onto `main` the moment #159 lands, as before.

PR #179, opened 2026-08-19
fix(fs): use expandTilde in statAbs for cross-platform path resolution

kdahal7:
statAbs previously checked p.startsWith('~/') directly, missing bare '~', Windows backslash paths ('~\\...'), and leading/trailing whitespace. Updated statAbs to use the central expandTilde function so tilde expansion works consistently across all OS platforms and path formats.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

## Discord (optional)

Discord: 

## Checklist

- [ ] `npm run typecheck` passes (the de-facto CI gate).
- [ ] `npm run build` succeeds.
- [ ] PR is focused on a single change.

### DIFF
--- a/src/main/fs.ts
+++ b/src/main/fs.ts
@@ -218,9 +218,8 @@ export function normalizeHiveHome(
  *  the home dir). Read-only metadata: returns whether a regular file exists and
  *  the normalized absolute path; never file contents. */
 export async function statAbs(p: string): Promise<{ exists: boolean; isFile: boolean; path: string }> {
-  let abs = p.startsWith('~/') ? join(homedir(), p.slice(2)) : p;
+  const abs = expandTilde(p);
   if (!isAbsolute(abs)) return { exists: false, isFile: false, path: p };
-  abs = normalize(abs);
   try {
     const s = await stat(abs);
     return { exists: true, isFile: s.isFile(), path: abs };
--- a/test/expand-tilde.test.cjs
+++ b/test/expand-tilde.test.cjs
@@ -13,7 +13,7 @@ const os = require('node:os');
 const path = require('node:path');
 const loadTs = require('./load-ts.cjs');
 
-const { expandTilde } = loadTs('src/main/fs.ts');
+const { expandTilde, statAbs } = loadTs('src/main/fs.ts');
 
 const HOME = os.homedir();
 
@@ -120,4 +120,25 @@ test('#140: both entry points agree on the same directory', () => {
   // different one — the failure this pair of fixes exists to prevent.
   const typed = '~/HarnessAgents';
   assert.equal(expandTilde(typed), normalizeHiveHome(typed).home);
+});
+
+test('statAbs expands bare ~, Windows-style ~\\, and whitespace paths', async () => {
+  const fileBasename = `.md-statabs-${process.pid}`;
+  const inHome = path.join(HOME, fileBasename);
+  fs.writeFileSync(inHome, 'x');
+  try {
+    const resSlash = await statAbs(`~/${fileBasename}`);
+    assert.equal(resSlash.exists, true);
+    assert.equal(resSlash.isFile, true);
+
+    const resWin = await statAbs(`~\\${fileBasename}`);
+    assert.equal(resWin.exists, true);
+    assert.equal(resWin.isFile, true);
+
+    const resPadded = await statAbs(` ~/${fileBasename} `);
+    assert.equal(resPadded.exists, true);
+    assert.equal(resPadded.isFile, true);
+  } finally {
+    fs.rmSync(inHome, { force: true });
+  }
 });


chaitanyagiri:
## TL;DR

Thanks @kdahal7 — small, correct, and it removes a real inconsistency.

**Context for anyone reading this cold:** `statAbs` in `src/main/fs.ts` is the main-process helper that answers "does this path exist, and is it a file?" for paths the UI hands over. Paths reaching it can come from a text field, so they may start with `~`. The repo already has one function whose whole job is turning `~` into a real home path — `expandTilde`, in the same file — but `statAbs` was doing its own thing instead, matching only the literal prefix `~/`.

**What this changes:** two lines become one. `statAbs` now calls `expandTilde` like every other consumer.

**What that fixes:** a bare `~`, a Windows-style `~\Projects\thing`, and a path with stray leading whitespace all now resolve. Before, each fell through unexpanded and `isAbsolute` rejected it, so `statAbs` reported `exists: false` for a file that was really there.

**Verdict:** correct, in the right direction, and safe. Typecheck clean, 258/258 tests pass, merges cleanly, 3 commits behind `main`. My only ask is a test. Details below.

---

## What I did

Read the diff, read `expandTilde`'s implementation on `main` to confirm the two behaviours are actually equivalent where they overlap, checked out the branch, ran the type-checker and the full suite, and tested the merge against current `main`.

## The one thing I wanted to be sure of

Your diff removes a line as well as changing one:

```diff
-  let abs = p.startsWith('~/') ? join(homedir(), p.slice(2)) : p;
+  const abs = expandTilde(p);
   if (!isAbsolute(abs)) return { exists: false, isFile: false, path: p };
-  abs = normalize(abs);
```

Dropping `normalize(abs)` is the part worth checking, because normalization is what collapses `..` and doubled separators — and a stat on an un-normalized path is a subtly different thing. So I read `expandTilde`:

```ts
export function expandTilde(p: string): string {
  ...
  if (!isAbsolute(out)) return t;
  return resolve(out);          // <-- this
}
```

`resolve()` normalizes as part of its contract, so on every path that reaches the `stat` call the normalization still happens. **Removing that line is safe.** I'd have hesitated if `expandTilde` had ended in a bare `join`, so I'm noting the reasoning rather than just the conclusion.

I also checked the early-return: `expandTilde` returns the *trimmed* string when the result isn't absolute, but `statAbs` returns `path: p` — the original, untrimmed input — in that branch. So a caller comparing the returned path against what it passed in still gets its own string back. Unchanged behaviour, and the right choice.

## What I tested

| Check | Result |
|---|---|
| `node --test test/*.test.cjs` | ✅ 258/258 |
| `normalize` / `homedir` still used elsewhere in `fs.ts` (no dead imports) | ✅ 2 and 3 other uses |
| Malware / obfuscation / network calls / new dependencies | ✅ none |

## The one thing I'd like added

**A test.** This repo's bar for a bug fix is a test that fails before the change and passes after, and here that's about six lines:

```js
test('statAbs expands a bare ~ and a Windows-style ~\\ path', async () => {
  const inHome = path.join(os.homedir(), `.md-statabs-${process.pid}`);
  fs.writeFileSync(inHome, 'x');
  try {
    assert.equal((await statAbs(`~${path.sep}.md-statabs-${process.pid}`)).exists, true);
    assert.equal((await statAbs(` ~/.md-statabs-${process.pid} `)).exists, true);
  } finally { fs.rmSync(inHome, { force: true }); }
});
```

Without it, nothing stops someone reverting to a hand-rolled prefix check in six months. It's the cheapest possible insurance on a fix this small, and it also documents *which* inputs you were fixing — your PR description lists three cases that the diff alone doesn't reveal.

## A note on the PR description

Your first paragraph is genuinely good: it names the three inputs that were broken and says why. But the template below it is still unfilled — the "What & why" section is empty and no checklist boxes are ticked. Since your real description is already better than the template's prompt, just delete the unfilled template or move your paragraph into it. A reviewer scrolling the queue sees the template first and can't tell the PR has a description at all.

## Related work you may want to know about

There are two other open PRs in the same neighbourhood: **#160** and **#165** both fix `~` handling on the harness-home path. Yours touches a different function and doesn't conflict with either, so no action needed — but if you enjoyed this one, that family of "a typed path with a `~` in it" bugs is clearly not exhausted, and a sweep for any remaining hand-rolled `startsWith('~')` checks would be a welcome follow-up.

## Questions

@kdahal7 — happy to add the test for you if you'd rather not, but I'd prefer it came from you since you know which of the three cases you actually hit. Which one was it — bare `~`, the Windows form, or the whitespace? That tells us whether there's a UI field somewhere that isn't trimming its input, which would be a second bug worth its own fix.

@chaitanyagiri — nothing needed from you here. This one is ready once a test lands.

kdahal7:
Thanks for the thorough review and helpful context! 

I have added the unit test covering bare `~`, Windows-style `~\`, and padded whitespace paths to `test/expand-tilde.test.cjs`, and force-pushed the updated commit. 

To answer your question: I hit the issue on Windows with `~\` backslash paths when navigating file references. 

I've also cleaned up the PR description template. Ready for another look whenever you have a moment!

PR #181, opened 2026-08-19
fix(webhook): dispatch through atomic addTask to avoid stale-ledger overwrite

TTAWDTT:
## Problem

`dispatchWebhookWork` re-read the task ledger, appended one card, and wrote the **whole array** back with `hive.writeTasks([...existing, card])`. This recreates exactly the stale-overwrite race that `hive.addTask` was added to prevent (the 2026-08-15 webhook card-loss regression, guarded by `test/hive-task-mutation.test.cjs`): if a Slack/god/voice/another-webhook caller added a card **between** this process's `tasks()` read and its later write, that card is silently dropped.

The existing `addTask(card)` reads the latest on-disk ledger and is idempotent by task id — precisely the safe primitive this path should use.

## Change

Route the webhook card through `hive.addTask(card)` instead of re-reading a snapshot and calling the whole-ledger `writeTasks([...existing, card])`.

## Test

Adds a source regression test asserting `dispatchWebhookWork` calls `hive.addTask(card)` and never rebuilds a stale `[...existing]` snapshot.

- `node --test test/hive-task-mutation.test.cjs test/webhook-endpoints.test.cjs` → 17/17 pass
- `npm run typecheck` → clean

### DIFF
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -610,7 +610,7 @@ export class HiveManager {
       lastSeen: Date.now()
     };
     if (meta.isGod) reg.godId = meta.id;
-    this.writeJson(join(root, 'registry.json'), reg);
+    this.atomicWriteJson(join(root, 'registry.json'), reg);
 
     this.appendLog({ kind: 'spawn', agentId: meta.id, name: meta.name, isGod: !!meta.isGod });
     // Only logs on an invalid cwd (rare) — not a per-spawn line, so no log spam.
@@ -808,7 +808,7 @@ export class HiveManager {
       if (!agent || agent.archived === archived) return;
       agent.archived = archived;
       agent.lastSeen = Date.now();
-      this.writeJson(join(root, 'registry.json'), reg);
+      this.atomicWriteJson(join(root, 'registry.json'), reg);
       this.appendLog({ kind: 'archive', agentId: id, archived });
       this.commit(`hive: ${archived ? 'archive' : 'unarchive'} ${id}`);
     } catch { /* best-effort — never crash a lifecycle handler */ }
@@ -830,7 +830,7 @@ export class HiveManager {
       if (!agent || agent.sessionId === sessionId) return; // unknown agent or unchanged → no write
       agent.sessionId = sessionId;
       agent.lastSeen = Date.now();
-      this.writeJson(join(root, 'registry.json'), reg);
+      this.atomicWriteJson(join(root, 'registry.json'), reg);
       this.appendLog({ kind: 'session', agentId, sessionId });
       this.commit(`hive: session ${agentId}`);
     } catch { /* best-effort — never crash a hook handler */ }
@@ -1037,7 +1037,7 @@ export class HiveManager {
     if (fresh.length === 0) return { block: false };
 
     cursor.lastProcessed = fresh[fresh.length - 1].id;
-    this.writeJson(cursorPath, cursor);
+    this.atomicWriteJson(cursorPath, cursor);
     this.appendLog({ kind: 'drain', agentId, count: fresh.length });
 
     const lines = fresh.map((m) => `- [from ${m.from}, ${m.act}] ${m.subject}: ${m.body}`).join('\n');
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -1680,8 +1680,6 @@ function dispatchWebhookWork(arg: {
   origin: 'webhook' | 'org';
 }): boolean {
   try {
-    const ledger = hive.tasks() as { tasks?: HiveTask[] };
-    const existing = Array.isArray(ledger?.tasks) ? ledger.tasks : [];
     const card: HiveTask = {
       id: arg.taskId,
       title: arg.title,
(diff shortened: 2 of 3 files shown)


chaitanyagiri:
## TL;DR

Thanks @TTAWDTT — you found a genuine regression of a bug this repo already paid for once, and you fixed it with the primitive that was added to prevent exactly this. Strong PR.

**Context:** the hive's kanban lives in one file, `tasks.json`. There are several things that can add a card: the god agent, a Slack message, a voice command, and an inbound webhook. The unsafe way to add one is read-the-whole-array → append → write-the-whole-array back, because anything that added a card between your read and your write gets overwritten. That exact bug caused a card-loss regression on 2026-08-15, and the fix was `hive.addTask()`, which appends against the *latest* on-disk ledger and is idempotent by task id.

**The bug:** `dispatchWebhookWork` never got the memo. It was still doing `hive.writeTasks([...existing, card])` — the unsafe pattern, in the one code path most likely to fire concurrently with the others, since a webhook arrives on someone else's schedule.

**The fix:** call `hive.addTask(card)` instead. Two lines of setup deleted, one call swapped.

**Verdict:** correct, minimal, and it comes with a test. Typecheck clean, 259/259 (1 new), merges cleanly, 2 commits behind `main`. One overlap to sort out with your own #182, and one broader question below.

---

## What I did

Read the diff, traced `addTask` on `main` to confirm it re-reads from disk rather than from a cached ledger, checked whether any *other* caller still has the unsafe pattern, checked out the branch, ran the type-checker and the full suite, ran your new test specifically, and tested the merge.

## Why this is the right fix and not just a different one

The distinction that matters is *when* the read happens:

```
   BEFORE — read early, write late              AFTER — addTask reads at write time

   t0  webhook reads ledger  → [A]              t0  webhook builds its card
   t1  Slack adds card B     → [A, B]           t1  Slack adds card B      → [A, B]
   t2  webhook writes [A, C] → [A, C]           t2  webhook addTask(C)
                                    ▲                  └─ re-reads → [A, B] → [A, B, C]
                             card B is gone
```

The window between t0 and t2 is not small — the webhook path does other work in between. And because `writeTasks` overwrites wholesale, the loss is total and silent: no error, no log, the card simply never existed.

I confirmed `addTask` really does re-read (`this.tasks()` inside the method, not a parameter), and that it's idempotent by id, so your comment — "a fresh taskId never collides, so this always adds" — is accurate rather than hopeful.

Your test asserts at the source level that `dispatchWebhookWork` calls `hive.addTask`. That's a blunt instrument, but for this specific defect it's the right one: the bug is "this call site uses the wrong primitive", and a source assertion is what catches a future revert. It's also consistent with the existing `test/hive-task-mutation.test.cjs` that guards the original regression.

## What I tested

| Check | Result |
|---|---|
| `node --test test/*.test.cjs` | ✅ 259/259 (258 + your 1) |
| `addTask` re-reads from disk, idempotent by id | ✅ confirmed on `main` |

## The overlap with your own #182

Your first commit here **is** #182 in its entirety (`fix(hive): atomically write registry.json and drain cursor.json`). Neither PR mentions the other. Whoever merges needs to know, or one of them turns into a confusing empty PR or a conflict.

I've written the same note on #182 — pick one:
1. Set this PR's base to #182's branch and say "stacked on #182, review the top commit only", or
2. Close #182 and keep this one as a clean two-commit PR.

Either is fine. I'd take (2).

## The bigger question this raises

While reviewing this I went looking for other callers of the unsafe pattern, and **`src/main/realtimeActions.ts` still has four**: lines 361, 428, 447 and 498 all call `hiveWriteTasks(...)` with a whole array. Those are the voice / Slack / realtime paths — i.e. precisely the *other* writers your race is against. So this PR closes the webhook side of the race while the counterparties remain open.

That is emphatically **not** a request to expand this PR — it's well-scoped and should stay that way. But it's worth its own issue, and you're the right person to file it since you already understand the failure mode. The clean end state is probably that `writeTasks` stops being reachable from callers who only want to add or patch one card.

## Questions

@TTAWDTT — (1) which option for the #181/#182 stack? (2) Would you file the `realtimeActions.ts` follow-up? (3) Did you hit this live — an actual webhook card that vanished — or find it by reading? If it was live, please add that to the description; a real incident is what moves a fix up the queue.

@chaitanyagiri — worth deciding whether `writeTasks` should become private, or take a merge callback, so the unsafe path stops being available. Three separate PRs in this queue (#149, #181, and the `realtimeActions` sites above) are all variations on "someone used the whole-ledger write". That's a design smell rather than three coding mistakes.

TTAWDTT:
@chaitanyagiri Thanks for the review — all three points answered:

1. #181/#182 overlap — handled as you suggested in option (2): #182 is closed, and this stays as a clean two-commit PR. The atomic-writes change ("atomically write registry.json and drain cursor.json") remains fully contained as the first commit here.

2. realtimeActions.ts follow-up — agreed, I'll file the issue. I'll list the four remaining whole-array hiveWriteTasks call sites (lines 361/428/447/498) and frame the end state you described: making writeTasks private, or giving it a merge callback, so the unsafe path stops being reachable. I'll link it back to this PR.

3. Live vs. by-reading — this was live, not found by reading. A webhook card was created and then silently vanished while a concurrently added Slack card survived; no error, no log, the card just never existed. I've added the incident to the PR description.

PR #182, opened 2026-08-19
fix(hive): atomically write registry.json and drain cursor.json

TTAWDTT:
## Problem

Four hive JSON writes used bare `writeFileSync` (`this.writeJson`), which **truncates the target then writes** — so a torn / interrupted write (process killed mid-write, disk error) can leave the file truncated or corrupt:

- `registry.json` on agent **spawn** (`register`), **archive** (`setArchived`), and **session** capture (`recordSession`) — a corrupt registry loses every agent's registration/session state.
- `cursor.json` in `drainForStop` (per-drain processed-message cursor) — a corrupt cursor could re-deliver already-processed messages on restart.

The project already owns the safe primitive — `this.atomicWriteJson` (write temp + rename), used for inbox message writes — but these four call sites bypass it.

## Change

Route all four writes through `atomicWriteJson` (temp-file + atomic rename). Behavior-preserving; same payload, same commit, same best-effort error handling.

## Test

- `node --test test/hive-task-mutation.test.cjs test/webhook-endpoints.test.cjs test/voice-messages.test.cjs` → 18/18 pass
- `npm run typecheck` → clean
- No behavior change to the written data — only the write mechanism (atomic rename instead of in-place truncate).

### DIFF
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -610,7 +610,7 @@ export class HiveManager {
       lastSeen: Date.now()
     };
     if (meta.isGod) reg.godId = meta.id;
-    this.writeJson(join(root, 'registry.json'), reg);
+    this.atomicWriteJson(join(root, 'registry.json'), reg);
 
     this.appendLog({ kind: 'spawn', agentId: meta.id, name: meta.name, isGod: !!meta.isGod });
     // Only logs on an invalid cwd (rare) — not a per-spawn line, so no log spam.
@@ -808,7 +808,7 @@ export class HiveManager {
       if (!agent || agent.archived === archived) return;
       agent.archived = archived;
       agent.lastSeen = Date.now();
-      this.writeJson(join(root, 'registry.json'), reg);
+      this.atomicWriteJson(join(root, 'registry.json'), reg);
       this.appendLog({ kind: 'archive', agentId: id, archived });
       this.commit(`hive: ${archived ? 'archive' : 'unarchive'} ${id}`);
     } catch { /* best-effort — never crash a lifecycle handler */ }
@@ -830,7 +830,7 @@ export class HiveManager {
       if (!agent || agent.sessionId === sessionId) return; // unknown agent or unchanged → no write
       agent.sessionId = sessionId;
       agent.lastSeen = Date.now();
-      this.writeJson(join(root, 'registry.json'), reg);
+      this.atomicWriteJson(join(root, 'registry.json'), reg);
       this.appendLog({ kind: 'session', agentId, sessionId });
       this.commit(`hive: session ${agentId}`);
     } catch { /* best-effort — never crash a hook handler */ }
@@ -1037,7 +1037,7 @@ export class HiveManager {
     if (fresh.length === 0) return { block: false };
 
     cursor.lastProcessed = fresh[fresh.length - 1].id;
-    this.writeJson(cursorPath, cursor);
+    this.atomicWriteJson(cursorPath, cursor);
     this.appendLog({ kind: 'drain', agentId, count: fresh.length });
 
     const lines = fresh.map((m) => `- [from ${m.from}, ${m.act}] ${m.subject}: ${m.body}`).join('\n');


chaitanyagiri:
## TL;DR

Thanks @TTAWDTT — this is a good catch and a clean fix, but **please read the last section first: this PR overlaps your own #181.**

**Context:** the hive keeps its state in JSON files on disk — `registry.json` is the list of every agent that has ever spawned (plus its session id and archived flag), and `cursor.json` records how far an agent has drained its inbox. `HiveManager` has two ways to write JSON: `writeJson`, which truncates the file and then writes it, and `atomicWriteJson`, which writes to a temp file and renames it into place. A rename is atomic on a single filesystem; a truncate-then-write is not, so a process killed mid-write leaves a half-written file.

**What this changes:** four call sites move from the unsafe primitive to the safe one that already exists in the same class. No new code, no behaviour change on the happy path.

**What it protects:** a corrupt `registry.json` loses every agent's registration and session id — which is what "Restart & Continue" reads. A corrupt `cursor.json` re-delivers already-handled mail.

**Verdict:** correct and worth taking. Typecheck clean, 258/258, merges cleanly, 2 commits behind `main`. Two asks: a test, and sorting out the relationship with #181.

---

## What I did

Read the diff, confirmed `atomicWriteJson` exists on `main` and does what the name says, checked every remaining `writeJson` call site to see whether the four you picked are the right four, ran the type-checker and full suite on your branch, and tested the merge.

## Why the fix is the right shape

`atomicWriteJson` is already in `HiveManager` (`src/main/hive.ts:1944`) and already used for inbox message writes. So this isn't introducing a pattern — it's applying one the file already committed to. That's the cheapest kind of fix to review and the easiest to defend.

```
   writeJson (before)                    atomicWriteJson (after)

   open registry.json, TRUNCATE          write registry.json.tmp
        │                                      │
        │  ← crash here and the file is        │  ← crash here and registry.json
        │    empty or half a JSON object       │    is still the last good copy
        ▼                                      ▼
   write bytes                           rename(tmp → registry.json)   [atomic]
```

The three `register` / `setArchived` / `recordSession` sites matter most, because they run on every spawn, every archive and every session capture — so they are the writes most likely to be in flight when the app is quit or the machine sleeps. The `drainForStop` cursor write matters for a different reason: it's on the hook path, which runs constantly.

## What I tested

| Check | Result |
|---|---|
| `node --test test/*.test.cjs` | ✅ 258/258 |
| `atomicWriteJson` exists and is temp+rename | ✅ `hive.ts:1944` |
| Same payload, same `appendLog`, same `commit` after the change | ✅ behaviour-preserving |

## The overlap you need to resolve

**#181 contains this PR.** I checked the commits:

- **#182** (`fix/hive-atomic-writes`) — one commit: *"fix(hive): atomically write registry.json and drain cursor.json"*
- **#181** (`fix/webhook-addtask`) — two commits: the **same commit**, then the webhook change on top.

Neither PR says so. That's a real hazard for whoever merges: take #181 first and #182 becomes an empty PR that still looks open and reviewable; take #182 first and #181 needs a rebase before it will go in. Someone merging quickly could also resolve a conflict by keeping one copy of a change that exists twice.

This is easy to fix and it's yours to choose:

1. **Keep both, label the stack** — edit #181's description to say "stacked on #182, review the top commit only", and set #181's base branch to #182's branch so GitHub shows only the new commit. Cleanest if you want the atomic-writes change reviewed on its own.
2. **Close #182, keep #181** — one PR, two commits, both reviewable. Fewer moving parts.

I'd take (2) unless you specifically want the atomic-write change to land first on its own.

## What's missing

**A test.** You describe this as behaviour-preserving, and it is on the happy path — but the whole point is the *unhappy* path, and nothing currently proves the four sites changed. A source-level assertion in the style you already used in #181 works fine here:

```js
test('registry and cursor writes go through atomicWriteJson', () => {
  const src = fs.readFileSync('src/main/hive.ts', 'utf8');
  assert.equal(/this\.writeJson\(join\(root, 'registry\.json'\)/.test(src), false);
  assert.equal(/this\.writeJson\(cursorPath/.test(src), false);
});
```

Not elegant, but it pins the fix, and it's the same technique your #181 test uses — so it's consistent with what you've already shown you're happy to write.

## Questions

@TTAWDTT — (1) which of the two options above do you want for the #181/#182 overlap? Happy either way, I just don't want to merge one and silently strand the other. (2) Did you hit an actual corrupt `registry.json`, or did you find this by reading? Either is a fine answer — but if you hit it, that's worth putting in the description, because a real recovery story is what makes a maintainer prioritise a fix like this. (3) Are there other `writeJson` call sites you deliberately left alone, and if so why? I'd rather the reasoning was recorded than rediscovered.

@chaitanyagiri — a small structural point: `writeJson` and `atomicWriteJson` sitting side by side in one class, with no rule about which to use, is how this happened. Worth considering whether `writeJson` should just *become* atomic, so the unsafe option stops existing. That would make this whole class of PR unnecessary.

TTAWDTT:
Closing as a duplicate of #181. This PR's single commit (`8cf185f`) is byte-identical to the first commit of #181 (`fix/webhook-addtask`), which carries both the atomic-writes change (registry.json + drain cursor.json) and the webhook addTask dispatch fix as a clean two-commit stack. Keeping #181 per the maintainer's recommendation — the atomic-writes change remains fully contained there.

PR #184, opened 2026-08-19
fix(control): cap the per-agent steer queue to bound memory on stalled agents

TTAWDTT:
## Summary

`ControlRegistry.steer()` pushes a note onto a per-agent queue (`steerQueue`) that is drained one-at-a-time by `takeSteer()` on each hook boundary. A halted/stalled agent never drains, so a burst of steers (a closing-time loop, a stuck caller, or repeated operator input) grows the queue in memory **without bound**.

This caps the queue at `MAX_PENDING_STEERS` (20) and drops the **oldest** note when full, so the most recent instruction still reaches the next hook boundary. Behavior is unchanged for the normal case (≤20 pending steers).

## Changes

- `src/main/control.ts` — add `MAX_PENDING_STEERS = 20`; in `steer()`, drop the oldest queue entry when at capacity before pushing the new note.
- `test/control.test.cjs` — regression tests: the queue caps at 20 with FIFO drop-oldest (survivors start at `note 6`, newest `note 25` delivered last), and whitespace-only steers are ignored.

## Verification

- `npm run typecheck` (node + web): pass
- `node --test test/control.test.cjs`: 4/4 pass

### DIFF
--- a/src/main/control.ts
+++ b/src/main/control.ts
@@ -17,6 +17,13 @@
  * Runs in the Electron main process; no electron import (unit-testable).
  */
 
+/** Max steer notes queued per agent before the OLDEST is dropped. Each note
+ *  rides the next hook's additionalContext; a stalled/halted agent never drains
+ *  the queue, so without a cap a burst of steers (closing-time loop, a stuck
+ *  caller) would grow memory forever. The latest instruction wins: when full we
+ *  drop from the front (FIFO) so a busy agent still hears the most recent note. */
+const MAX_PENDING_STEERS = 20;
+
 export interface AgentControlSnapshot {
   paused: boolean;
   halted: boolean;
@@ -70,7 +77,10 @@ export class ControlRegistry {
   }
   steer(id: string, text: string): void {
     const t = text.trim();
-    if (t) this.ensure(id).steerQueue.push(t.slice(0, 10000)); // hook additionalContext cap
+    if (!t) return;
+    const q = this.ensure(id).steerQueue;
+    if (q.length >= MAX_PENDING_STEERS) q.shift(); // keep the newest note, drop the oldest
+    q.push(t.slice(0, 10000)); // hook additionalContext cap
   }
   /** Request a graceful stop at the next hook boundary. */
   halt(id: string): void { this.ensure(id).halted = true; }
--- a/src/main/control.ts
+++ b/src/main/control.ts
@@ -79,7 +79,13 @@ export class ControlRegistry {
     const t = text.trim();
     if (!t) return;
     const q = this.ensure(id).steerQueue;
-    if (q.length >= MAX_PENDING_STEERS) q.shift(); // keep the newest note, drop the oldest
+    if (q.length >= MAX_PENDING_STEERS) {
+      // Drop the oldest so the newest instruction still reaches the next hook
+      // boundary. Log a breadcrumb so a note silently falling off the front is
+      // diagnosable — a full queue means the agent has been unreachable for a long time.
+      console.warn(`[control] ${id}: steer queue full (${MAX_PENDING_STEERS}) — dropping oldest note`);
+      q.shift(); // keep the newest note, drop the oldest
+    }
     q.push(t.slice(0, 10000)); // hook additionalContext cap
   }
   /** Request a graceful stop at the next hook boundary. */
--- a/test/control.test.cjs
+++ b/test/control.test.cjs
@@ -30,3 +30,28 @@ test('persisted delivery pauses replace stale in-memory state', () => {
   assert.equal(control.isAutoDeliveryPaused('dev2'), true);
   assert.equal(control.isAutoDeliveryPaused('dev3'), true);
 });
+
+test('steer queue is capped so a stalled agent cannot accumulate unbounded notes', () => {
+  const control = new ControlRegistry();
+  for (let i = 1; i <= 25; i++) control.steer('dev9', `note ${i}`);
+
+  // Only the cap is retained, never all 25.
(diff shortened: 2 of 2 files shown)


chaitanyagiri:
## TL;DR

Thanks @TTAWDTT — clean, well-scoped, with a test that actually proves the behaviour. This is the easiest PR in the current queue to say yes to.

**Context:** `ControlRegistry` (`src/main/control.ts`) is how the operator and the orchestrator steer a running agent without typing into its terminal. A "steer" is a short note that gets queued per agent and delivered on the agent's next hook boundary — one note at a time, via `takeSteer()`. The queue is a plain in-memory array.

**The bug:** delivery only happens when the agent *reaches* a hook boundary. An agent that is halted, paused, or simply wedged never reaches one, so it never drains. Meanwhile anything can keep calling `steer()` — a loop, a stuck caller, an operator hammering the control. The array grows for as long as the process lives, with nothing ever taking from it.

**The fix:** cap the queue at 20 and drop the *oldest* when full, so the newest instruction is the one that survives to the next boundary.

**Verdict:** correct, and the drop-oldest choice is the right one. Typecheck clean, 260/260 (2 new), merges cleanly, 2 commits behind `main`. One design question below, nothing blocking.

---

## What I did

Read the diff, read `steer()` and `takeSteer()` on `main` to confirm the drain really is one-at-a-time and boundary-gated, checked out your branch, ran the type-checker and full suite, ran your two new tests specifically, and tested the merge against current `main`.

## Why drop-oldest is the right call

This is the part of the PR I'd have argued about if you'd gone the other way, so it's worth saying plainly that you got it right.

```
   queue is full (20)          drop NEWEST                drop OLDEST (this PR)

   [n1 n2 … n20]               [n1 n2 … n20]              [n2 n3 … n20 n21]
   steer("n21") arrives         n21 discarded              n1 discarded
        │                            │                          │
        ▼                            ▼                          ▼
   next boundary delivers…      n1 — the STALEST note      n2, then eventually n21
                                and n21 never arrives      — the latest note lands
```

A steer is an *instruction*. If the queue is saturated it's because the agent is stuck, and the most recent instruction is overwhelmingly the one the operator meant — very likely "stop" or "do this instead". Discarding the newest would mean the agent, once unstuck, acts on advice that is 20 notes out of date and never hears the correction. Dropping from the front is the only version that degrades sensibly.

Your comment says this, which is why I didn't have to guess:

> *The latest instruction wins: when full we drop from the front (FIFO) so a busy agent still hears the most recent note.*

## What I tested

| Check | Result |
|---|---|
| `node --test test/*.test.cjs` | ✅ 260/260 (258 on `main` + your 2) |
| Your cap test — survivors start at `note 6`, newest `note 25` delivered last | ✅ passes, and asserts the *order*, not just the length |
| Whitespace-only steer still ignored | ✅ covered |
| Behaviour unchanged at ≤20 pending | ✅ |

The test is the reason this review is short. It asserts survivors *and* delivery order, so it would fail if someone "fixed" the cap by dropping from the back — which is exactly the mistake the test needs to catch.

## One design question, not a blocker

**Nothing tells anyone a note was dropped.** `q.shift()` is silent, so an operator whose instruction fell off the front has no way to know, and neither does anyone reading a bug report later. The whole scenario is already abnormal — a queue at 20 means an agent has been unreachable for a long time — so it's worth a breadcrumb:

```ts
if (q.length >= MAX_PENDING_STEERS) {
  console.warn(`[control] ${id}: steer queue full (${MAX_PENDING_STEERS}) — dropping oldest note`);
  q.shift();
}
```

One line, fires only in the abnormal case, and turns "my steer never arrived" from a mystery into a log line. I'd take it; your call.

Secondary, and genuinely just a question: is 20 the right number? A note is capped at 10,000 chars, so the worst case is ~200KB per agent — trivial memory-wise. The cap is really about *usefulness*, not memory: if 20 notes are pending, notes 1–15 are probably meaningless by the time they land. I'd be equally happy with 5. Not asking you to change it, just curious whether 20 came from somewhere.

## Questions

@TTAWDTT — (1) any objection to the `console.warn`? (2) Did you observe an unbounded queue in practice, or find it by reading `control.ts`? If it happened live, the trigger is worth naming in the description — a "closing-time loop" that can emit unbounded steers might itself be a bug worth its own fix, and this PR would then be treating the symptom.

@chaitanyagiri — no decision needed. Worth noting this is the third PR from @TTAWDTT in this queue (#181, #182, #184) and all three are small, single-purpose, and independently reviewable. That is exactly the shape we want more of.

TTAWDTT:
Thanks for the thorough review — and for defending the drop-oldest choice so clearly. I've taken the console.warn suggestion into the branch (fc23e76): the drop path now logs [control] <id>: steer queue full (20) — dropping oldest note, firing only in the abnormal full-queue case, so a steer that never arrives is a log line rather than a mystery. Typecheck clean, control.test.cjs 4/4 still green.

To your two questions:

- On your question (1) — objection to the console.warn? None at all — adopted as above.

- On your question (2) — observed live, or found by reading control.ts? Reading, not a live incident. This branch is a defensive guard against a latent growth path, and I'd rather cap it before it bites than after. On the "closing-time loop emitting unbounded steers" angle you raised: I don't currently have evidence of a live loop, and I deliberately kept this PR to the memory bound so the cap is reviewable on its own. If you'd rather I scope a follow-up to hunt the source of repeated steers (where they're produced, and whether any caller can loop), say the word and I'll open it — but I didn't want to fold that into this already-clean cap.

- On the right value: 20 was a round guard, and your 5 suggestion is justifiable — with a note capped at 10KB the worst case is ~200KB/agent either way, so it's about usefulness, not memory. I'm leaving it at 20 unless you feel strongly.

PR #185, opened 2026-08-19
fix(hive): bounce and log mail to an id with no inbox instead of dropping it

### LINKED ISSUE #183: Hive router silently discards messages to unknown recipients, and logs them as delivered
**Version:** 0.4.4 (macOS universal, prebuilt DMG). Code paths verified against current `main` — `src/main/hive.ts` is unchanged since 2026-08-17, so line numbers below match HEAD.

## Summary
When a message's `to` does not match an existing agent directory, `deliver()` returns without writing, logging, or bouncing. The message is destroyed, while `log.jsonl` records it as routed and the sender's outbox archives it to `.sent/`. Every observability surface reports success.

## Repro
1. In any agent's `outbox/`, write a message whose `to` is not a registered agent id (e.g. a display name like `dwight` instead of `dwight-mszxlim0`, or a non-agent label like `scheduler`).
2. Wait for the router poll (~1.5 s).

**Expected:** the sender learns it failed — a bounce to god (as already happens for provider-mismatch), a `kind:"drop"` entry in `log.jsonl` (as already happens for hop-cap), or both.

**Actual:** the file is renamed into `outbox/.sent/` and counted by `routed++`; `log.jsonl` gains `{"kind":"message", ...}` as though delivered; the commit message says `hive: routed N message(s)`. Nothing is written to any inbox. No error anywhere.

## Cause
- `src/main/hive.ts` L1168–1172 — `deliver()`: `if (!existsSync(inbox)) return; // unknown recipient — dropped (logged by caller)`. **The caller does not log it.**
- L1256 — `this.appendLog({ kind: 'message', ... })` runs unconditionally after the target loop, with no per-target delivery status.
- L1194 — `resolveTo` maps only `human`/`god`; any other value is passed through unresolved. (Not itself a defect — the documented contract in the generated PROTOCOL.md is `"to": "<agent-id>"` — but it is how an unknown id reaches `deliver()`.)
- L1228 — the `!canReceiveInbox(...)` bounce is unreachable for an unknown id: `reg.agents[t]` is `undefined`, so `canReceiveInbox(undefined)` falls back to the claude preset and returns true. The unknown-recipient case is the only delivery failure with no bounce and no log.

## Impact
On our floor this destroyed 14 of 32 messages (44%) before anyone noticed, including five task dispatches to agents that consequently never received work and nine scheduled standup replies. The orchestrator could not distinguish "agent ignored the task" from "the task never arrived", because the log said delivered. Yes, the trigger was our own mis-addressing — but the silence is what made it undetectable for hours.

## Suggested fix
Make `deliver()` report failure — return a boolean and have `routeMessage` bounce to god plus `appendLog({ kind: 'drop', reason: 'no-inbox', ... })`, matching the existing hop-cap drop and undeliverable-provider patterns. Logging resolved-and-delivered targets at L1256 rather than intent would also make `log.jsonl` trustworthy.

We're happy to submit a PR implementing the above if useful.

hyperstream-pro:
Closes #183

## What & why

Fixes #183 — the hive router destroyed any message addressed to an id with no
agent directory, while every surface reported success.

`deliver()` returned early when `agents/<to>/inbox` did not exist. Its own
comment said the drop was `logged by caller` — the caller never logged it — so
`routeOnce` still archived the outbox file to `.sent/` and still counted it in
`routed++`, and `log.jsonl` still recorded `kind:"message"`. On our floor that
silently destroyed 14 of 32 messages, including five task dispatches, and the
orchestrator could not tell "the agent ignored it" from "it never arrived".

Unknown-recipient was the **only** delivery failure with neither a bounce nor a
log. The prep-assistant and hookless-provider cases already bounce to god; the
hop-cap case already writes `kind:"drop"`. It slips past both because
`reg.agents[t]` is `undefined` for an unknown id, so `canReceiveInbox(undefined)`
falls back to the claude preset and returns `true`.

The change follows the patterns already in `hive.ts`:

- `deliver()` returns whether it wrote. The no-inbox case now appends
  `{ kind: 'drop', reason: 'no-inbox', from, to, id }` — the shape the hop-cap
  drop already uses — and bounces the message to god with an
  `[undeliverable — no agent "<id>"…]` subject, mirroring the two existing
  undeliverable bounces.
- the `kind:"message"` entry gains `delivered`: the targets that actually took
  delivery, rather than the sender's intent. `terminal-handoff` already logs a
  field by that name.

Behaviour on the happy path is unchanged, and a broadcast still reaches every
live agent when one roster entry is stale. I deliberately left the two adjacent
items from the issue thread alone to keep this focused: `resolveTo` still maps
only `god`/`human` (the documented contract is the agent id, so that is not a
defect), and `routeOnce` still renames to `.sent/` rather than a `.failed/`
lane — that one is a feature request, and the drop log now makes the failure
visible without it.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

## Screenshots / clips

Not applicable — no visual change; this is main-process routing only.

## Tests

`test/hive-unknown-recipient.test.cjs`, in the style of the existing
`test/hive-*.test.cjs` files (`node:test` + `load-ts.cjs`), added to the
`test:focused` list. Three cases: the bounce and drop log, `delivered` reporting
on both a good and a lost message, and a broadcast surviving a stale roster
entry. **All three fail on `main` and pass with this patch** (verified by
stashing the `hive.ts` change and re-running).

CONTRIBUTING.md says there is no test suite yet — `package.json` now has one, so
I followed the code rather than the doc. Happy to drop the test if you would
rather keep the fix to a single file.

## Verification

Ran on macOS 15 (Darwin 25.2.0), Node 20:

- `npm run typecheck` — passes (node + web); this is the CI gate.
- `npm run build` — succeeds, including `copy:main-assets`.
- `npm run test:focused` — 242/242 pass, including the 3 new cases.

What I could **not** verify: I have not run the fix inside a live Electron
session. `npm install --ignore-scripts` was needed here (no C/C++ toolchain for
the `node-pty` rebuild), and relaunching the app would have terminated the agent
sessions this was found on. So the routing logic is covered by the tests and the
type-checker, but not by a manual end-to-end run in the UI.

Discord:


chaitanyagiri:
## TL;DR

Thanks @hyperstream-pro — this is the best-diagnosed PR in the current queue. You didn't just find a dropped message, you found the one delivery failure in the whole router that had *neither* a bounce nor a log, and you proved it with numbers from a real floor.

**Context:** agents talk to each other by writing a JSON file into `agents/<recipient>/inbox/`. `HiveManager.deliver()` does that write; `routeOnce()` walks outboxes and calls it. Every other way a message can fail already tells someone: mail to the send-only prep assistant bounces to god, mail to a hookless-provider agent bounces to god, and a message that exceeds the hop cap gets a `kind:"drop"` line in `log.jsonl`.

**The bug:** if the recipient id has no agent directory, `deliver()` returned early and did nothing. Its own comment claimed the drop was "logged by caller" — the caller never logged it. So `routeOnce` still moved the outbox file to `.sent/`, still counted it in `routed++`, and `log.jsonl` still recorded `kind:"message"`. Every surface reported success for a message that ceased to exist. On your floor that destroyed **14 of 32 messages**, five of them task dispatches.

**The fix:** `deliver()` returns a boolean; unknown recipients get a `kind:"drop"` log line and a bounce to god, matching what the other two failure modes already do.

**Verdict:** correct, well-tested, merge-ready. Typecheck clean, 261/261 (3 new), merges cleanly, 1 commit behind `main`. One edge case worth a line of code — details below.

---

## What I did

Read the diff, read `deliver()` and `routeOnce()` on `main` to confirm the drop really was silent on every surface, checked the two existing bounce paths to see whether your new one matches their shape, checked out the branch, ran the type-checker and full suite, ran your new tests, and tested the merge.

## Why the "logged by caller" comment is the heart of this

The old code:

```ts
private deliver(msg: HiveMessage, toId: string): void {
  const inbox = join(this.agentDir(toId), 'inbox');
  if (!existsSync(inbox)) return; // unknown recipient — dropped (logged by caller)
  ...
}
```

That comment is a promise the codebase never kept, and it's *worse* than no comment: anyone auditing the router reads it, believes the drop is accounted for somewhere, and moves on. I did exactly that reasoning myself before checking, which is how I know it works.

The failure mode it produces is the expensive one — not "a message was lost", but **"the orchestrator cannot tell 'the agent ignored my instruction' from 'my instruction never arrived'"**. Every recovery behaviour built on top of that distinction is broken as long as it's indistinguishable.

```
   BEFORE                                  AFTER

   send → route → deliver()                send → route → deliver() → false
                    │ no inbox                                          │
                    └─ return (silent)                                  ├─ log kind:"drop" reason:"no-inbox"
   outbox → .sent/  ✅                                                   └─ bounce to god
   routed++         ✅                     outbox → .sent/  ✅ (delivery was attempted)
   log kind:"message" ✅                   log now reports who ACTUALLY took it
        └── all three lie
```

I also like that you tracked `delivered[]` and made the log report actual delivery rather than intent. That's the part that stops the same class of bug recurring for a *different* reason later — the log is now derived from outcomes, not from the loop's input.

## What I tested

| Check | Result |
|---|---|
| `node --test test/*.test.cjs` | ✅ 261/261 (258 + your 3) |
| `test/hive-unknown-recipient.test.cjs` alone | ✅ passes |
| The other two bounce paths are shaped the same way | ✅ your branch matches them |
| `routeOnce` still archives to `.sent/` after a bounce | ✅ correct — delivery *was* attempted |
| Merge against current `main` | ✅ clean, 1 commit behind |

## The edge case I'd close

Your bounce does this:

```ts
if (t !== godId) {
  this.deliver({ ...msg, to: godId, subject: `[undeliverable …] ${msg.subject}` }, godId);
}
```

That call's return value is discarded — so **if god's own inbox is missing, the bounce vanishes exactly the way the original message did.** That is not hypothetical: a hive whose god directory was never created, or was cleaned up, is precisely the state where ids stop resolving in the first place, so the two failures are correlated rather than independent.

One line fixes it:

```ts
if (t !== godId && !this.deliver({ ...msg, to: godId, subject: … }, godId)) {
  this.appendLog({ kind: 'drop', reason: 'no-god-inbox', from: msg.from, to: t, id: msg.id });
  console.warn(`[hive] bounce for "${t}" could not reach god — no inbox at ${godId}`);
}
```

The log line already exists as a concept in your diff; this just applies your own rule one level up. Worth adding — the whole point of the PR is that a silent drop is the unacceptable outcome, and there's currently one left.

Second, much smaller: the same "check the return value" argument applies to the two *pre-existing* bounce sites you touched (the hookless and proxy-tier paths). Those were already discarding it before your PR, so it's not a regression you introduced and I'm not asking you to fix them here — but it's the same latent hole, and it'd be a natural follow-up.

## Questions

@hyperstream-pro — (1) any objection to the god-inbox guard above? Happy for it to be a follow-up if you'd rather freeze this one. (2) Do you know *why* those 14 messages were addressed to ids with no directory — stale ids from a previous run, a typo in god's routing, or agents archived mid-flight? The fix is right either way, but the cause might be a second bug, and you're the only person with that floor's `log.jsonl`. (3) You mention issue #183 — is that yours? If so it should be linked with `Closes #183` in the description so it auto-closes.

@chaitanyagiri — one for you: `deliver()` returning `void` is what made this possible, and the same shape exists elsewhere in the router. Worth a rule that every hive delivery primitive returns success and every caller is required to handle it — otherwise this class of bug is one refactor away from returning.

hyperstream-pro:
Thanks — taking the three in order.

### (1) The god-inbox guard — no objection, you're right

It's this PR's own rule applied one level up, and the correlation argument is the part that
settles it: a hive whose god directory is missing is *exactly* the state where ids stop
resolving, so the two failures aren't independent. Leaving one silent drop in a PR whose
thesis is "a silent drop is the unacceptable outcome" would be odd.

Your line is what I'd write. Say the word and I'll push it here with a test (the existing
`floor()` helper makes a no-god-inbox case cheap), or take it as a follow-up together with
the two pre-existing bounce sites — your call on whether you'd rather freeze this one.

### (2) Why those ids had no directory — not stale, not archived mid-flight

Checked against that floor's `log.jsonl` and `registry.json`. It's **two distinct
mis-addressing patterns, both ours**, and neither is the second bug you were reasonably
worried about:

| `to` | count | what it actually was |
|---|---|---|
| `scheduler` | 14 | **never an agent** — a role label for hourly standups |
| `breaker` | 2 | **never an agent** — the circuit breaker is a sender, never a recipient |
| `toby`, `dwight`, `pam`, `stanley`, `angela` | 1 each | **display name instead of the agent id** — each maps to a currently-registered agent |

So: no stale ids from a previous run (every bare name resolves to an agent that is registered
*right now*), and nothing archived mid-flight. The bulk was mail addressed to a *concept* that
was never registered, and the rest was our orchestrator writing `dwight` where the contract
says `dwight-mszxodmp`.

Worth adding that your own docs already told us: the generated `PROTOCOL.md` specifies
`"to": "<agent-id>"`, and the live roster is auto-injected with full ids. We had the ids in
context and used names anyway. That's on us, not on the router — the router's fault was only
that it said nothing.

One corroboration for the diagnosis: the rate collapsed once we started using full ids. At the
time I opened this PR it was 14 of 32 messages (44%); that log is now 583 messages with 21
orphan-addressed, and the newer 7 are the `scheduler`/`breaker` labels rather than name-typos.
If the cause had been staleness or an archive race, growth wouldn't have fixed it.

That does suggest a cheap follow-up in the same spirit as this PR: validate `to` against the
registry at *send* time and refuse loudly, so the sender learns immediately rather than the
orchestrator learning hours later. Happy to open that separately if useful — I've kept it out
of here to stay single-concern.

### (3) `Closes #183`

Yes, #183 is mine. Added to the description just now.

And noted on `deliver()` returning `void` — agreed that's the root shape. Every delivery
primitive returning success, with callers required to handle it, would close this class rather
than this instance.

PR #186, opened 2026-08-19
test(hooks): cover Notifications/Stop idle-detection branches

BUGHUNTER-SACHIN:
## Context

Issue #104 asks contributors to help understand/improve "the bridge" — how
each CLI's terminal output gets turned into structured events the harness
acts on. For Claude Code, that's `HookServer` in `src/main/hooks.ts`.

## What this adds

Test coverage for the two branches that call `notify()`:
- `Notification` event: fires only when `notification_type === 'idle'`,
  or (fallback for payloads without the structured field) when `message`
  case-insensitively contains "waiting for your input"
- `Stop`/`SubagentStop`: always notifies "finished — idle" unless
  `stop_hook_active` is set (already re-entered boundary)

Neither branch had any existing test coverage.

## Testing

- `node --test test/hooks-notification.test.cjs` — 8/8 pass
- `node --test test/*.test.cjs` — full suite still green (11 pre-existing
  failures on `main`, all Windows path-separator issues in unrelated files:
  `cli-install-ladder`, `codex-remote`, `expand-tilde`,
  `transcript-project-dir` — confirmed via `git status` that my change
  touches only the new test file)
- `npm run typecheck` — clean

### DIFF
--- a/test/hooks-notification.test.cjs
+++ b/test/hooks-notification.test.cjs
@@ -1,0 +1,114 @@
+'use strict';
+/**
+ * HookServer's Notification/Stop branches are what tell the harness an agent is
+ * idle and safe to deliver queued mail to. Both branches call notify() but
+ * neither had test coverage:
+ *   - Stop/SubagentStop always notifies "finished — idle" (unless stop_hook_active)
+ *   - Notification only notifies when notification_type === 'idle' OR the message
+ *     text contains "waiting for your input" (case-insensitive) — this string
+ *     fallback exists for CLI versions/locales that don't send the structured
+ *     notification_type field, so both paths need to be proven independently.
+ */
+const test = require('node:test');
+const assert = require('node:assert/strict');
+const fs = require('node:fs');
+const os = require('node:os');
+const path = require('node:path');
+const loadTs = require('./load-ts.cjs');
+
+const electron = require.resolve('electron');
+const notifications = [];
+require.cache[electron] = {
+  id: electron,
+  filename: electron,
+  loaded: true,
+  exports: {
+    Notification: class {
+      constructor(opts) { this.opts = opts; }
+      show() { notifications.push(this.opts); }
+      static isSupported() { return true; }
+    }
+  }
+};
+
+const { HiveManager } = loadTs('src/main/hive.ts');
+const { HookServer } = loadTs('src/main/hooks.ts');
+const CONFIG = { notifications: true };
+
+function tmpHome() {
+  return fs.mkdtempSync(path.join(os.tmpdir(), 'md-hooks-notif-'));
+}
+
+async function floor(t) {
+  const home = tmpHome();
+  t.after(() => fs.rmSync(home, { recursive: true, force: true }));
+  const hive = new HiveManager(() => home);
+  await hive.ensureAgent({ id: 'jim-1', name: 'Jim', provider: 'claude', cwd: home });
+  const server = new HookServer(hive, () => null, () => CONFIG, undefined, undefined);
+  const fire = (payload) => server.handle({ agent_id: 'jim-1', session_id: 's1', ...payload });
+  notifications.length = 0;
+  return { home, hive, server, fire };
+}
+
+test('Notification with notification_type "idle" fires a toast', async (t) => {
+  const { fire } = await floor(t);
+  await fire({ hook_event_name: 'Notification', notification_type: 'idle', message: 'anything' });
+  assert.equal(notifications.length, 1);
+  assert.equal(notifications[0].title, 'jim-1');
+});
+
+test('Notification with message "waiting for your input" fires a toast (string fallback)', async (t) => {
+  const { fire } = await floor(t);
+  await fire({ hook_event_name: 'Notification', message: 'Claude is Waiting For Your Input' });
+  assert.equal(notifications.length, 1, 'must be case-insensitive');
+});
+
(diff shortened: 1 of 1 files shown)


chaitanyagiri:
## TL;DR

Thanks @BUGHUNTER-SACHIN — a pure test PR that adds coverage where there genuinely was none. Easy yes, and your PR description also surfaced something more important than the tests themselves.

**Context:** `HookServer` in `src/main/hooks.ts` is the piece that turns Claude Code's hook events into things the app acts on. Two of those events decide when an agent is considered *idle* — which is what drives the floor avatar, the message-queue drain, and the reaper. `Notification` fires when the CLI wants the user's attention, and `Stop` / `SubagentStop` fire when a turn ends. Both call `notify()`, and the logic for *when* is fiddly: `Notification` only counts as idle if `notification_type === 'idle'`, or, for payloads that lack that field, if the message text contains "waiting for your input". `Stop` always counts unless `stop_hook_active` says we're already inside a re-entered boundary.

**What this adds:** `test/hooks-notification.test.cjs`, 114 lines, 8 tests, covering both branches including the string fallback and the `stop_hook_active` guard. No production code touched.

**Verdict:** merge-ready as far as I'm concerned. Typecheck clean, 266/266, merges cleanly, 3 commits behind `main`. But please see the last section — the claim in your description about failures on `main` needs following up, and it may matter more than this PR.

---

## What I did

Read the test file, read the two branches in `hooks.ts` it targets, confirmed neither had existing coverage, checked out the branch, ran your file alone and then the whole suite, and tested the merge.

## Why this coverage is worth having

Idle detection is load-bearing in a way that isn't obvious from the outside:

```
   Claude Code CLI
        │  Notification / Stop hook
        ▼
   HookServer.notify()  ← the branches you just covered
        │
        ├──► floor avatar switches to idle
        ├──► the message queue drains (a queued nudge finally gets typed)
        └──► the idle reaper starts its clock on a worker
```

If the `Notification` branch wrongly returns idle, an agent gets interrupted mid-turn. If it wrongly *doesn't*, a worker sits with undrained mail — which is the failure two other open PRs (#162, #175) exist to work around from the other end. So this is not incidental logic, and "no tests at all" was a real gap.

The string fallback is the branch I'm most glad you covered. `message.toLowerCase().includes('waiting for your input')` is exactly the kind of check that gets broken by an upstream wording change, and a test at least means we find out at CI time rather than from a wedged floor.

## What I tested

| Check | Result |
|---|---|
| `node --test test/hooks-notification.test.cjs` | ✅ 8/8 |
| `node --test test/*.test.cjs` (full suite, your branch) | ✅ 266/266 |
| Full suite on `main` @ current tip, same machine | ✅ 258/258 — **no failures** |
| `npm run typecheck` | ✅ exit 0 |
| Neither branch had prior coverage | ✅ confirmed |
| Production code touched | ✅ none — test file only |

## The thing in your description that matters more than the tests

You wrote:

> *`node --test test/*.test.cjs` — full suite still green (11 pre-existing failures on `main`, all Windows path-separator issues)*

I want to be careful here, because I think you've found something real and I don't want it lost in a test PR. **On macOS, `main` is 258/258 green** — I ran it on the current tip before reviewing you. So those 11 failures are almost certainly genuine, Windows-only, and *nobody is tracking them*: CI runs on `macos-latest` only, so Windows breakage has never once shown up in a check.

That's a more valuable finding than the coverage you're adding. Could you open a separate issue with the actual failure output — the 11 test names and one example assertion? Concretely:

```
node --test test/*.test.cjs 2>&1 | grep -A6 "not ok"
```

I'd rather that lived in its own issue than in a comment on a test PR, because it needs a decision from the maintainer (fix the tests? add a Windows CI job? declare Windows unsupported?) and it'll get lost here.

Please don't fix them in this PR. It's currently a clean single-purpose change and it should stay that way.

## Small notes on the tests themselves

Nothing blocking, two observations:

1. Your tests assert on `notify()` being called. If the idle *semantics* ever change — say a new `notification_type` value should also count as idle — these tests pin the current behaviour, which is what you want. Good.
2. Consider one negative case for a `Notification` with an unrelated `notification_type` **and** a message that happens to contain the phrase, to pin the precedence between the structured field and the string fallback. Right now I can read the intent from the code, but the tests don't force it.

## Questions

@BUGHUNTER-SACHIN — (1) will you open that Windows-failures issue? If you'd rather not, say so and I'll do it, but the output is on your machine and not mine. (2) Are you on Windows day to day? If so you're one of very few people testing this project there, and that's worth us knowing.

@chaitanyagiri — two things. This PR is fine to merge as-is. Separately: **CI has never run the test suite at all**, on any platform, so the 11 Windows failures above are not an anomaly — they're the expected outcome of a suite that no automated job executes. That's the Phase 0 item from the community audit, and this PR is a decent argument for doing it sooner rather than later.

BUGHUNTER-SACHIN:
Thanks for the detailed review!

1. Yes, happy to open the Windows-failures issue — I'll grab the full
   output with the grep command you gave and file it separately.
2. Yes, I'm on Windows day to day (Git Bash on Windows 11) — this is
   actually the only environment I've run this project in so far.

BUGHUNTER-SACHIN:
Opened the Windows test-failures issue: #253

Covers all 11 failures (cli-install-ladder, codex-remote, expand-tilde,
transcript-project-dir) with the full logs and a screenshot per the
bug-report template. Also confirmed I'm on Windows day-to-day, Git Bash,
as mentioned above.

PR #187, opened 2026-08-19
fix(hive): stop stale inbox nudges waking agents against an empty inbox

hyperstream-pro:
## What & why

Agents on our floor kept being woken with "You have new hive inbox message(s)"
and finding nothing pending. Three hit it independently; one checked immediately
**and again at +2s** and no file ever appeared. It recurred after being reported,
so it isn't a one-off.

**It is not a write/notify race.** I checked that first, because that was our
working theory: `deliver()` writes tmp-then-`renameSync` and runs *before*
`emitMessage`, and the 4s poll in `useHive` reads the real inbox over IPC before
it nudges. Delivery is already atomic-before-notify. The defect is staleness
*after* the notification.

The nudge is **queued** the moment fresh mail is seen and **typed** only once the
agent is idle, off cooldown, past boot grace and with no user draft in the way —
and it persists to `localStorage`, so it survives a reload. Mail arriving while
an agent is mid-turn queues one nudge per poll. The agent drains its *whole*
inbox on the first one and files it under `inbox/.done/`, and every nudge behind
it lands on a directory that agent just emptied. That is the repeat.

`enqueueMessage` already carries exactly this invariant for `/compact`, with a
comment that describes this bug precisely — *"idempotent in the worst way: the
first `/compact` does the work and every one behind it answers 'nothing to
compact', so a queue that accumulates them spends a delivery slot and a model
round-trip per copy to achieve nothing"*. The inbox nudge is the second command
with that property, so this puts it under the same rule, in the same place, for
the same stated reason.

- **`@shared/hiveNudge`** builds the nudge and recognises it, mirroring
  `isCompactionCommand`.
- **The nudge now names the message ids that prompted it.** Without them an agent
  cannot distinguish "I handled this last turn" from "I was woken for nothing",
  which is the ambiguity that costs the round-trip.
- **`enqueueMessage` keeps one nudge pending per agent.** Dropping the copy loses
  nothing — the surviving nudge sends the agent to the same directory, where the
  newer mail is waiting too.
- The ids are **diagnostic, not a work list**: a suppressed copy leaves its ids
  unnamed, so the text keeps the pending inbox authoritative and tells the agent
  to work everything still in it. Getting this backwards would trade a wasted
  wake for lost mail, which is why it is spelled out in the payload.

I deliberately did **not** add per-item revalidation inside the queue-drain
effect. That effect is documented as the single gate for every typed message
("One gate, one place, and this loop stops needing prompt logic of its own"), and
teaching it about one message type would undo that.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

## Screenshots / clips

Not applicable — no visual change. The only user-visible difference is the text
of the queued nudge, quoted in full above.

## Tests

`test/hive-nudge.test.cjs`, added to `test:focused`. Six cases covering the
predicate the invariant depends on: every nudge the app builds is recognised
whatever ids it carries, two nudges with *different* ids both match (matching the
whole string instead of the fixed head would never dedupe — that is the bug),
prose that merely mentions the inbox is not a nudge, and the payload keeps the
pending inbox authoritative.

**Scope of that, honestly:** this pins the predicate, not the store. It is the
same depth as your own coverage of the identical `/compact` invariant — whose
test file says "the queue's one-pending-compact invariant depends entirely on
this predicate" — because `store.ts` can't be loaded by the `node:test` harness.
The three-line store change itself is verified by inspection and the type-checker
only. I did not claim these tests fail on `main`: they exercise a new module, so
that would be trivially true and prove nothing.

## Verification

macOS 15 (Darwin 25.2.0), Node 20:

- `npm run typecheck` — passes (node + web); the CI gate.
- `npm run build` — succeeds, including `copy:main-assets`.
- `npm run test:focused` — 245/245 pass, including the 6 new cases.

**Not verified:** I have not watched this in a live Electron session. `npm install`
needed `--ignore-scripts` here (no C/C++ toolchain for the `node-pty` rebuild),
and relaunching would have killed the agent sessions the bug was observed on. So
the timing behaviour is reasoned from the code and the reports, not observed
post-fix. The nudge itself is unchanged in *when* it fires — only in what it says
and whether copies pile up — which is what keeps that gap tolerable.

Independent of #185 (branched from `main`); the two touch different files.

Discord:


hyperstream-pro:
Further field evidence, from a forensic review of one agent's transcript on v0.4.4. It
raises this from noise reduction to operator-control reliability, and it also suggests a
second path that this PR's diff does **not** touch — I've kept them separate below so you
can judge the split.

### The cadence this PR addresses

That agent received **22** "You have new hive inbox message(s)" prompts in three hours —
**eight in the final 25 minutes** (13:03, 13:05, 13:11, 13:17, 13:21, 13:22, 13:24, 13:25).
It was being restarted almost continuously. That is the pile-up this PR removes.

### The part that makes it a control problem

Verbatim `user_message` payloads and timings from the transcript:

| Time (Z) | Event |
|---|---|
| 13:24:21 | operator types **`please pause`** — agent is mid-turn |
| 13:24:29 | the turn completes |
| 13:24:33 | **a fresh inbox nudge is injected, 12s after the pause request**, starting a new task |
| 13:24:51 | operator types **`please pause`** again |
| 13:25:05 | agent replies "Paused. I won't process further inbox messages until you ask me to resume." |
| 13:25:10 | a `user_message` arrives as the literal string **`please pauseYou have new hive inbox message(s) — …`** — the operator's keystrokes and an automated nudge concatenated into one prompt; turn aborted |
| 13:25:15 | another nudge; turn aborted |

The agent had in fact complied, explicitly, 44 seconds after the first request. What made it
look uncontrollable was the injector firing between the operator's two requests and then
merging with the operator's own keystrokes. The practical cost on our side: that agent was
very nearly replaced for cause, on the strength of behaviour it did not produce.

### On the splice specifically — likely a different path from this diff

I checked before reporting this: `50e8f94` ("fix(queue): one gate for every automatic
writer, and stop guessing at the prompt", 2026-08-04) **is** an ancestor of `v0.4.4`, so this
happened *with* the one-gate queue in place, not before it. The fusion is therefore not the
pre-queue direct-write behaviour that commit retired.

We could not determine which path let it through, and I'd rather not guess in your tracker.
Two candidates we could see from the outside, offered as questions rather than claims:

- `isTerminalAutomationSafe` returns `true` when the pty has no entry in the renderer's
  terminal pool (`terminalPool.ts:296`). A terminal the renderer isn't currently holding open
  has no observable draft, so the draft gate cannot block on one.
- `automationStateOf` lets `promptLineHasText` overrule `inputDirty` to `false`, past the 1s
  echo grace. If the rendered row reads as empty while the operator has text in flight, the
  gate opens.

Either way it is an input-handling question rather than the drain race this PR fixes, so I
have **not** expanded the diff. Happy to move this into its own issue if you'd prefer it
tracked separately — say the word and I'll re-file it there.

### One unverified hypothesis, flagged as such

A second observer on our floor suggested nudges may also be triggered by an inbox *write*
rather than only by the poll. We have **not** verified that and are not asserting it; noting
it only so it isn't lost if it turns out to matter.

chaitanyagiri:
## TL;DR

Thanks @hyperstream-pro — a careful diagnosis, and the fix introduces the right primitive rather than patching the symptom. There is one interaction with other open PRs you couldn't have known about; that's the main thing to read below.

**Context:** when hive mail lands for an agent, the renderer queues a nudge — "you have new inbox message(s), go read them". The nudge is *queued* immediately but *typed* only when the agent is idle, off cooldown, past boot grace, and not sitting on a user's half-typed draft. That gap can be long, and the queue persists to `localStorage`, so it survives a reload.

**The bug:** by the time the nudge is finally typed, the agent has often already drained that mail itself and filed it under `inbox/.done/`. So it wakes up, reads an empty inbox, and burns a turn working out that nothing happened. Three agents hit this independently on your floor, and it recurred after being reported.

**What I appreciated:** you ruled out the obvious theory first and said so. Delivery is `renameSync`-atomic and runs *before* `emitMessage`, and the renderer reads the real inbox over IPC before nudging — so it isn't a write/notify race. Naming the theory you disproved is worth more than the fix.

**The fix:** name the message ids in the nudge, and move the queue's one-pending rule from string equality to a predicate, `isInboxNudge()`, in a new `src/shared/hiveNudge.ts`.

**Verdict:** correct and well-built. Typecheck clean, 264/264 (6 new), merges cleanly, 1 commit behind `main`.

---

## What I did

Read the diff, read the existing `isCompactionCommand` pattern you're mirroring, checked out the branch, ran the type-checker and full suite, ran your new tests, tested the merge, and then cross-checked how the other open inbox-wake PRs dedupe their nudges — which turned up the interaction below.

## Why a predicate, not a constant

This is the design decision that makes the PR worth taking, and it's easy to miss.

```
   ONE FIXED STRING                        THIS PR: VARIABLE TEXT + PREDICATE

   text === INBOX_NUDGE                    isInboxNudge(text)
        │                                       │
        │ works only while every nudge          │ matches the fixed head, ignores the ids
        │ is byte-identical                     │
        ▼                                       ▼
   add ids to the text →                   add ids to the text →
   dedupe silently stops matching          dedupe keeps working
   → agents get nudged twice
```

You needed the ids in the text (that's the actual fix), which means the text can no longer be a constant — so the dedupe *had* to become a predicate. Doing both in one PR is correct; doing only the first would have quietly broken deduplication.

Mirroring `isCompactionCommand` is the right call too: same problem, same shape, and a reader who knows one now knows the other.

I also want to note the care in the nudge wording itself. The text says the ids are diagnostic and the inbox directory is authoritative:

> *Your inbox directory is authoritative: work everything still pending in it, and if a named id is already in inbox/.done/ you handled it on an earlier turn and can ignore that one.*

That matters because your own one-pending rule means a suppressed nudge's ids never get named — so an agent that treated the id list as a work list would miss mail. You anticipated the failure your own dedupe creates. That's the difference between a fix and a good fix.

## What I tested

| Check | Result |
|---|---|
| `node --test test/*.test.cjs` | ✅ 264/264 (258 + your 6) |
| `test/hive-nudge.test.cjs` alone | ✅ passes |
| Negative case (a sentence merely *containing* the phrase) rejected | ✅ covered |
| One-pending rule still suppresses a second nudge | ✅ covered |
| Merge against current `main` | ✅ clean, 1 commit behind |

## The interaction you couldn't have known about

There are two other open PRs adding a **main-process** watchdog that queues this same nudge — **#162** and **#175** (they're duplicates of each other; that's a separate problem I've raised on both). Both dedupe by exact string match:

- **#162** defines `const INBOX_NUDGE = '…'` and checks `queued.some((m) => m.text === INBOX_NUDGE)`
- **#175** defines `WORKER_WAKE_NUDGE` with the same literal text and its own cooldown map

**If your PR lands with either of those, their dedupe stops working** — your nudges carry ids, so they will never be byte-equal to their constant, and an agent can be nudged by both the renderer and the watchdog for the same mail. That's the exact double-nudge those PRs were partly written to prevent.

The good news: your `isInboxNudge()` is the primitive that fixes it for all three. Whichever lands first, the others should import it rather than compare strings. Nothing for you to change here — I'm flagging it so it gets sequenced deliberately rather than discovered in a merge, and I've cross-referenced it on #162 and #175.

@chaitanyagiri — this is the ordering decision: **merge #187 first**, then require whichever watchdog PR you take to use `isInboxNudge`. The reverse order means someone has to notice the broken dedupe on their own.

## Two small things

1. `package.json +1/-1` — you rewrote the `test:focused` line. You're now the **eighth** PR to do that from a different baseline, so those eight will all conflict with each other. Not your fault; the line shouldn't exist in that form. Leave it, it'll be sorted centrally.
2. `NUDGE_HEAD` is matched with `startsWith` after `trim()`. That's right for the current call sites, but it does mean a nudge with leading content — say, a standing-goal `<goal>…</goal>` block prepended for a hookless provider (see #153) — would not be recognised. Worth a moment's thought, since #153 does exactly that prepending for Cursor agents.

## Questions

@hyperstream-pro — (1) does point 2 above actually bite? If a Cursor agent gets `<goal>` prepended to a queued nudge, `isInboxNudge` returns false and the one-pending rule stops holding for it. Worth checking whether the prepend happens before or after the queue. (2) You say three agents hit this and one checked again at +2s — do you have the `log.jsonl` window? It'd be useful to confirm the mail really was drained-then-nudged rather than never-delivered, which is the failure #185 describes.

hyperstream-pro:
Thanks — and the #162/#175 cross-reference is genuinely useful; I had no way to see that from
outside. Agreed on the ordering, and `isInboxNudge` is exactly the thing for all three to
share rather than each holding its own constant.

### (1) The #153 `<goal>` prepend — I checked, and it does **not** bite

Good catch to raise it; the answer turns on *where* the prepend happens, and in #153 as it
currently stands it is at **type time, not enqueue time**.

From #153's diff of `useHive.ts`, inside the drain's delivery closure:

```diff
           () => submitToPty(
             target.ptyId!,
-            wrap ? wrap(next) : (next.instruction ?? next.text),
+            withStandingGoal(
+              target,
+              wrap ? wrap(next) : (next.instruction ?? next.text)
+            ),
             inferAgentProvider(target.command, target.provider)
           ),
```

`withStandingGoal` wraps the *argument to `submitToPty`*. The queued item is never modified —
`next.text` still holds the bare nudge. Since the one-pending rule runs in `enqueueMessage`,
against `trimmed`, it is looking at text that has not been prepended, so `isInboxNudge` still
matches and dedupe holds for Cursor agents too. The other call site (the TUI seed) is likewise
type-time, and a seed is not a nudge.

So no change needed here, and none in #153.

**The condition that *would* make it bite**, stated so it doesn't get lost: if the prepend ever
moved to enqueue time — i.e. if `<goal>` were baked into `QueuedMessage.text` rather than
applied on the way to the PTY. Worth keeping "decorate at type time, not in the queue" as the
invariant, because it protects the compaction dedupe identically — `isCompactionCommand` reads
the leading verb, so a prepended goal block would break `/compact` dedupe the same way. If you'd
rather have belt and braces, `isInboxNudge` could skip a leading `<goal>…</goal>` block, but
I'd rather not add speculative parsing to a predicate whose whole value is being obvious.

### (2) The `log.jsonl` window — I can prove *delivered*, but not *nudged*, and here's why

Straight answer first: **`log.jsonl` contains no nudge events at all.** I grepped that floor's
whole log — the kinds present are `message`, `spawn`, `session`, `tasks`, `archive`,
`cwd_invalid`, `condense-abort`. The renderer's inbox-wake path writes nothing. So there is no
window that shows "drained, then nudged", and I'd rather say that than assemble something that
looks like one.

What I *can* rule out is the failure you're distinguishing from — never-delivered (#185). For
the instance I documented in my own session, both messages have delivery records and both files
exist under `inbox/.done/`:

```
13:27:09Z  kind=message  from=god             to=ryan-mt01w0g8  id=…-ryan-go-confirmed
13:27:32Z  kind=message  from=toby-mszxuht9   to=ryan-mt01w0g8  id=…-520c33
```

(Filesystem mtimes on the `.done` copies are *delivery* times, not filing times — `mv` preserves
mtime — so I can't timestamp the filing from disk either. Not going to present them as if I could.)

The positive evidence I do have is first-person and, I think, better than a log window would be:

- I handled both of those messages in one turn and filed them to `.done`.
- I was then woken with the nudge against an inbox I verified empty four ways: no files of any
  extension in `inbox/`, the two messages present in `.done`, no quarantined message in
  `outbox/.sent/bad-*`, and `cursor.json` clean. Nothing was lost.
- I was woken **again**, with the inbox directory's mtime unchanged — so no new mail had
  arrived in between.

That second wake is the informative one. A single stale nudge cannot fire twice: `nudged` is a
ref-held set and ids are marked seen once, so a repeat requires a *fresh enqueue*, and there was
no fresh mail to trigger one. That points at a **queue holding several nudges and draining one
per idle turn** — the pile-up this PR removes — rather than one stale nudge.

**Follow-up worth considering separately:** a `kind:"nudge"` line in `log.jsonl` when one is
queued and when it is typed. It would have made your question answerable from the log in one
grep, and right now a floor genuinely cannot audit its own notifications. I've deliberately kept
it out of this PR.

PR #207, opened 2026-08-20
fix(build): ad-hoc sign the mac app when there is no Developer ID identity

gts-47:
Without a Developer ID cert in the keychain, the `afterSign` hook returns early and the build ships with **no valid signature at all**:

```
$ codesign --verify --deep --strict "Munder Difflin.app"
Munder Difflin.app: code object is not signed at all
```

macOS will not persist a TCC grant for a bundle whose identity it cannot validate, so the folder-access prompt comes back on **every agent file touch** — handing one Desktop screenshot to an agent produced ten consecutive Allow / Don't Allow dialogs. That is precisely what the usage strings and entitlements already in `electron-builder.yml` exist to prevent; the comment there says so out loud:

> for a signed app, remembers the grant — so the user is asked ONCE instead of on every agent action

**Fix:** ad-hoc sign as a fallback when no Apple credentials are present.

An ad-hoc signature is not distributable and Gatekeeper still blocks it for other users — this changes nothing for released builds, which continue to be Developer ID signed and notarized. But it is self-consistent and carries the real bundle identifier (`in.munderdiffl.app`) instead of Electron's linker-signed `Electron`, which is what TCC needs to remember the grant on the machine that built it. Anyone running a local `npm run dist:mac` gets the one-time prompt the app was designed around.

Best-effort by design: if `codesign` is unavailable the hook warns and the build still completes, so a contributor without the command line tools is no worse off than today.

Verified locally — before / after on the same bundle:

```
before:  code object is not signed at all   Identifier=Electron
after:   (verifies clean)                   Identifier=in.munderdiffl.app  Signature=adhoc
```

https://claude.ai/code/session_01DUpadJJuEQz1zpQ6Vk83j3

### DIFF
--- a/build/notarize.cjs
+++ b/build/notarize.cjs
@@ -12,6 +12,41 @@
 const { execFileSync } = require('node:child_process');
 const path = require('node:path');
 
+/**
+ * Ad-hoc sign the bundle when there is no Developer ID identity to sign with.
+ *
+ * Without this the app ships with NO valid signature at all — `codesign
+ * --verify` reports "code object is not signed at all", and macOS will not
+ * persist a TCC grant for a bundle whose identity it cannot validate. The
+ * practical effect is the one this config exists to prevent: the folder-access
+ * prompt returns on EVERY agent file touch instead of being answered once. An
+ * ad-hoc signature is not distributable and Gatekeeper still blocks it for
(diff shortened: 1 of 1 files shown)


gts-47:
Closing — this only affects local unsigned builds. Released builds are Developer ID signed, so upstream doesn't have this problem.

PR #210, opened 2026-08-20
docs: the art licence claims are true again — Modern Interiors is bought

chaitanyagiri:
The LimeZu **Complete Version** licence for *Modern Interiors — RPG Tileset [16X16]* was purchased on **2026-08-20**. Until now, eight places across the repo and the website still told readers the art was non-commercial-only — including `docs/terms.html`, which is a legal-facing page.

## What the licence actually says

Straight from [the vendor page](https://limezu.itch.io/moderninteriors), Complete Version tier:

| | |
|---|---|
| **Can** | "Edit and use the asset in any commercial or non commercial project" |
| **Can't** | "Resell or distribute the asset to others" · "Edit and resell the asset to others" |
| **Credits** | **Required**, linking to https://limezu.itch.io/ |

The credit requirement is new to us — the free tier didn't demand it, so no file mentioned it. It's now written down in every place that could tempt someone to drop the acknowledgement (README, LICENSE, ATTRIBUTION.md, CONTRIBUTING.md, terms.html).

## Two things that were wrong regardless of the purchase

**The cast hasn't been LimeZu art for a while.** `cast.ts` says so in its own header — "The LimeZu base sheets are no longer used for the cast" — yet five files still described the Office characters as recoloured derivatives inheriting a LimeZu restriction. They're drawn procedurally in `portraitArt.ts` and carry no third-party licence at all. The old text *understated what we own.*

**Six of the ten licensed files were dead.** `A2 Office Floors.png`, `A4 Office Walls.png`, and all four `Adam/Alex/Amelia/Bob_walk.png` sheets had **zero references anywhere in the tree**. Deleted. The four tilesets that remain — `interiors`, `room-builder`, `office-tileset`, `a5-office-floors-walls` — are all genuinely loaded by the maps.

## Coverage, verified rather than assumed

Every remaining file was checked against what the maps and `themeRegistry.ts` actually load, not assumed from the folder name. `CONTRIBUTING.md` wasn't on the list of places to fix — a whole-repo sweep found it.

## Checks

- `typecheck:web` — unchanged (1 pre-existing `vite/client` error, absent `node_modules`)
- `test:focused` — 211/215, identical to baseline; the 4 failures are `Cannot find module 'electron'/'node-pty'` in a worktree without `node_modules`
- No dangling references to any deleted file


chaitanyagiri:
**Update: it's 3, not 4.** `room-builder.png` is runtime-dead too — pushed as `eb08bcf`.

It survived the first audit because it *does* have a reference: `lobby.tmj` names it. But a Tiled `"image"` path is **authoring metadata, not a load**. `themeRegistry.ts` imports its three tilesets explicitly and resolves tiles through its own firstgid table — and it never imports `lobby.tmj` at all. The map that referenced the file is itself unreachable, so both are gone.

Before deleting I checked that "unreferenced" really means "unbundled":

- no `import.meta.glob` anywhere in the tree
- no `publicDir`, and no `files`/`extraResources` in the electron-builder config
- `copy-main-assets.cjs` moves two `.cjs` sidecars and nothing else

An asset under `src/` reaches the bundle only by being imported — so `room-builder.png` was shipping **in the git repo and nowhere else**, which is exactly the surface that matters. Both files date from the inception commit and have never been touched since: vendored dead weight, not a parked feature.

**10 LimeZu files → 3**, and all three draw the office floor.

Checks unchanged: `typecheck:web` clean apart from the pre-existing `vite/client` error; `test:focused` byte-identical before and after the deletion (211 tests, same 4 environment failures from absent `node_modules`) — verified by running the suite against both trees and diffing per-file results, not by comparing totals.

PR #213, opened 2026-08-20
Arabic/RTL text support: terminal rendering recipe + markdown/composer bidi

abo123v-glitch:
## Problem

xterm.js implements no part of the Unicode bidi algorithm and no Arabic shaping (xtermjs/xterm.js#701, open since 2017 — VS Code's integrated terminal has the same complaint). For anyone running agents in Arabic, every reply renders as **disconnected letters in left-to-right order** — unreadable. The markdown preview and composers also laid Arabic out LTR.

macOS Terminal.app renders the same output perfectly because its text engine does shaping + bidi natively. This PR gets the app's terminals to the same place by letting the **browser's** text engine do that work.

## The terminal recipe (4 parts, opt-in)

1. **DOM renderer** — when the mode is on, the WebGL lease is skipped so rows are real DOM text nodes the browser can shape (`leaseWebglRenderer` returns early; everything else about the pool is untouched).
2. **Character joiner** (`terminal/arabicJoiner.ts`) — `registerCharacterJoiner` merges each Arabic phrase into one render range → one span, so the font applies real contextual joining (including lam-alef, marks on their letters).
3. **Bidi CSS** (`design/global.css`) — rows get `unicode-bidi: plaintext` (per-row base direction, like `dir=auto`: Arabic rows lay out right-to-left and right-aligned, Latin rows are untouched) and spans drop to `display: inline` — inline-block boxes are atomic to the bidi algorithm and can never be reordered. `!important` is load-bearing: the DOM renderer injects a runtime stylesheet that otherwise wins the cascade.
4. **Letter-spacing fix** (`terminal/arabicSpacingFix.ts`) — a MutationObserver strips xterm's cell-fitting letter-spacing from Arabic spans only. Any letter-spacing makes the browser disable cursive joining *and* stretches phrases across their full cell span. Latin spans keep theirs, so TUI box-drawing stays cell-aligned.

Plus one timing fix: terminals open at boot, often before webfonts finish loading, so xterm caches glyph widths measured against fallback metrics — a *different* broken layout every launch. On `document.fonts.ready` the pool pokes `options.fontFamily` (self-assign) to force a re-measure and repaint.

**Defaults:** on when the system locale is an RTL script (ar/fa/he/ur), off otherwise — so non-RTL users keep the GPU renderer exactly as before. A toggle in Settings → General overrides either way.

## HTML surfaces (always on, inert for LTR content)

- `markdown/rehypeAutoDir.ts` — a rehype plugin stamping `dir="auto"` per block in `MarkdownPreview` (headings, paragraphs, lists, cells; `pre`/`code` deliberately stay LTR, and inline `code` is pinned `direction: ltr` in CSS so a shell command never flips inside an Arabic paragraph). No `rehype-raw`; attributes only.
- Markdown preview CSS moves to logical properties (`padding-inline-start`, `border-inline-start`, `text-align: start`) so quotes/lists/tables mirror correctly.
- `dir="auto"` on the user-prose textareas (queue composer, ask-me answers, notes, schedules) and on agent-output display blocks (memory, kanban cards, Q&A, trigger history).
- **IBM Plex Sans Arabic** added to the font stacks as a per-glyph fallback — placed *before* `system-ui`/`-apple-system`, which resolve to SF Pro (it carries Arabic and would otherwise win the fallback, making the choice a no-op on macOS).

## Verification

- Captured real `claude` pty output bytes (an Arabic reply, colors and cursor addressing included) and replayed them through the exact terminal config in a browser — joined, correctly ordered, right-aligned; Latin rows byte-identical behavior.
- Reproduced the late-font failure mode (stylesheet injected 1.5 s after `open()`) and confirmed the `fonts.ready` re-measure repairs it.
- `npm run typecheck`, `npm run build`, and the focused test suite (239 tests) all pass.

Happy to split this into smaller PRs (terminal vs. markdown/composer) if you prefer.

### DIFF
--- a/src/renderer/index.html
+++ b/src/renderer/index.html
@@ -10,7 +10,7 @@
          (small caps labels only); body/UI moves to Inter and code/terminal to
          JetBrains Mono — the pixel body fonts (Pixelify/VT323) read as toy-like
          and strained at data-dense sizes. -->
-    <link href="https://fonts.googleapis.com/css2?family=Press+Start+2P&family=Inter:wght@400;500;600;700&family=JetBrains+Mono:wght@400;500;700&display=swap" rel="stylesheet" />
+    <link href="https://fonts.googleapis.com/css2?family=Press+Start+2P&family=Inter:wght@400;500;600;700&family=JetBrains+Mono:wght@400;500;700&family=IBM+Plex+Sans+Arabic:wght@400;500;600;700&display=swap" rel="stylesheet" />
     <style>
       #cth-splash {
         position: fixed; inset: 0; display: flex; flex-direction: column;
--- a/src/renderer/src/components/AddAgentModal.tsx
+++ b/src/renderer/src/components/AddAgentModal.tsx
@@ -941,6 +941,7 @@ export function AddAgentModal({ onClose, config, onConfigChange }: AddAgentModal
 
                     <Row label="Goal (optional)">
                       <textarea
+                        dir="auto"
                         value={goal}
                         onChange={(e) => setGoal(e.target.value)}
                         placeholder="long-running directive injected on every prompt"
--- a/src/renderer/src/components/AgentStrip.tsx
+++ b/src/renderer/src/components/AgentStrip.tsx
@@ -199,6 +199,7 @@ export function AgentStrip({ config }: AgentStripProps) {
                       line per bullet) and the fullscreen roster renders every
                       line — an <input> would silently eat the newlines. */}
                   <textarea
+                    dir="auto"
                     autoFocus
                     rows={3}
                     value={a.note ?? ''}
--- a/src/renderer/src/components/AskMeTab.tsx
+++ b/src/renderer/src/components/AskMeTab.tsx
@@ -195,12 +195,13 @@ export function AskMeTab() {
 
             <div style={{ padding: 9, display: 'flex', flexDirection: 'column', gap: 8 }}>
               {/* the question */}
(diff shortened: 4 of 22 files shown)


chaitanyagiri:
Your Arabic work is in v0.4.6, and we did not handle this PR the way we should have.

Three of your files shipped byte for byte identical to what you wrote here:

- `src/renderer/src/terminal/arabicJoiner.ts`
- `src/renderer/src/terminal/arabicSpacingFix.ts`
- `src/renderer/src/markdown/rehypeAutoDir.ts`

They went in through `59d721ed` (the terminal half) and `6f7ffb89` (the Arabic UI half, gated on the chosen language). Both commits are authored by the maintainer rather than by you, and both name this PR in their message. That was the wrong call. The right one was to merge your branch and keep your authorship on the commits.

The release announces that the interface speaks Arabic. The joiner and the spacing fix are the reason that claim is true, and the release notes credit you by name and say the work was re-implemented rather than merged.

The rest of the UI wiring here, 19 files, was rebuilt differently around a language gate, so this branch now conflicts and cannot be merged as it stands. That is why this closes rather than shows as merged. It is not a rejection of the work, it is us closing the record on a branch that the shipping code has already moved past.

If you want the credit recorded differently anywhere, including co-author trailers on those two commits, say so and we will do it. Thank you, and sorry for how this was handled.

abo123v-glitch:
Thank you for the transparency and the honest apology. I really appreciate how you handled this once it was caught.

Yes, please add the co-author trailers to both commits (59d721ed and 6f7ffb89). That would mean a lot.

I'm glad the Arabic support landed and that Arabic speakers can now use the app in their language. Happy to contribute again in the future.

PR #217, opened 2026-08-20
feat(memory): bundle mempalace so a fresh install works out of the box

Joyboy-World:
## What & why

Munder Difflin resolves the `mempalace` CLI from the user's PATH and a few common
uv/pip locations. On a machine that has never installed it, resolution returns
`null` and the memory layer is silently unavailable — the failure a new user hits
first and understands least.

This adds a **third and final** resolver step: the copy bundled with the packaged
app, via electron-builder `extraResources`, following the existing `kg.cjs`
idiom already in the repo.

It is deliberately **last**. A user's own install (steps 1–2) always wins, so
this changes nothing on a machine that already has mempalace — it only makes a
bare machine work out of the box.

**The catch that makes this more than a file copy:** `mempalace.exe` from pip is a
`simple_launcher` shim with this machine's absolute Python path embedded in it.
The naive "copy the binary in" bundle passes every test on the developer's
machine and fails on every user's. Per-platform payload slots plus their LICENSE
files are laid out under `resources/mempalace/{win,mac,linux}/` so the real
per-platform artifact drops in without reshaping the build.

Rebased onto `main` at v0.4.5. One commit.

### Before
image

The battery against unpatched `main` — **3 pass, 1 fail**, and the failing row is
exactly the one this PR fixes:

`✖ a bare machine resolves the BUNDLED copy — the whole point`
`AssertionError: no PATH hit, no pip spots — the extraResources copy must be found`
`actual: null`

### After
image

The same battery on this branch — **4 pass, 0 fail.**

The other three rows pass on *both* sides. That is the point of the control: the
suite fails only the row the fix addresses, so a green run here means the
fallback works, not that the test is lenient.

## Type of change

- [x] Bug fix / new behaviour behind existing idiom
- [ ] Breaking change

## Checklist

- [x] `npm run typecheck` passes
- [x] `npm run build` passes
- [x] New test added (`test/memory-bundled-bin.test.cjs`) — picked up automatically
      by the `test:focused` glob added in v0.4.5
- [x] Rebased on latest `main`, no conflicts
- [x] Behavioural control run against unpatched `main` (above)

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -55,6 +55,15 @@ extraResources:
   # repo resources/skills). Directory copy preserves the <id>/SKILL.md tree.
   - from: resources/skills
     to: skills
+  # MemPalace CLI, bundled per platform so a fresh install works on a machine
+  # that has never seen Python or pip (today the app expects a user-installed
+  # `mempalace` and silently no-ops without one — the memory layer's whole
+  # prerequisite). memory.ts bin() resolves the user's own install FIRST (PATH,
+  # then the common uv/pip spots); <resourcesPath>/mempalace is strictly the
+  # fallback. Binaries land in resources/mempalace/<os>/ (see its README for
+  # the per-platform build recipe); LICENSE (MIT) ships beside the binary.
+  - from: resources/mempalace/${os}
+    to: mempalace
 
 asarUnpack:
   - "**/node_modules/node-pty/**"
--- a/resources/mempalace/README.md
+++ b/resources/mempalace/README.md
@@ -1,0 +1,53 @@
+# Bundled MemPalace
+
+This directory feeds the `extraResources` rule in `electron-builder.yml`: at
+package time, the contents of `resources/mempalace/<os>/` (win | mac | linux)
+are copied to `<resourcesPath>/mempalace/` inside the installed app.
+
+## Resolution order (see `src/main/memory.ts` `bin()`)
+
+1. The user's `PATH` (`where` on Windows, login-shell `which` elsewhere).
+2. Common install spots (`~/.local/bin`, Homebrew, pip's Scripts dir).
+3. **The copy bundled here — strictly last.** A user's own install always
+   wins; the bundled binary exists so a machine that has never seen Python
+   or pip still gets semantic memory out of the box instead of the silent
+   no-op degrade.
+
+## What goes in `<os>/`
+
+A **standalone** `mempalace` executable (`mempalace.exe` on Windows) that runs
+with no Python installation present, plus the `LICENSE` file already placed in
+each directory (MemPalace is MIT — the notice ships beside the binary).
+
+> ⚠ **Do NOT drop in the executable from `~/.local/bin`.** On Windows that
+> file is pip's `simple_launcher` shim: ~108 KB, embeds an absolute path to
+> the machine's `python.exe`, and only appears to work because the backing
+> environment exists locally. On a user's machine it dies immediately. The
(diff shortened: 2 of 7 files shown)


chaitanyagiri:
## TL;DR

Thanks @Joyboy-World. The problem you are solving is real and it is one of the sharper edges a new user hits: the memory layer is described as core to this app, and today it silently no ops on any machine without a user installed `mempalace`. Making a fresh install work out of the box is worth doing.

**What it does:** adds an `extraResources` rule to `electron-builder.yml` so a per platform `mempalace` binary is copied to `<resourcesPath>/mempalace/` at package time, and teaches `src/main/memory.ts` to fall back to that copy when the user has no install of their own.

**Verdict: needs work.** The code half is correct and I have no complaints about it. **But the PR ships no binaries**, so the feature it promises does not work when merged.

---

## What I did

Static review of the diff against `origin/main` @ `eb2818a8`. I checked the diff for committed binary content specifically, since that is the crux. **I did not check out your branch or run a package build.**

## Finding 1: the resolution order is right, and I want to say so first

```
1. the user's PATH
2. common uv/pip spots (~/.local/bin, Homebrew, pip's Scripts dir)
3. the bundled copy — strictly last
```

Putting the bundle last is the correct call and the comment explains exactly why: a user who already has `mempalace` keeps whatever version they chose, and bundling changes nothing for them. Guarding on `process.resourcesPath` so it is skipped under plain node and in dev, where it would point at Electron's own dist resources, is the right defensive detail. This part is ready.

## Finding 2: no binaries are in this PR, so the feature does not work

The changed files under `resources/mempalace/` are:

```
resources/mempalace/README.md
resources/mempalace/linux/LICENSE
resources/mempalace/mac/LICENSE
resources/mempalace/win/LICENSE
```

There is no executable in the diff, and no binary content at all. So `from: resources/mempalace/${os}` copies a `LICENSE` file and nothing else, and `existsSync(bundled)` at the new fallback is false on every platform. Merged as is, the behaviour is identical to today and the README describes a capability the app does not have.

Your README is honest about this. It tells a maintainer what to drop in and warns them off copying the executable from `~/.local/bin` on Windows. So I do not think this is an oversight so much as a PR that is waiting on a step only a maintainer can take.

## Finding 3: the decision this PR is really asking for, which is not a code decision

The next step after this merges is that somebody commits three PyInstaller built standalone executables into the repository. That is a supply chain and repository decision and it belongs to the maintainer, not to you:

- A committed binary is **opaque to review**. No reviewer on any future PR can tell whether `mempalace.exe` changed for a good reason, which is the exact property that makes bundled binaries a common attack path.
- They are **large and permanent**. A PyInstaller bundle of a Python CLI is tens of megabytes per platform, and git keeps every version forever. This repo has already had a `.git` size problem raised.
- They need an **update story**. Nothing in this PR says who rebuilds them when MemPalace releases, or how a user learns their bundled copy is stale.

None of that means do not do it. Bundling is a normal thing for a desktop app to do and the out of box experience argument is strong. It just needs the maintainer to say yes to the binaries specifically.

Two routes, and I would let the maintainer pick.

**Route A, keep this PR as the plumbing.** Say clearly in the description that this lands the resolution logic and the packaging rule only, and that the binaries follow separately. Then it is safe to merge today because it is a strict no op until the files appear. This is my preference.

**Route B, make it complete.** Add the three binaries, and with them: a documented build recipe, recorded checksums, and a note on who owns rebuilding them.

Either way, one small addition would help: when the fallback is used, log which `mempalace` was resolved and from where. Right now a user with a broken memory layer has no way to tell which of the three paths won.

PR #218, opened 2026-08-20
Fix race-safe realtime task writes

AmIrRX0:
## Summary

- Routes realtime create, assign, update, and delete task actions through HiveManager's atomic addTask, patchTask, and deleteTask methods.
- Returns a spoken failure when an atomic mutation does not succeed instead of reporting success.
- Adds regression coverage that prevents realtime voice actions from restoring the unsafe whole-ledger write path.

## Scope

This covers every task-ledger writer in src/main/realtimeActions.ts. The other write-capable actions in that file update separate stores such as messages, schedules, control state, and config; they do not write tasks.json and are outside this task-race fix.

The previously bundled Antigravity readiness and Crush config-path changes have been removed from this PR.

### Before
Before: four unsafe whole-ledger writes

The regression assertion finds four whole-ledger writes on the pre-fix base and exits with status 1.

### After
After: atomic task mutations pass

The same assertion passes on this branch; all 549 focused tests and the full typecheck pass too.

## Testing

- npm run typecheck
- node --test --test-reporter=dot test/*.test.cjs (549 passing)


chaitanyagiri:
## TL;DR

Thanks @AmIrRX0. Moving task writes off direct file access and onto main process IPC is the right architectural call, and the class of bug it closes is one this project has been bitten by before: a card that silently vanishes because two writers raced.

**What it does:** routes `addTask`, `patchTask` and `deleteTask` through main via `hiveAddTask` / `hivePatchTask` / `hiveDeleteTask`, returns a spoken failure when a write does not succeed instead of assuming it did, adds `CRUSH_GLOBAL_CONFIG` for the Crush config path, and a small antigravity readiness change.

**Verdict: needs work.** No defect found in the approach. It is CONFLICTING and it bundles three unrelated things, which is what needs sorting.

---

## What I did

Static review of the diff against `origin/main` @ `eb2818a8`. **I did not check out the branch or run the tests.**

## What I like

**Returning a failure instead of assuming success is the actual fix**, more than the IPC move is. These read well:

```
if (!deps.hiveAddTask(card)) return { ok: false, spoken: 'I can’t create that task because it already exists.' };
```

Silently succeeding on a failed write is how cards go missing, and a spoken message means the user finds out at the moment it happens rather than noticing an empty board later.

**Putting the write behind a `deps` object** keeps `realtimeActions.ts` testable without Electron. Good shape.

## Finding 1: this is three PRs in a coat

The title covers race safe task writes, antigravity readiness, and the Crush config path. Those are unrelated: one is a correctness fix on a known bug class, one is a provider readiness tweak, one is a single env var.

That matters because the task write fix is the one worth expediting, and right now it cannot be merged without also merging the other two. If you split it, the race fix could land quickly on its own merit.

## Finding 2: it is conflicting, and the file it conflicts on is busy

`src/main/index.ts` and `src/main/hive.ts` are among the most contended files in the open queue. A rebase now is worth more than a rebase later.

## Finding 3: a question rather than a finding

There has been previous work in this area noting that `realtimeActions.ts` had several unsafe writers, not only the ones a single change reached. Your diff converts the task write paths. Worth stating in the description whether you believe the remaining writers in that file are safe, or whether they are simply out of scope here. Either answer is fine; leaving it unsaid means the next person has to redo the analysis.

1. Rebase.
2. Split the Crush config path and the antigravity readiness change out, or at minimum call them out separately in the description so a reviewer knows they are along for the ride.
3. One line on whether the other writers in `realtimeActions.ts` are in scope.

The core change is one I would want. It is the packaging that needs work.

AmIrRX0:
Thanks for the review — addressed all three points:
- Rebased onto the latest main; the PR is now conflict-free.
- Removed the unrelated Antigravity readiness and Crush config-path changes.
- Clarified the scope in the description: all task-ledger writers in realtimeActions.ts are covered; writers targeting other stores are out of scope.
- Added regression coverage and before/after evidence.
All 549 focused tests and the full typecheck pass. Ready for another look when you have time.

AmIrRX0:
Hi @chaitanyagiri,

Thank you for merging #218. I noticed that my GitHub account, @AmIrRX0, has not appeared in the repository’s Contributors list.

It looks like the commit was authored as `RARz <farhang@RARz.local>`, which is not linked to my GitHub account, even though the pull request itself was submitted by @AmIrRX0.

Could you please confirm whether this is only a delay in GitHub’s contributor statistics or an attribution issue? If possible, I would appreciate your help in ensuring that my contribution is credited to @AmIrRX0.

I have corrected my Git author name and email for future contributions.

Thank you!

chaitanyagiri:
Hi, you're right and it's not a delay.

Your commit was authored as `farhang@RARz.local`, and GitHub only counts a commit toward the contributors list when that email is on a GitHub account. A `.local` address can't be added to one because it isn't a real domain, so that commit won't link back to you retroactively.

I'm not going to rewrite main's history over it. That breaks every clone out there for one line of metadata.

What I'm doing instead: you're going into CONTRIBUTORS.md by handle, and you're credited in the 0.5.0 release notes. PR #218 permanently shows you as the author and that's the link people actually follow.

Good call fixing your git config. The next one will count properly.

Thanks for the fix, by the way. It closed a real race where a task card could get erased by a simultaneous write.

PR #205, opened 2026-08-20
feat(i18n): add react-i18next multilingual UI (en + zh-CN)

lukatao:
## What & why

This PR is an interest-driven implementation of i18n for Munder Difflin, done out of genuine appreciation for the project — we wanted Chinese-speaking users to be able to use it in their own language.

It adds a full `react-i18next` setup with **English (default)** and **简体中文 (zh-CN)** built in, and an architecture that extends to any language.

## Technical summary

- **Infrastructure**: `src/renderer/src/i18n/` — i18next init with en + zh-CN resources, `fallbackLng: 'en'`, initial language detected from `navigator.language`, choice persisted to localStorage; extensible to any locale by adding a `locales/<code>.json` and one `LANGUAGES` entry.
- **Switcher**: Settings → General now has a **Language** dropdown — switching applies instantly and persists.
- **Coverage**: 45 components across the core UI and the long tail — office floor (ambient flavor lines included), Command Center, Settings, Add Agent, Onboarding, agent cards/strips, kanban, queue composer, IDE, file tree, git panes, memory graph/panel, triggers, skills, workers, integrations, updates — **1,122 translation keys** per locale, keyset fully aligned between en and zh-CN.
- **Quality**: `npm run typecheck` (node + web) passes with zero errors. Translations are natural, not machine-translated.
- Deliberately kept English: Slack/webhook API documentation blocks, agent-prompt templates (HIRE_PROMPT), CLI endpoint hints, model/engine proper nouns.

## Note to maintainers

If there is already a better / more complete i18n effort upstream or from the community, we are happy to close this PR — we just wanted Chinese users to be able to use Munder Difflin in their own language. Happy to adapt, split, or rebase as you prefer.
## Review guide (why this PR is large, and how to review it efficiently)

This PR touches 53 files (+4.4k / −1.3k). That size is inherent to internationalizing a whole UI: every visible string in 45 components moved from a hardcoded literal to a `t()` call, plus the two locale files. We kept the change mechanically uniform so it reviews fast:

- **One pattern everywhere**: each component adds `const { t } = useTranslation()` and replaces literals with `t('layer.key')` — no logic, state, or styling changes.
- **Key naming**: `layer.component.field` (e.g. `settings.title`, `officeTheme.notBuiltYet`, `commandCenter.tabs.floor`) — grouped by surface so a reviewer can spot-check one layer and trust the rest.
[...]

PR #220, opened 2026-08-21
fix(wake): wake headless workers and give them Auto Mode permissions

lukatao:
## What & why

This PR fixes two linked bugs that silently broke **headless workers spawned via the spawn-requests path** (the `hive/spawn-requests/*.json` mechanism). Workers were never woken to read their dispatched objective, and when they did run they'd park forever on a permission prompt nobody was watching. Both are done out of genuine appreciation for the project — we run Munder Difflin as our own multi-agent hive and hit these in the wild.

Branch: `fix/wake-bug` — 3 commits, main-only (no i18n work mixed in).

---

### Symptom

A worker spawned via `spawn-requests` sat at an idle prompt forever. Its objective (dispatched into its inbox by main) was never read, and the idle-reaper eventually killed it. In the wild: `worker-i18n-01/02` showed `inboxBacklog: 1`, `lastActive` == spawn time, then got reaped 20–30 min later — never woken even once.

### Root cause

Delivery to a worker's TUI depends on the renderer's inbox-wake loop (useHive effect #3), which **iterates store agents**. But `processSpawnRequest` (src/main/index.ts) never told the renderer the worker's PTY existed: unlike the voice-spawn path (`spawnAgent`), it never emitted `hive:agentSpawned`. So no floor card was created, the wake scan skipped the pty entirely, and the dispatched objective sat in the inbox forever. (Not a terminal-pool issue: the renderer pre-warms a pooled terminal per agent at App mount, and `writePty` is plain IPC to the main session map — both were fine.)

### Fix

- **main** (`src/main/index.ts`, `processSpawnRequest`): broadcast `hive:agentSpawned` on a successful spawn — the same descriptor `spawnAgent` emits (id, name, provider, cwd, command, role, worktreePath). The renderer builds a floor card from it and the nudge+drain loops deliver the objective like any other agent.
- **renderer** (`src/renderer/src/hooks/useHive.ts`): the inbox-wake scan now also probes live ptys with no store entry; a pty that answers with inbox mail is carded on the spot (idempotent), so a worker whose broadcast was lost to a renderer reload still gets woken. Card construction is shared (`cardForSpawnedAgent`).

### Verification

`npm run typecheck` (node + web) exit 0; `npm run test:focused` 239/239 pass. Post-restart regression test: a background agent was auto-woken with no human input (~22s latency) and processed its inbox message on its own.

---

### Symptom

With the app in Auto (skip-permissions) mode, a spawn-request worker's first tool call hung on a `(y/n)` permission prompt that nobody was watching. Renderer-spawned agents (via Add Agent) automatically got `--permission-mode bypassPermissions`; spawn-request workers didn't.

### Root cause

`processSpawnRequest` took the bare `defaultCommand` (`'claude'`) and never appended the provider's auto-mode flag — renderer's `buildSpawnCommand` does (`config.autoMode && preset.autoFlag`), the main-side spawn path didn't.

### Fix

- `processSpawnRequest` now appends the provider's own auto flag when `config.autoMode` is set, using `autoModeFlagForProvider` from shared/agentProvider.ts (claude → `--permission-mode bypassPermissions`, codex → `--dangerously-bypass-approvals-and-sandbox`, kimi → `--auto`, agy → `--dangerously-skip-permissions`, grok → `--yolo`; no-flag providers untouched). Idempotent — an explicit `raw.command`/`raw.model` that already carries the flag is never doubled.
- **Important detail (the T4b lesson):** the flag goes into `spawnOpts.args`, **never** the command string. `resolveCommand` treats the whole command as the binary to `which`/`existsSync`, so `'claude --permission-mode bypassPermissions'` as a command fails to resolve and the worker dies instantly with `process exited (code 1)`. The renderer path is clean because it builds the command and then `tokenizeCommand`s it into `[exe, ...args]`; the main-side path now mirrors that.

### Verification

`npm run typecheck` (node + web) exit 0; `npm run test:focused` 239/239 pass.

---

## Note to maintainers

This project is one we use and enjoy, and this PR is interest-driven. If there is already a better / more complete implementation of headless-worker wake or auto-permissions upstream, we're happy to close this PR — just let us know. Happy to adapt, split, or rebase as you prefer.

_Companion i18n PR (same fork): #205._

---

## Visual evidence

The floor before the fix: a headless worker whose dispatched objective never reaches its terminal (inbox stays undrained).

### Before
before
### After
after

> **Rebase note:** this branch was rebased onto current upstream `main`. The original three-commit fix was largely absorbed upstream (`processSpawnRequest` now broadcasts `hive:agentSpawned`, and `buildWorkerLaunch` handles auto-mode flags in `spawn args`), so the branch now carries the remaining renderer increment as a single commit: the inbox-wake scan cards and wakes main-spawned workers whose broadcast was lost (e.g. a renderer reload), so their inbox mail still gets delivered.


lukatao:
Closing this because the core headless-worker wake and Auto Mode fixes are now upstream; the remaining renderer fallback is stale and conflicts with current main.

PR #222, opened 2026-08-21
feat: surface God task conclusions in the UI

jhinzzz:
## Summary

- Require God to hand off local task conclusions through the top-level task card and final result block.
- Show God-owned local results in the persistent completion toast and expose the same result on task details.
- Keep worker, Slack, webhook, and scheduled task delivery paths unchanged.

## Validation

- `npm run test:focused` — 242 passed
- `npm run typecheck`
- `npm run build` (existing Vite chunk warning only)
- Manual Electron smoke check: Tasks view and DONE task `RESULT` detail are visible.

## Scope

- No new IPC, dependency, database, or task storage protocol.
- Existing unrelated `package-lock.json` worktree change is intentionally not included.

### Before

Before: completed task without a visible result

### After

After: completed task with a visible result


jhinzzz:
@chaitanyagiri The requested evidence is now attached under `### Before` and `### After`, but the required check still reports both sections as missing.

I reproduced the workflow parser against the current PR body. In `section()`, the regex combines the `m` flag with the lazy capture terminator `|$`. Under multiline mode, `$` matches the blank line immediately after each heading, so both captured sections are empty.

A minimal fix is to replace that alternative with a true end-of-string check, for example `(?![\\s\\S])`, and rerun the workflow.

Failed run: https://github.com/chaitanyagiri/munder-difflin/actions/runs/32628109985

PR #226, opened 2026-08-21
fix(cost): report lifetime spend, not spend since the last app restart

chaitanyagiri:
## What is wrong

`fleet.json` reports 59% less spend than the floor has actually cost.

Measured on the live ledger on 2026-08-21: **59% of true spend is invisible**, across dozens of reset events. Angela found this and I reproduced her figures independently before changing anything. The hidden amount matches her number to the cent.

## Why

`AgentUsageSample.usd` counts from process start, not from the agent's beginning.

`TelemetryCollector` accumulates into in memory maps (`sessions` and `agentSessions`, `src/main/telemetry.ts:130-132`). An app restart rebuilds them empty, so the counter starts again at zero under the **same `session_id`**, because the agent resumes and only our accumulator forgot.

The sharp edge is at `src/main/telemetry.ts:180-182`:

```ts
const live = this.aggregateLive(agentId);
if (live) return live;
return this.transcriptFallback(agentId);
```

`aggregateLive` returns a truthy object as soon as an agent has one session registered, even at zero. So the moment a single span lands after a restart, the transcript fallback that holds the true total is **never consulted again**. `appendCostLedger` then writes that reset value straight to disk, which is why the ledger itself shows the drops.

The resets are unmistakably restarts. They land floor wide in the same minute: four agents at 2026-08-10T12:44, three at 2026-08-21T14:35, and most fall from a large peak to exactly zero.

## Part 1: the breaker. CONFIRMED, and it is the part that matters

The floor wide cost cap at `src/main/breaker.ts:227-237` reads the same resetting counter:

```ts
const usd = i.sample?.usd ?? 0;
total += usd;
...
if (total <= cfg.costCapUsd) topSpender = null;
```

After a restart every agent's `sample.usd` is near zero, so `total` is near zero, so nobody is blamed. **The cost cap measures spend since the last app restart, not lifetime.** It is not that it can never fire, it is that it re arms to zero on every restart, and with dozens of resets in this ledger the lifetime cap has never been approached.

The same hole applies to `costCapTokens` at `:245` and the per agent token cap at `:304`, both of which read the same sample.

**I have not touched breaker.ts.** A guardrail that starts tripping correctly could stand agents down mid task, so per the brief this is written up and left to the founder. The velocity trip at `:320-322` is unaffected in practice, because the breaker's own `lastSample` state resets in the same process restart, so the two reset together and no negative delta is produced.

## Part 2: the reporting fix, which is what this PR contains

The ledger keeps every line, so the pre reset peaks are still on disk. Within one agent and session the counter only climbs, so a decrease is the reset signature and nothing else. Lifetime is the sum of each closed segment's peak plus the open segment.

**Any decrease counts.** A dollar threshold was considered and rejected: two genuine restarts in the live ledger drop from sub dollar peaks straight to zero, and Angela's $1 threshold silently misses both. A cumulative counter has no legitimate reason to fall, so the magic number bought nothing and cost coverage. The full spread across thresholds is about one percent of the total, so this is a correctness choice rather than a material one.

New `src/main/costLifetime.ts`. Folding is **async and incremental**, because the ledger is already large enough that a full pass takes about 800ms, which must not land on the Electron main thread. Each pass reads only the bytes appended since the last one, so steady state is a few hundred bytes.

`fleet.json` now carries lifetime in `usd` and keeps the old figure as `sessionUsd`. That means the LIVE ROSTER line every agent sees (`src/main/hive.ts:1861`) is correct with no second change. Until the first fold completes, `usd` falls back to the session figure rather than publishing a confident cold zero.

## Verification

- 15 new tests, covering resets, sub dollar resets, incremental folding matching one shot folding, torn trailing lines, truncation, malformed rows and a non numeric `usd`.
- Cross checked against the real ledger by a second implementation written separately. Both agree to the cent, with no per agent disagreement.
- `npm run typecheck` clean on both projects. Focused suite 254 tests.

## Honest notes

- **One flaky test, not mine.** `test/hive-hook-node.test.cjs` fails roughly half the time under full suite load with EPIPE, and passes 4/4 in isolation every time. It is a socket race that pre dates this change, though adding a 16th test file to a concurrent run does make it surface more often. I did not fix it because it is outside this card.
- **The ledger grows without bound**, and it has grown steadily over the weeks it has been running. The incremental fold means only the first pass after launch pays for size, but that first pass grows linearly and will eventually be slow. Rotation or a checkpoint file is worth a card. Not done here.
- **I did not change what `usd` means for the renderer.** `ToolWaterfall` and `costStore` read telemetry samples directly and still show session figures, which is correct for a live view.
- Numbers are dated 2026-08-21 and move, because the floor is still spending. The hidden fraction has been stable at about 59%.

https://claude.ai/code/session_013Rue1dRSfggYgRh5aHzqZn


chaitanyagiri:
Shipped in **v0.4.5** — closing as delivered, not abandoned.

This landed via 6c19ede, which is the commit this branch carried when it was integrated. Your branch was updated afterwards to 88c8c3c, so GitHub can no longer see its head on `main` and wouldn't auto-close it. The content is identical — I checked with `git patch-id --stable` and both sides give `7bfa7b6d8614`.

The fix is live: spend is folded from the ledger instead of read from an accumulator that reset on every app restart, with a separate session figure kept alongside it. [v0.4.5 release](https://github.com/chaitanyagiri/munder-difflin/releases/tag/v0.4.5).

PR #231, opened 2026-08-22
SELF-TEST: does the evidence check actually block?

chaitanyagiri:
### Before
before

### After
after


chaitanyagiri:
Self-test complete and the check behaves. Blocked an empty description, commented with what was missing, then flipped to ✅ when before/after were added by editing the body. Closing; the branch was never meant to land.

PR #235, opened 2026-08-22
fix(wall-sync): the scope-filter comment keeps the reasoning, drops the count

chaitanyagiri:
`scripts/wall-sync.mjs:33` carried an absolute count of records in our Razorpay account, in a public repo. The count is not load-bearing — the comment only has to explain why the epoch filter exists, and it does that just as well without a magnitude attached.

Same shape as #233 on `costLifetime.ts`: keep the reasoning, drop the figure. Comment-only, no behaviour change.

### The rule this applies

Absolute internal figures never go public. A base-free ratio that characterises a bug or a behaviour is not an internal figure and may stay. The test is one question: **can a reader recover a business quantity from it?** If yes it goes, if no it stays.

`~9,800 legacy records` is an absolute count of our payment history, so it goes. The `59%` in `costLifetime.ts:14` has no base attached and recovers nothing, so it stays — which is the same line #233 was already drawing in practice when it stripped the dollar pair and left the ratio.

**What this does and does not do:** it reduces the surface going forward. It retracts nothing. The git history still carries the figure.

### Before

`scripts/wall-sync.mjs:32-33`

```js
// Same scope filter as the dashboard collector: the engraving-notes key is the
// discriminator for wall payments; the date guards against ~9,800 legacy records.
```

### After

`scripts/wall-sync.mjs:32-34`

```js
// Same scope filter as the dashboard collector: the engraving-notes key is the
// discriminator for wall payments; the date guards against the much larger body
// of legacy records that predates the wall.
```

The reason the guard exists survives the edit; only the magnitude is gone.

### Sweep

I ran the rule across the tree rather than the log, spelling out every form a figure can take (currency, thousands separators, `k`/`MB`/`GB`, counts of tokens/records/installs/users/payments, `~`-prefixed approximations) rather than searching for the shape of the last leak:

```
git grep -inE '\$[0-9][0-9,]*|[0-9]{1,3},[0-9]{3}|[0-9,]+ ?(tokens|records|installs|users|customers|payments|rows)|install base|~[0-9,]+'
```

Everything else it surfaced fails the recover-a-business-quantity test and stays: technical timings and byte sizes, base-free performance ratios (`~216× real-time`, Opus undercosted `~5×`), and Apple's public `$99/yr` Developer Program fee. The previously-scrubbed absolutes are confirmed gone from every published surface — `CHANGELOG.md`, `docs/`, `blog/`, `RELEASE.md` return no hits for the spend pair, the token total, or the install-base pair.

This branch is cut from `main` at 31ac125, so it includes #233.

https://claude.ai/code/session_0111uSG4Ge2U6nUxnea7ZfeY

### DIFF
--- a/scripts/wall-sync.mjs
+++ b/scripts/wall-sync.mjs
@@ -30,7 +30,8 @@ const DRY = process.env.DRY_RUN === "1";
 const PR_BRANCH = "wall/additions";
 
 // Same scope filter as the dashboard collector: the engraving-notes key is the
-// discriminator for wall payments; the date guards against ~9,800 legacy records.
+// discriminator for wall payments; the date guards against the much larger body
+// of legacy records that predates the wall.
 // Epoch is Aug 13 UTC because the first wall payment is 2026-08-13T19:xx UTC
 // (= Aug 14 IST, the wall's listed date).
 const MD_EPOCH = Math.floor(Date.parse("2026-08-13T00:00:00Z") / 1000);


chaitanyagiri:
Waived with `no-visual-change`.

The diff is one hunk in a comment. There is no behaviour to record before and no behaviour to record after, so there is no honest asset to attach. The evidence gate only matches uploaded images and video by construction, which means no wording of a text before/after can clear it. That is the case the waiver exists for.

Worth saying out loud: the right move here was asking for the label rather than screenshotting a terminal to turn the check green. A gate you can satisfy with a prop is not a gate. The before/after comment text in the body does the job a screenshot would have faked.

The rule the PR states is the one we are working to. Absolute internal figures never go public, a base-free ratio that characterises a bug does not count as one, and the test is whether a reader can recover a business quantity. One caveat to carry forward: a ratio is only unrecoverable while its base is unpublished, so check the base before calling a ratio safe.

PR #238, opened 2026-08-22
fix(privacy): stop sending IP and derived geolocation to PostHog

chaitanyagiri:
`TELEMETRY.md` and the public privacy policy at `munderdiffl.in/privacy.html` both promised PostHog does not retain the IP on the event. It does. Measured live on project 555703 just now:

| property | events carrying it | of total |
|---|---|---|
| `$ip` | 221,393 | **221,393 (100%)** |
| `$geoip_city_name` | 198,809 | 89.8% |
| `$geoip_postal_code` | 188,963 | 85.4% |
| `$geoip_latitude` / `longitude` | 221,343 | 100.0% |

Window 2026-08-13 → 2026-08-22. A published promise, in a public repo, that anyone can verify is false. Reality moves to match the promise; the promise does not move to match reality.

### The app was not a passive victim of this

The card recorded this as PostHog-side and not caused by our code. That was wrong, and it is the reason this looked unfixable in code. `src/main/analytics.ts` explicitly set `disableGeoip: false`, overriding posthog-node's default of OFF, and the comment beside it restated the very promise it was breaking:

> `posthog-node defaults GeoIP OFF (server lib); we re-enable it for country-level numbers only — PostHog discards the IP after the lookup`

Nothing in this codebase ever read a country. `git grep -niE 'geoip|country|\$ip'` across `src/`, `dashboard/` and `scripts/` returns only that flag and its own comment. So the flag bought us nothing and cost city, postal code and coordinates on every event.

### What changed

Three collection surfaces, all reporting into the same project:

| surface | SDK | change |
|---|---|---|
| app | posthog-node | `disableGeoip: true`, and `$ip: null` on every captured event |
| blog | posthog-js | `before_send` nulls `$ip`, sets `$geoip_disable` |
| marketing site | posthog-js | same `before_send` |

PostHog fills `$ip` from the connection **only when the event does not carry one**, so sending it explicitly is the one client-side lever that exists for a server-set property. The 272 generated blog pages are rebuilt from source, so the fix reaches the published site rather than only the template.

### Before

`src/main/analytics.ts` — geoip explicitly re-enabled, comment restating the false promise:

```ts
// posthog-node defaults GeoIP OFF (server lib); we re-enable it for
// country-level numbers only — PostHog discards the IP after the
// lookup and the event is anonymous either way (TELEMETRY.md).
disableGeoip: false
```

`TELEMETRY.md` and `docs/privacy.html`:

> IP-based geolocation is used only to derive a country for aggregate stats; PostHog does not retain the IP on the event.

### After

```ts
// GeoIP stays OFF (posthog-node default). It was explicitly re-enabled
// here for country-level numbers, but nothing in this codebase ever read
// a country, and enabling it made PostHog derive city, postal code and
// lat/long as well — far past the country TELEMETRY.md describes.
disableGeoip: true
```

```ts
$process_person_profile: false, // anonymous event — no person profile
$ip: null
```

Both published promises now describe the stronger posture: *no geolocation of any kind is derived*. That is a tightening, not a softening — the doc is not being rewritten to excuse what we do.

### Two things this PR deliberately does NOT do

**1. It does not flip the project-level "Discard client IP data" toggle.** `anonymize_ips` is confirmed `false` on the project. Per PostHog's own docs that toggle is the only thing that stops `$ip` for clients that are *already deployed* — this code fix only takes effect for the app once a user installs a build containing it, so every v0.4.5 install in the wild keeps sending IPs until it updates, and one that never updates never stops. The toggle is a live production change, not a branchable one, and a standing instruction says the PostHog setting ships as one deliberate release when the founder decides. Raised rather than taken.

**2. It does not touch retention.** Collection and disposal are different questions and are not folded into one commit. Written up separately.

### Evidence and the positive control

A check that comes back empty proves nothing until you have shown it *could* have come back non-empty. The verification query splits cleanly by SDK, so each surface is provable on its own:

```sql
SELECT properties.$lib AS lib, count() AS n,
       countIf(properties.$ip IS NOT NULL) AS ip,
       countIf(properties.$geoip_city_name IS NOT NULL) AS city
FROM events WHERE timestamp > now() - INTERVAL 24 HOUR GROUP BY lib
```

**Positive control, run against the pre-fix window — the query detects IPs when they are there:**

```
web            n=12919   ip=12919   city=11542
posthog-node   n=7995    ip=7995    city=7126
```

Post-deploy the identical query over a post-deploy window must return `ip=0` for `web`, and `ip=0` for `posthog-node` restricted to `app_version` ≥ the build carrying this commit. Comparing against the numbers above is what makes a zero meaningful rather than a query that silently matched nothing.

CI runs typecheck — this worktree has no `node_modules`, so it was not run locally.

https://claude.ai/code/session_0111uSG4Ge2U6nUxnea7ZfeY

### DIFF
--- a/TELEMETRY.md
+++ b/TELEMETRY.md
@@ -55,8 +55,9 @@ identifiers, or API keys. Nothing free-form — the property allowlist in
   starts, so a build other than this one starting afterwards means something
   else did the installing. Only that one-word result leaves the machine — no
   line, path or message from that log is ever sent.
-- IP-based geolocation is used only to derive a country for aggregate stats;
-  PostHog does not retain the IP on the event.
+- No geolocation of any kind is derived. The app sends `$ip: null` on every
+  event and disables the GeoIP lookup, so no IP address, country, city, postal
+  code or coordinate is derived from your connection or stored on the event.
 
 ## Opting out
 
--- a/blog/src/_includes/base.njk
+++ b/blog/src/_includes/base.njk
@@ -58,6 +58,16 @@
         api_host: 'https://us.i.posthog.com',
         person_profiles: 'never',
         autocapture: false,
+        before_send: function (event) {
+          // PostHog fills $ip from the connection unless the event carries one.
+          // Null it here and disable the GeoIP lookup so no IP, city, postal
+          // code or lat/long is ever derived or stored for this page view.
+          if (event && event.properties) {
+            event.properties.$ip = null;
+            event.properties.$geoip_disable = true;
+          }
+          return event;
+        },
         disable_session_recording: true,
         respect_dnt: true
       });
--- a/docs/blog/404.html
+++ b/docs/blog/404.html
@@ -49,6 +49,16 @@
         api_host: 'https://us.i.posthog.com',
         person_profiles: 'never',
         autocapture: false,
+        before_send: function (event) {
+          // PostHog fills $ip from the connection unless the event carries one.
+          // Null it here and disable the GeoIP lookup so no IP, city, postal
+          // code or lat/long is ever derived or stored for this page view.
+          if (event && event.properties) {
+            event.properties.$ip = null;
+            event.properties.$geoip_disable = true;
+          }
+          return event;
+        },
         disable_session_recording: true,
         respect_dnt: true
       });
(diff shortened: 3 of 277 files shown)


chaitanyagiri:
Shipped in v0.4.6 as `51979565`. IP and derived geolocation are no longer sent to PostHog. Closing, the work is in the release.

PR #239, opened 2026-08-22
hive: retry the proxy-bridge bind and surface a degraded spawn

chaitanyagiri:
## What

A transient proxy-bridge bind failure used to degrade a proxy-tier agent (Crush) for its entire session: one log line nobody reads, no retry, nothing the user sees. `hive.ts` gates `installCrushConfig` behind `if (port > 0)` on purpose and that gate is unchanged. Deliberate degradation stays; silent degradation goes.

- `startProxyBridgeWithRetry`: up to three attempts with a 250/750 ms backoff. Each attempt kills the previous sidecar first (`startProxyBridge` is idempotent), so a retry never leaks a listener.
- When all three fail, routing is left untouched and the CLI still runs, but now `log.jsonl` records a `proxy-degraded` event, the renderer receives `hive:degraded`, the spawn result carries a one-line `degraded` reason, and main shows it as a native toast on the same notifications gate the circuit breaker uses.

## Why both

Retry and surfacing answer different failures. A loopback bind on port 0 fails only transiently, so a short retry ladder fixes the common case for free. The rare permanent failure is kept as a degradation we intend, and the user is told.

## Test

`test/proxy-bridge-retry.test.cjs` stubs `startProxyBridge` on the instance, no real sidecar or sockets:
- fails once then binds: exactly two attempts, agent fully proxied, no degraded event
- never binds: three attempts, no half-applied Crush config, `degraded` on the result, `hive:degraded` emitted, `proxy-degraded` in `log.jsonl`

554/554 tests, typecheck clean.

https://claude.ai/code/session_01G5X7vQmQDwrTCn2qBzuBiJ

### DIFF
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -181,6 +181,10 @@ export interface SpawnInjection {
    *  bare TUI rejects a positional seed. The renderer types it through the same
    *  per-pty write-chain as the inbox-wake nudge. (ondev-b) */
   seedPrompt?: string;
+  /** Set when the agent spawned in a DEGRADED posture the user should know about
+   *  (today: the proxy-bridge sidecar never bound after retries, so a proxy-tier
+   *  agent such as Crush runs without hive events). Human-readable, one line. */
+  degraded?: string;
 }
 
 const HOP_CAP = 12;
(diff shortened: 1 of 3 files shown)


chaitanyagiri:
Shipped in v0.4.6 as `3a605a3f`. The proxy bridge bind retries and a degraded spawn is surfaced. Closing, the work is in the release.

PR #240, opened 2026-08-22
security: keep the OS sandbox on in auto mode

chaitanyagiri:
> **Do not merge until the founder signs off on the permission posture.** This changes what an agent in auto mode is allowed to write. See "Behaviour change" below.

## The mechanism, verified before anything was changed

Auto mode ran every agent with no OS sandbox at all. The reason was a path layout: a hive worker writes to its agent folder at `<harnessHome>/hive/agents/<id>/`, which sits outside the project cwd.

- **Codex**: the preset's auto flag was `--dangerously-bypass-approvals-and-sandbox`, and the code comment gave exactly that reason: `-s workspace-write` blocked the agent folder. Codex 0.137 has `--add-dir <DIR>` (extra writable roots alongside the workspace), so the folder never needed to move.
- **Claude**: `--permission-mode bypassPermissions` only silences prompts. Claude's sandbox is a separate, opt-in setting the app never enabled, so nothing was "dropped"; it was never on. The fix there is the per-session `settings.json` the hive already writes.
- gemini `--yolo`, grok, agy and the rest have no OS sandbox to restore; untouched.

## The fix, zero new dependencies, no layout change

- Codex auto flag is now `-a never -s workspace-write`; `hive.ts` appends `--add-dir` for the agent dir, the hive root (research deliverables; the board for god) and the MemPalace dir. An explicit posture already on the command line, including the old bypass flag, still wins (`autoStanceTokens` / `hasAutoModeStance`).
- Claude's per-session `settings.json` gets `sandbox.enabled: true` with `sandbox.filesystem.allowWrite` **and** `permissions.additionalDirectories` for the same dirs. Both layers are needed: one governs Bash children, the other the Edit/Write tools; with only one the agent deadlocks on its own inbox.
- `sandbox.failIfUnavailable` stays off, so a platform without a sandbox (Windows) spawns exactly as before.

## Proof the sandbox is back

Real run, claude 2.1.239, `--permission-mode bypassPermissions --settings <new block>`:

```
touch $HOME/md-sandbox-probe   → touch: Operation not permitted   (file does not exist afterwards)
touch <agent dir>/probe        → ok
touch ./probe (cwd)            → ok
```

`test/auto-mode-sandbox.test.cjs` pins the settings shape and the codex flag. 554/554 tests, typecheck clean.

## Behaviour change, read this

**Agents writing anywhere outside cwd + their agent dir + the hive root (+ palace) are now DENIED by the OS.** Concrete case from this floor this week: I cherry-picked a commit onto another agent's worktree (Kevin's) from my own session. Under this PR that write fails with "Operation not permitted". That is the single-writer rule enforced by the OS, which is the point, but it is a change in what auto mode can do and deserves a line in the release notes.

## What was NOT tested

The codex behavioural write test could not run: there are no OpenAI credits on this machine. The `-a never -s workspace-write --add-dir` semantics come from `codex --help` (0.137) and a prior banner check (`sandbox: workspace-write [workdir, /tmp, $TMPDIR, <agent dir>]`), not from a live codex run that attempted a write. The Claude side is the part proven live.

https://claude.ai/code/session_01G5X7vQmQDwrTCn2qBzuBiJ

### DIFF
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -613,6 +613,10 @@ export class HiveManager {
        *  copied into the agent's `.claude/skills/` per spawn; undefined or missing
        *  is a no-op (tolerated until Kevin populates the resource dir). */
       skillsDir?: string;
+      /** Extra directories the agent's sandbox may write (e.g. the shared
+       *  MemPalace dir, which `mempalace` mutates). Absolute paths; ignored
+       *  for providers without a sandbox. */
+      extraWritableDirs?: string[];
     } = {}
   ): Promise<SpawnInjection> {
     const root = this.root();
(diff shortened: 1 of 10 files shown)


chaitanyagiri:
Shipped in v0.4.6 as `e42b0e2f`. The OS sandbox stays on in auto mode. Closing, the work is in the release.

PR #241, opened 2026-08-22
fix(update): the update installs itself, and says something true while asking

chaitanyagiri:
> **DO NOT MERGE.** Opened for review only, at the founder's instruction. The
> evidence check below is expected to be red and is deliberately not waived —
> see the last section.

0.4.5 shipped a restart prompt whose "What's new" read `{ box-sizing: border-box; }`,
and a first-run release page that could not fire on the release that introduced it.
Both are fixed here, from both ends, and an update no longer waits on a button press
that most people never make.

## What was actually wrong

Three faults in one path.

**`autoInstallOnAppQuit` was false.** A downloaded update sat there until someone
pressed "restart to update", and was replaced by the next one if they never did.
Measured on the existing install base, only 41.7% took 0.4.4 within four days while
56.7% stayed on 0.4.3. Quitting the app is now the install trigger. The explicit
restart is still the fast path and still the only thing that interrupts; nothing
installs under a running app.

**The first-run check could never fire on the release that introduced it.** The
version stamp shipped *in* 0.4.5, so every user upgrading from 0.4.4 arrived with no
stamp, took the `previous &&` branch, and was read as a brand-new install.
`telemetry-install-id` has been minted on first run since 0.4.2, so finding one with
no stamp beside it proves an older build already ran in this profile. Installs with
telemetry off have no id and stay quiet, which is the honest answer: they cannot be
told apart from a genuinely new install.

**`update-downloaded` trusted `info.releaseNotes`.** electron-updater builds those
from `releases.atom` — GitHub's *rendered* markdown. HTML comments do not survive
rendering, so a `<!-- drop -->` never can, and a `<style>` block leaves its
declarations behind as prose. The notes are now read from the release itself,
unrendered. A failed read leaves no notes at all, deliberately: a missing digest
beats a digest made of somebody's stylesheet.

Also: only a packaged build writes the stamp. Dev shares its user-data directory with
the installed app, and a dev run stamping the new version hours before the real
upgrade made that upgrade read as no change at all.

## And the same fault from the other end

The digest now drops stylesheet text by structure, so CSS cannot win again whichever
path feeds it. Four shapes, all live in our own feed:

- a whole rule on one line — `* { box-sizing: border-box; }` is character for
  character a markdown bullet, `*` being both the universal selector and a list
  marker, and it was the **only** bullet-shaped line in a 766 line feed;
- a selector opening a block with declarations on later lines (`:root {`), tracked by
  brace depth so nested at-rules do not stop the skip one rule early;
- CSS comments, which span lines and read as sentences ("Two pages, no JavaScript…"),
  the one kind of CSS indistinguishable from release prose by content;
- at-rules, which GitHub renders as **user mentions** — `@media` arrives as an anchor
  to a real account, which is also why `@keyframes` came back as `@Keyframes`, that
  being GitHub's casing for the org of that name. So an at-rule never starts its own
  line there, and the structural test runs against the line's text, not its markup.

Entities are decoded too, and decoded first: `&` in the tagline reached the toast as
`&amp;` for two releases, and an escaped `&lt;style&gt;` has to become a tag again
before the tag strip can remove it.

The filter is deliberately narrow. A bullet that mentions `display: flex` survives,
and there is a test that says so.

## Before

Run against the live `releases.atom`, which is the exact string a 0.4.4 client digests:

```
v0.4.5 →  1. { box-sizing: border-box; }
v0.4.4 →  1. Munder Difflin v0.4.4 A local hive of … Grok &amp; Copilot…
          2. page { display: none; height: 100%; flex-direction: column; padding: clamp(24px…
          3. content { flex: 1; min-height: 0; overflow-y: auto; } .center {…
```

## After

Same feed, same function, this branch:

```
v0.4.5 →  1. Munder Difflin v0.4.5 A local hive of Claude Code, Antigravity, Codex,
             Gemini, Cursor, Grok & Copilot agents…
          2. Munder Difflin · v0.4.5 Thank you for being here early. Every bug you
             filed, every PR you sent, every floor…
          3. Also new Triggers Schedule on weekdays at a time of day Not…

v0.4.4 →  1. Munder Difflin v0.4.4 A local hive of Claude Code, Antigravity, Codex,
             Grok & Copilot agents that run…
          2. Munder Difflin 0.4.4 The release where Windows finally joined the floor —
             and the first run stopped quietly…
          3. Start → The headline Agents can talk to each other on Windows…
```

## Testing

`npm run test:focused` — 556 pass, 0 fail. `npm run typecheck` — clean on both
projects. Four new tests, and the fixture is a **verbatim slice of GitHub's rendered
feed**, not hand-written. Every existing test in that file passes on hand-written
markdown, and every one of them passed while the CSS prompt was live. A hand-written
fixture would reproduce the same blind spot that let this ship twice.

## Why the evidence check is red and not waived

There is a real visible surface here: the toast's "What's new" block. An honest
before/after of it needs two builds side by side, which I have not produced. The
`no-visual-change` label would therefore be dishonest — the standard is whether a
rendered surface changed, not whether capturing it is convenient, and this one
changed. The check staying red is also a useful second lock while this must not
merge. It should be satisfied with a real screenshot pair before anyone considers
merging, not labelled away.

## Not in this PR

`RELEASE.md` still has to be rewritten for 0.4.6 through the release-body rules, or
0.4.6 reintroduces the CSS prompt on day one — the digest always reads the newest
release. That work is tracked separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,38 @@ All notable changes to this project are documented here. The format is based on
 
 ## [Unreleased]
 
+## [0.4.6] — 2026-08-22
+
+**The update that installs itself, and tells the truth while it asks.** 0.4.5 shipped a restart
+prompt whose "What's new" read `{ box-sizing: border-box; }`, and a first-run release page that
+could not fire on the release that introduced it. Both are fixed, and an update no longer waits
+on a button press that most people never make.
+
+### Fixed
+
+- **A downloaded update installs when you quit.** `autoInstallOnAppQuit` was off, so a staged
+  update sat until someone pressed "restart to update" and was replaced by the next one if they
+  never did. On the existing install base only 41.7% took 0.4.4 within four days, while 56.7%
+  stayed on 0.4.3. The explicit restart is still the fast path and still the only thing that
+  interrupts; quitting now finishes the job for everyone else. Nothing installs under a running
+  app.
+- **The first run after an update shows that release's page, including for anyone coming from a
+  build older than the stamp.** The version stamp shipped in 0.4.5, so every user upgrading from
+  0.4.4 arrived with no stamp and was read as a brand-new install. `telemetry-install-id` has
+  been minted on first run since 0.4.2, so finding one with no stamp beside it proves an older
+  build already ran here. Installs with telemetry off have no id and stay quiet, which is the
+  honest answer: they cannot be told apart from a genuinely new install.
+- **The restart prompt no longer shows CSS.** Its notes came from electron-updater, which builds
+  them from `releases.atom` — GitHub's rendered markdown, where HTML comments do not survive and
+  a `<style>` block leaves its declarations behind as prose. `* { box-sizing: border-box; }` is
+  character-for-character a markdown bullet, and it was the only one in the entire feed, so it
+  beat the whole release note. The notes are now read from the release itself, unrendered. A
+  failed read leaves no notes at all, deliberately: a missing digest beats a digest made of
+  somebody's stylesheet.
+- **A dev run can no longer eat the installed app's update signal.** Dev shares its user-data
+  directory with the packaged app, and a dev build stamping the new version hours before the real
+  upgrade made that upgrade read as no change at all. Only a packaged build writes the stamp.
+
 ## [0.4.5] — 2026-08-22
 
 **The release that fixes the things you trusted and were quietly wrong.** Cost reporting was off
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "munder-difflin",
-  "version": "0.4.5",
+  "version": "0.4.6",
   "private": true,
   "description": "Clones for you and your team, working 24/7 — a free, local-first multi-agent harness that turns the CLI coding agent you already run into a clone of you, coordinating an office of agents on your machine.",
   "main": "out/main/index.js",
--- a/src/main/updater.ts
+++ b/src/main/updater.ts
@@ -1,7 +1,7 @@
 import { app, ipcMain, shell } from 'electron';
 import type { WebContents } from 'electron';
 import { request as httpsRequest } from 'node:https';
-import { appendFileSync, mkdirSync, readFileSync, writeFileSync } from 'node:fs';
+import { appendFileSync, existsSync, mkdirSync, readFileSync, writeFileSync } from 'node:fs';
 import { join } from 'node:path';
 import { readConfig } from './config';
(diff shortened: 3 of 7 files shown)


chaitanyagiri:

---
Two of the three things here are strong. **The third has a false positive I can demonstrate.**

The dev-build diagnosis is excellent: a dev run shares `userData` with the installed app, so it
stamped 0.4.5 hours before the real 0.4.4 to 0.4.5 upgrade and consumed the one-shot signal.
Gating the WRITE on `app.isPackaged` while still reading in dev is exactly right. So is
`A detector that only fires on a stamp it wrote itself can never fire on the release that
introduces it`, and using `telemetry-install-id` to settle the ambiguity, and being honest that a
telemetry-off profile stays quiet because it genuinely cannot be told apart from a new install.

**`CSS_SELECTOR_RE` matches ordinary release prose.** I ran it:

```
"- Fixed the picker,"                -> true
"* Faster startup, smaller bundle,"  -> true
"Two changes land here,"             -> true
"  #pg1:checked ~ .stage .p1,"       -> true   (the intended case)
```

The character class is `[\w\-.#:>~+*\[\]="'(),\s]`. **I first described that as letters, spaces,
hyphens, periods and commas; Angela recounted it and it is wider than that** — it also carries
`:` `>` `~` `+` `*` `[` `]` `=`, quotes and parens, so it swallows more prose than I said, not
less. `- Fixed the picker (see #241),` matches, and so does `- Added David Wallace to the cast,`.
**The trailing comma is the entire trigger**: `Faster startup, smaller bundle, fewer crashes.`
survives only because it ends in a period. In `usableLines`
that hits `if (looksLikeCssOpener(probe)) { cssDepth = …; continue; }` with `braceDelta` of 0, so
the line is dropped and the rest of the digest carries on. A release note whose line happens to end
in a comma silently loses that line, in the function whose entire job is to keep the real ones.

A CSS selector list that is continued across lines is always followed by more selector lines and
then an opening brace. Requiring that, or requiring at least one character that prose does not
use, `{`, `}`, `;` or a `prop: value`, would keep the intended case and drop the false positives.
The other three predicates look tight to me: `CSS_OPEN_LINE_RE` needs the brace,
`CSS_BLOCK_START_RE` needs a declaration after it, and the at-rule list is explicit.

**Third thing, on packaging rather than code.** This PR flips `autoInstallOnAppQuit` to `true`,
which changes how every user receives every future release, and it does it in the same commit as
the notes parser and the stamp fix. Reverting any one of those reverts all three. The reasoning
for the flip is sound, a staged installer nobody presses a button for is just replaced by the next
one, but it is the change with the widest blast radius in the batch and it would be easier to
reason about on its own.

PR #243, opened 2026-08-22
feat(ask me): the question renders as markdown, and agents are told to write it

L422Y:
## What & why

An ASK ME question was printed as raw text. Agents were creating very hard to read questions, without any formatting, so I prompted so that they created markdown formatted questions, which was better, but the ASK ME view didn't support Markdown, so a card that said `**Apply**` or wrapped a path in backticks put
those characters on screen, and a question with three numbered options arrived as a
paragraph of punctuation. The question is the one thing on that card a human has to
read carefully, and it was the worst-formatted text in the app.

It now goes through the same `MarkdownPreview` the file preview and the IDE already
use, via a new `card` variant: it inherits the host card's mono face and size instead
of imposing the document type scale, and drops the 72ch measure and page padding so a
one-line question stays one line. The task detail's Q&A trail renders the same way,
answers included, since that is where the "view earlier answers" link lands.

Rendering it is only half the fix. Nothing had ever told the agents to format these,
so the orchestrator prompt and `PROTOCOL.md` now do, and `PROTOCOL.md` gains the ASK ME
section it never had (the `humanQA` contract was only documented in a source comment).
Shipping the renderer without that leaves the feature inert, which is why both are here.

Two details worth flagging for review:

- **No new dependency.** `react-markdown` and `remark-gfm` were already in the tree.
  CommonMark folds a single newline into a space, which would have turned every existing
  plain-text question into a run-on paragraph, so the `card` variant runs a twelve-line
  local remark plugin that splits `text` nodes on `\n` into `break` nodes. That is what
  `remark-breaks` does, and it did not seem worth a package. Code blocks are `code` nodes
  with a raw `value`, so their newlines are untouched.
- **Still no `rehype-raw`.** Agent-written markdown is untrusted and the existing file
  header says to keep it that way. Raw HTML in a question renders as text.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

## Evidence

Both shots are the same view, same 860x800 viewport, same dark theme, same two task
cards, differing only by this commit. The first card is a question written in markdown;
the second is deliberately plain prose with single newlines, to show that a question
with no markdown in it renders exactly as it did before.

### Before
ASK ME on main: the question shows its asterisks and backticks literally

### After
ASK ME with this change: bold, inline code and a numbered list render, and the plain-text card is unchanged

## How I tested it

- OS: macOS (Darwin 25.3), Node 24.

- Steps:
  - `npm run typecheck` (both projects), `npm run test:focused` (552 pass), `npm run build`.
  - For the screenshots: mounted the real `AskMeTab` in a throwaway page outside the repo,
    against a stubbed `window.cth` returning a fixed four-task ledger, with the app's own
    `tokens.css` and `global.css` and the same Google fonts. Shot it in Chromium at one
    fixed viewport on `main`, then on this branch, changing nothing but the checkout. It is
    the real component and the real stylesheet, not the whole Electron window, so the
    Command Center tab bar is not in frame.
  - Checked the soft-break plugin against real mdast from `mdast-util-from-markdown`: a
    soft newline inside a paragraph becomes a `break`, and the newlines inside a fenced
    code block survive.
  - Rendered `PROTOCOL_MD` out of its template literal to confirm the escaped backticks
    and the fenced JSON block come out right in the file agents actually read.

What I have not done is run it in the packaged app against a live hive.

## Discord (optional)

Discord:

## Checklist

      spacing, or fonts. The card variant adds no colors; every size in it is
      relative, so the host card sets the scale.
      `ATTRIBUTION.md`. (No art.)


L422Y:
Unrelated to the diff, but worth knowing since it cost me a round trip: the `PR evidence` check fails when the image sits a blank line below its heading, which is exactly what GitHub's drag-and-drop produces.

The section regex in `pr-evidence.yml` uses the `m` flag, so `$` in `(?=\n#{1,6}\s|$)` matches at the first end-of-line rather than end-of-string. The lazy group then stops immediately and the section captures an empty string:

```js
section("### Before\n\n![a](x.png)")  // "" -> fails
section("### Before\n![a](x.png)")    // "![a](x.png)" -> passes
```

The failure mode is confusing rather than obvious: you get the near-miss message telling you both assets are in one place, when they are already under their own headings. Dropping the `m` flag, or using `(?=\n#{1,6}\s|$(?![\r\n]))`, fixes it. Happy to open that as its own PR if it is useful.


EDIT: looks like https://github.com/chaitanyagiri/munder-difflin/pull/318 will address it

chaitanyagiri:
Shipped in v0.4.6, and it is called out in the release notes.

Your commit `9edf34ea` is in. The ASK ME card renders markdown, the Q and A trail in the task detail renders too, and the orchestrator prompt now asks agents to format what they write. Raw HTML still renders as text, which is the part that keeps agent written markdown safe to show.

Thank you.

PR #248, opened 2026-08-22
fix(office): retry a floor rebuild that cannot get a WebGL context

djbiz:
## What & why

When the GPU process dies — a driver reset, a TDR, a Chromium GPU crash — the floor loses its context and `glRecovery` rebuilds it 1500ms later, exactly as designed. If the GPU process has not come back by then, `getContext()` returns null, Pixi turns that into `Error: This browser does not support WebGL. Try using the canvas renderer`, and `OfficeFloor`'s `init().catch` painted that stack onto the floor and stopped.

So a condition that usually clears itself in a second or two cost a full app restart, behind a message that blames the browser for something the browser did not do. This is the failure I hit on v0.4.5 (Windows 11, Intel UHD, Electron 32) — the office came up as a wall of minified PixiJS frames and stayed that way.

The fix classifies that specific rejection and rebuilds through the existing generation-bump path, with the retry budget in a ref so it survives the rebuilds it schedules. Out of budget, it shows a note naming the actual cause instead of a stack. Anything **not** recognised as a context failure still reports its stack, so a broken theme or tileset is not hidden behind six seconds of retries.

Worth stating explicitly, because it surprised me: **this is not the eviction path.** I first assumed context starvation caused it, and that turned out to be wrong. A `getContext()` that pushes past Chromium's ~16-context cap evicts somebody else and *succeeds*, so the floor's rebuild always wins its slot back. I verified that over CDP — 24 held contexts plus a steady drip every 250ms produced the eviction warning, `[OfficeFloor] WebGL context lost ... rebuilding`, and then a healthy floor. It takes an absent GPU process for the request itself to fail. The comments in `glRecovery.ts` and the test header record this so the next person does not re-derive it.

The policy lives in `glRecovery.ts` next to the loss policy, as a pure classifier and planner, so it is testable with no browser, no GPU and no Pixi — same as the existing path.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

## Evidence

Reproduced over CDP against a **shipped v0.4.5 build**, using the same technique the repo used to confirm the original blank-office bug. Steps for both shots, identical apart from the build under test:

1. Launch with `--user-data-dir` pointing at a scratch config (empty hive, `autoMode: false`) and `--remote-debugging-port`.
2. Open the config, wait for the floor to render (`[OfficeFloor] map 34x22, 1147 tile sprites rendered`).
3. Kill the app's `--type=gpu-process` five times, three seconds apart, so Chromium's GPU crash limit is reached and the GPU stays down.
4. Wait 15s, screenshot.

The floor immediately before step 3, in both runs:

healthy office floor

### Before
before: a wall of PixiJS stack trace where the office floor should be

v0.4.5 as shipped. The floor is replaced by the PixiJS stack, and stays that way until the app is restarted.

```
[warning] [OfficeFloor] WebGL context lost (Chromium evicted the oldest context) — rebuilding the scene, attempt 1/3
[error]   [OfficeFloor] init failed: Error: This browser does not support WebGL. Try using the canvas renderer
    at _GlContextSystem2.createContext (.../assets/WebGLRenderer-Dq0SMKmP.js:369:15)
    at _GlContextSystem2.init (.../assets/WebGLRenderer-Dq0SMKmP.js:315:12)
    at WebGLRenderer.init (.../assets/index-fwsZ7mgl.js:20389:41)
    at async autoDetectRenderer (.../assets/index-fwsZ7mgl.js:20771:3)
    at async _Application2.init (.../assets/index-fwsZ7mgl.js:20962:21)
    at async init2 (.../assets/index-fwsZ7mgl.js:38254:7)
```

### After
after: a readable note explaining the GPU context could not be obtained

Same steps, same window size, same theme, same scratch config, this branch. Three retries, then a note that says what happened and what to do.

```
[warning] [OfficeFloor] WebGL context lost (Chromium evicted the oldest context) — rebuilding the scene, attempt 1/3
[warning] [OfficeFloor] could not get a WebGL context (the GPU process may be restarting) — retrying, attempt 1/3
[warning] [OfficeFloor] could not get a WebGL context (the GPU process may be restarting) — retrying, attempt 2/3
[warning] [OfficeFloor] could not get a WebGL context (the GPU process may be restarting) — retrying, attempt 3/3
[error]   [OfficeFloor] still no WebGL context after 3 retries: Error: This browser does not support WebGL...
```

This shot is the *worst* case, where the GPU is gone for the rest of the session and the retries cannot win. In the commoner case — the GPU process restarts within the retry window — the floor comes back on its own and there is nothing to screenshot, which is the point of the change.

## Tests

Five tests added to `test/office-gl-recovery.test.cjs`, written before the code, covering: Pixi's exact wording and the wider family classified as a context failure; a broken theme or tileset explicitly **not** classified as one; retry with a delay rather than a stack; the cap and the give-up; and a real init bug reported immediately and never retried.

Run on Windows 11, Node 24:

- `npm run typecheck` — clean.
- `npm run build` — clean.
- `npm run test:focused` — 554 tests, 541 pass, 11 fail. Those 11 are environmental on my machine, not from this change: `main` at the same commit gives 549 / 536 / **the same 11** (`node_modules/electron` failed its binary download here, so the electron-dependent tests cannot run). This branch is +5 tests, +5 passing, no change in failures.

### DIFF
--- a/src/renderer/src/scene/office/OfficeFloor.tsx
+++ b/src/renderer/src/scene/office/OfficeFloor.tsx
@@ -12,7 +12,9 @@ import { hexToNumber, DEFAULT_CHARACTER } from './cast';
 import { pickSoloLine, pickExchange, type BreakSpot } from './cafeteriaLines';
 import { colors } from '@/design/tokens';
 import { loadTheme, resolveThemeMap, themeTilesetUrls } from './themeLoader';
-import { installContextLossRecovery } from './glRecovery';
+import {
+  installContextLossRecovery, planInitFailure, DEFAULT_MAX_INIT_RETRIES
+} from './glRecovery';
 import type { Tile, Facing, ErrandKind, ErrandSpot } from './themeRegistry';
 
 // The map, tileset atlases, desk-claim order, errand spots, coffee-economy
@@ -164,6 +166,10 @@ export function OfficeFloor() {
   // whole scene is torn down and rebuilt through the existing mount path rather
   // than through a second, parallel recovery routine.
   const [glGeneration, setGlGeneration] = useState(0);
+  // Retries spent on an init that could not GET a context (see glRecovery.ts).
+  // A ref, not state: the budget has to survive the rebuilds it schedules, which
+  // re-run the effect below and would reset anything scoped to it.
+  const initRetriesRef = useRef(0);
   // The active office theme (store mirror of config.officeTheme). Changing it
   // tears down and rebuilds the whole scene on the new map/cast (see deps below).
   const officeTheme = useStore((s) => s.officeTheme);
@@ -258,15 +264,10 @@ export function OfficeFloor() {
         onRebuild: () => { if (mountIdRef.current === mountId) setGlGeneration((n) => n + 1); },
         onGiveUp: () => {
           if (mountIdRef.current !== mountId) return;
-          const note = document.createElement('div');
-          note.style.cssText =
-            'position:absolute;inset:0;display:flex;align-items:center;justify-content:center;' +
-            'padding:24px;color:#ffd0b5;font-family:monospace;font-size:13px;text-align:center;white-space:pre-wrap;';
-          note.textContent =
+          host.appendChild(floorNote(
             'The office floor lost its GPU context.\n\n' +
             'Too many terminals are using the GPU at once.\n' +
-            'Close a few agent terminals, or restart the app, to bring it back.';
-          host.appendChild(note);
+            'Close a few agent terminals, or restart the app, to bring it back.'));
         }
       });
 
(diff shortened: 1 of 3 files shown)


chaitanyagiri:
Shipped in v0.4.6.

Your commit `1155c344` is in the release: the floor retries a rebuild when it cannot get a WebGL context, with `src/renderer/src/scene/office/glRecovery.ts` and the tests in `test/office-gl-recovery.test.cjs`.

Worth noting it landed alongside #323, which releases the WebGL context when a terminal renderer is torn down. The two fixes sit either side of the same problem.

Thank you.

PR #249, opened 2026-08-22
fix(copilot): bind the GitHub host to the agent's repo so *.ghe.com agents stop getting a github.com token

marclundgren:
## What & why

A Copilot agent spawned in a **GitHub Enterprise Cloud** tenant checkout
(`microsoft.ghe.com` — tenant cloud with data residency, not GHES, not github.com)
fails with `401 unauthorized: AuthenticateToken authentication failed` and an empty
model catalog, even when the user is correctly signed in to that tenant.

The Copilot CLI resolves its **host** and its **token** from two unrelated env lookups.
The token is the first non-empty of `COPILOT_GITHUB_TOKEN` / `GH_TOKEN` / `GITHUB_TOKEN`
with no awareness of which host is active — so a github.com PAT sitting in `GH_TOKEN`
(which `gh` exports, and which any dotcom checkout on the machine leaves lying around)
is transmitted to `api.<tenant>.ghe.com`, where it can only ever 401. It also silently
outranks the correctly-scoped keyring credential, so the CLI fails while holding the
right key. A tenant is a fully separate identity plane; tokens are not portable between
it and github.com in either direction.

We spawn `copilot` as a child PTY, so the part of this we own is **the environment we
hand it**. At spawn time a copilot agent now:

- resolves its host from `COPILOT_GH_HOST` → `GH_HOST` → **its own git remote** → `github.com`
  (the remote is the new signal; it is what makes `cd`-ing between a dotcom repo and a
  tenant repo work with no re-login and no env juggling);
- binds `COPILOT_GH_HOST`/`GH_HOST` so the CLI — and any `gh` the agent shells out to —
  agrees with the repo it is sitting in rather than an ambient default;
- passes through **only a credential actually scoped to that host**. `GH_ENTERPRISE_TOKEN`
  joins the lookup for non-dotcom hosts and is carried over **as `GH_TOKEN`**, because
  against tenant cloud `GH_ENTERPRISE_TOKEN` itself returns 401 (that variable targets GHES);
- **removes** every GitHub token from the child when nothing in the environment is scoped
  to the host, rather than blanking it, so the keyring credential wins;
- logs one line naming the active host, where the host came from, which variable supplied
  the credential, and what was refused and why. Token values never appear in it.

Precedence, documented in `githubHost.ts` and pinned by tests — enterprise host, highest first:
`COPILOT_GITHUB_TOKEN` → `GH_ENTERPRISE_TOKEN` → `GH_TOKEN` → `GITHUB_TOKEN` → keyring.
The three host-agnostic names are only eligible when `COPILOT_GH_HOST`/`GH_HOST` names that
same host, which is the only evidence available that they were meant for a tenant.

**github.com is untouched.** For a dotcom host the plan comes back empty — no set, no omit —
so the default path is what it always was. That is asserted directly by a test.

Scoped to the `copilot` provider on purpose: every other engine authenticates elsewhere, and
rewriting `GH_*` for all of them is a much larger blast radius than the bug.

<details>
<summary>What this deliberately does <b>not</b> fix (upstream-only)</summary>

Three items from the original report live inside the Copilot CLI and cannot be reached from
the harness: normalizing its keyring account key to scheme+host (the observed key,
`https://microsoft.ghe.com/bic/:user`, carries the **org path**, so two orgs on one tenant
become two entries for one identity), turning `model_count: 0` into a distinct actionable
error instead of a generic "Authentication failed", and a `copilot auth status` diagnostic.
The log line added here covers the harness-side half of the diagnostic gap.
</details>

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

## Evidence

Same repro on both sides: a checkout whose remote is `https://microsoft.ghe.com/bic/…`, a
github.com PAT exported as `GH_TOKEN`, and a `copilot` agent spawned into it. The images are
the real output of a repro harness that builds the child environment through the actual
`buildPtyEnv` code path on each branch — only the branch differs.

### Before
Before — a github.com PAT is handed to the tenant

The child gets `GH_TOKEN=ghp_…`, no host is bound, and the CLI defaults to `github.com` while
the repo is a tenant. That dotcom PAT is what reaches `api.microsoft.ghe.com` — the 401.

### After
After — the host is bound to the repo and no dotcom credential is forwarded

The host is bound to the repo's tenant, the mismatched PAT is removed rather than blanked so
the CLI falls through to its keyring entry for that tenant, and the log line says exactly
what happened and how to override it.

## How I tested it

- **OS:** Linux (WSL2, Ubuntu), Node 26.3.0
- **Steps:**
  1. `npm run typecheck` — clean (both the node and web TS projects).
  2. `npm run test:focused` — **553 tests, 549 pass, 4 fail**. Those 4 are pre-existing and
     environmental, not from this change: an identical run on `main` with the same
     `node_modules` gives **531 tests, 527 pass, the same 4 fail**
     (`hive-runtime-path`, `update-download-asset`, `win-cmd-shim` — this checkout was
     installed with `--ignore-scripts`, so `node-pty`'s native module was never rebuilt and
     Electron was never downloaded). The 22 new tests are the whole delta.
  3. `npm run build` — succeeds (electron-vite build + main-asset copy).
  4. Ran the repro harness above on `main` and on this branch to produce the two images.
- **New tests:** `test/github-host.test.cjs` (19) covers remote parsing for every remote
  shape git accepts, host classification and precedence, the dotcom-PAT-into-a-tenant case,
  the `GH_ENTERPRISE_TOKEN` carry-over, that no value we set points at dotcom while a tenant
  is active, that github.com produces an empty plan for every token combination, and that the
  diagnostic never contains a token value. `test/pty-env.test.cjs` (+3) covers the new
  `omitEnv` layer: an omitted name is **absent**, not empty, and a deliberate per-agent value
  still outranks the omit list.

I do not have access to a `*.ghe.com` tenant, so the end-to-end "model catalog loads" leg is
not something I can show. The evidence above is the part that is verifiable from here: which
host is bound, and which credential does and does not reach the child.

## Checklist

- [x] `npm run test:focused` passes (see the note on the 4 pre-existing failures above).
- [x] Any new UI derives from `DESIGN.md` / `tokens.ts` — no new UI in this PR.
- [x] If I added art, it's my own or compatibly licensed — no art in this PR.

### DIFF
--- a/src/main/git.ts
+++ b/src/main/git.ts
@@ -144,6 +144,18 @@ export async function isRepo(cwd: string): Promise<boolean> {
   return res.ok && res.stdout.trim() === 'true';
 }
 
+/** The fetch URL of a remote (default `origin`), or null when there is no such
+ *  remote — or no repo at all. Used at spawn time to learn WHICH GitHub an agent
+ *  is sitting in, so a `*.ghe.com` worker is not handed a github.com credential
+ *  (shared/githubHost.ts). Short timeout and null-on-failure: this is a hint that
+ *  refines a spawn, never a gate that can block one. `--get` is deliberate — it
+ *  reads the config file and, unlike `git remote get-url`, never touches the
+ *  network. */
+export async function getRemoteUrl(cwd: string, remote = 'origin'): Promise<string | null> {
+  const res = await runGit(cwd, ['config', '--get', `remote.${remote}.url`], 3000);
+  return res.ok && res.stdout.trim() ? res.stdout.trim() : null;
+}
+
 const MAX_DIFF_BYTES = 2 * 1024 * 1024; // 2 MB — keep the diff view responsive
 
 /** A file's two sides for a working-tree-vs-HEAD diff. `head` is the committed
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -22,7 +22,7 @@ import { normalizeWeekly, weeklyDelayMs } from '../shared/weeklySchedule';
 import {
   getBranch, getStatus, getLog, getBranches, getAheadBehind, isRepo, getDiff, mainRepoRoot,
   addWorktree, removeWorktree, worktreeHasUnintegratedWork, worktreeIsGcSafe,
-  getLogGraph, getCommitFiles, getFileAtRev, compareRefs, listWorktrees, checkoutRef
+  getLogGraph, getCommitFiles, getFileAtRev, compareRefs, listWorktrees, checkoutRef, getRemoteUrl
 } from './git';
 import { HiveManager, type AgentMeta, type HiveMessage, type HiveTask } from './hive';
 import { HookServer } from './hooks';
@@ -76,6 +76,7 @@ import {
   installInfoForProvider,
   type AgentProvider
 } from '../shared/agentProvider';
+import { planCopilotGithubEnv, describeGithubEnvPlan } from '../shared/githubHost';
 import { buildMissingCliScript, chooseInstallRung } from './cliInstall';
 import { detectNodeVersion, nodeIsUsable, resolveNodeInstaller } from './nodeInstall';
 import { toolCatalog, type ToolStatus } from '../shared/toolCatalog';
@@ -2877,6 +2878,32 @@ async function spawnAgentCore(opts: AgentSpawnOptions, owner: Electron.WebConten
       extra.OPENCODE_CONFIG_CONTENT = JSON.stringify(oc);
     }
     opts.env = { ...(opts.env ?? {}), ...extra };
+  }
+  // ── GitHub host binding for Copilot agents (GHE Cloud, *.ghe.com) ───────────
+  // The Copilot CLI resolves its host and its token from separate, unrelated
+  // env lookups, so a github.com PAT in GH_TOKEN — which `gh` exports and which
+  // any dotcom checkout on this machine leaves lying around — is transmitted to
+  // `api.<tenant>.ghe.com`, where it can only ever 401. Worse, that env token
+  // outranks the correctly-scoped keyring credential the user already logged in
+  // with, so the CLI fails while holding the right key. Bind the host to THIS
+  // agent's repo and hand the child only a credential actually scoped to it;
+  // when nothing is, hand it none so the keyring wins.
+  //
+  // Copilot-only on purpose. Every other engine authenticates elsewhere, and
+  // rewriting GH_* for all of them is a much bigger blast radius than the bug.
+  // github.com is untouched: the plan comes back empty for a dotcom host, so the
+  // default path is byte-for-byte what it was.
+  if (provider === 'copilot') {
+    const remoteUrl = await getRemoteUrl(opts.cwd).catch(() => null);
+    const plan = planCopilotGithubEnv({ env: process.env, remoteUrl });
+    if (Object.keys(plan.set).length > 0 || plan.omit.length > 0) {
+      // The plan sits UNDER opts.env: an explicit per-agent override still wins.
+      opts.env = { ...plan.set, ...(opts.env ?? {}) };
+      opts.omitEnv = [...(opts.omitEnv ?? []), ...plan.omit];
+      // Names only, never values — this line is what people paste into issues,
+      // and it answers the questions a bare "Authentication failed" does not.
+      console.log(`[copilot] ${describeGithubEnvPlan(plan)}`);
+    }
   }
   // Codex Remote is daemon-based (there is no `/remote-control` slash command).
   // Start/enable the daemon under this agent's isolated CODEX_HOME and connect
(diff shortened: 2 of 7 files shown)


chaitanyagiri:
## TL;DR

Thanks @marclundgren. This is a careful PR and the bug behind it is a good catch: an agent working in a `*.ghe.com` repo inheriting github.com credentials from the environment is the kind of thing that fails confusingly, or worse, quietly succeeds against the wrong host.

**What it does:** reads the repo's `origin` remote, works out which GitHub host the agent should be talking to, and sets the matching `GITHUB_*` environment for a `copilot` agent. Adds an `omitEnv` mechanism so variables are deleted from the inherited environment before the child process sees them, rather than merely overwritten.

**Verdict: needs work, and the work is verification rather than repair.** I did not find a defect. I also did not read all 339 lines of the new host matching logic closely enough to sign it off, and I want to be straight about that rather than approve on vibes.

---

## What I did

Static review of the diff against `origin/main` @ `eb2818a8`, focused on the integration points: `getRemoteUrl` in `src/main/git.ts`, the `copilot` branch in `src/main/index.ts`, and the new `omitEnv` handling in `src/main/pty.ts` and `src/main/ptyEnv.ts`. **I did not check out your branch and I did not run your tests**, so the 236 line test file is unverified by me.

## What I like

**`omitEnv` is the right primitive.** Overwriting a credential variable with an empty string and deleting it are different things, and plenty of tooling treats "set but empty" as "set". Deleting before the child sees it is the correct semantics and it is generally useful beyond this feature.

**Putting the decision in a pure module is the right shape.** `src/shared/githubHost.ts` takes `{ env, remoteUrl }` and returns a plan, and `describeGithubEnvPlan` renders it for the log. That means the interesting logic is testable without a PTY, which is why you were able to write 236 lines of test for it. This is how the rest of this codebase's better modules are built.

**`getRemoteUrl` is appropriately boring.** `git config --get remote.origin.url` with a 3 second timeout and a null on failure. Good.

## What I want a maintainer to check before this lands

1. **The host matching itself.** `githubHost.ts` is 339 lines of new logic in a security relevant position, since getting it wrong means sending a token to the wrong host. The SCP style remote regex in particular (`/^(?:[^@/]+@)?([^:/]+\.[^:/]+):/`) is doing real work on user controlled input. I read it. I did not convince myself it is exhaustive across every remote form in the wild, including `insteadOf` rewrites and non standard ports.

2. **What happens when there is no remote, or several.** Agents in this app frequently run in a fresh worktree or a directory that is not a repo at all. `getRemoteUrl` returns null there, and I would like the plan for that case stated in the description rather than inferred from the code.

3. **The log line.** `console.log('[copilot] ' + describeGithubEnvPlan(plan))` runs on every copilot spawn. Please confirm it can never render a token value, even truncated. This is the one place where a mistake would be genuinely costly and it is a thirty second check.

## Smaller notes

- `opts.env = { ...plan.set, ...(opts.env ?? {}) }` puts existing `opts.env` last, so an explicitly passed value wins over the plan. That looks deliberate and correct. Worth a one line comment saying so, because the next reader will wonder.
- The feature is gated on `provider === 'copilot'`. If the intent is that this generalises later, a note in the description about why it is scoped now would help.

1. A maintainer read of the remote parsing in `githubHost.ts`, or a short note from you listing the remote forms you tested against.
2. Confirmation that the log line cannot print a secret.
3. A line in the description about the no remote and non repo cases.

Nothing here is a blocker in the sense of "this is broken". It is a blocker in the sense of "this touches credentials and deserves a second pair of eyes on the matching logic before it goes in".

chaitanyagiri:

---
**Scope of this comment: I read `src/shared/githubHost.ts` closely and skimmed the `ptyEnv` and
`index.ts` halves. Take the below as a review of the resolution logic, not of the wiring.**

The problem statement is the best I have read in this batch. The Copilot CLI resolving its host
and its token independently, so a github.com PAT in `GH_TOKEN` gets transmitted to
`api.<tenant>.ghe.com` and the user sees a generic failure naming neither, is a real leak
described precisely. Keeping the module dependency free so the resolution is a pure function is
the right shape for something this fiddly.

`parseGitRemoteHost` correctly returns only the hostname from
`https://user:tok@host/org/repo`, so a remote with an embedded credential cannot leak through it.
`isGheCloudHost` excluding bare `ghe.com` while requiring the `.ghe.com` suffix is right, and it
does not match `ghe.com.attacker.example`.

**One case falls through in the direction the module exists to prevent.**

```
const scp = /^(?:[^@/]+@)?([^:/]+\.[^:/]+):/.exec(raw);
```

The dot requirement is there to stop `C:\src\repo` reading as a host named `c`, and the comment
says so. It also excludes every single-label hostname, which is legal and common on an internal
network. `git@gitserver:org/repo.git` returns null, falls through to the default, and the agent is
handed a github.com host for a repo that lives on an internal GHES. That is the same class of
mismatch the module was written to stop, just narrower. A Windows drive letter is one character;
requiring more than one before the colon would keep the drive-letter guard and stop excluding
internal hosts.

Smaller question: `normalizeHost` strips a trailing `:port`. A GHES on a non-default port then
produces an `origin` without it, so anything built from that origin points somewhere else. If the
only consumer is the keyring key shape, that may be exactly right. Worth a sentence saying which.

PR #252, opened 2026-08-23
build: scope postinstall electron-rebuild to better-sqlite3 (stop force-rebuilding node-pty)

cyberfalcon-us:
## What & why

The `postinstall` runs `electron-rebuild -f`, which force-rebuilds **every** native module. But `node-pty` 1.1.0 is Node-API (`node-addon-api ^7`) and ships complete prebuilds (`win32-x64`/`arm64` with conpty + winpty binaries, `darwin-x64`/`arm64` with `spawn-helper`), so it needs no Electron-ABI rebuild — and its loader (`lib/utils.js`: `build/Release` → `build/Debug` → `prebuilds/<platform>-<arch>`) already ends up running the prebuild anyway, as the header comment in `tools/ensure-pty-perms.cjs` documents. Forcing the rebuild is exactly what the "most common setup failure" warning in CONTRIBUTING is about: on Windows, winpty's vcxproj requires the VS "Spectre-mitigated libs" component (MSB8040) and its gyp configure runs `cd shared && GetCommitHash.bat`, which breaks under the `NoDefaultCurrentDirectoryInExePath=1` hardening some environments set.

`better-sqlite3` is V8/NAN and genuinely does need the Electron rebuild — so scope the rebuild to it:

```
electron-rebuild -f --only better-sqlite3 && node tools/ensure-pty-perms.cjs && node tools/patch-node-pty-conpty.cjs
```

Cross-platform notes:

- **Windows / macOS** — node-pty runs from its shipped prebuilds (as before). `tools/ensure-pty-perms.cjs` walks `build/Release|Debug`, `prebuilds/*`, and `bin/*` for `spawn-helper`, so its macOS fix-up is unaffected.
- **Linux** — node-pty ships no Linux prebuilds; npm's own install step (`node scripts/prebuild.js || node-gyp rebuild`) already compiles it from source, and being Node-API that build loads fine under Electron. Toolchain requirements there are unchanged.
- **Packaging** — unaffected: `electron-builder.yml` has `npmRebuild: true`, so `npm run dist` / the release workflow still rebuilds natives for the target exactly as today.
- Also syncs the stale `package-lock.json` version stamp the 0.4.5 bump left behind (0.4.4 → 0.4.5) and updates the README/CONTRIBUTING/ci.yml comments that described the old behavior.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [x] Build / CI

## Evidence

Verified on Windows 10 Pro (VS2022 **without** Spectre-mitigated libs, `NoDefaultCurrentDirectoryInExePath=1` set), Node 20.20.2 / npm 10.8.2 — i.e. a machine where main's postinstall cannot complete.

### Before
before — forced node-pty electron-rebuild fails

The node-pty Electron rebuild that `electron-rebuild -f` forces (isolated here with `--only node-pty` for a fast repro) fails before it can even configure — and with the hardening var unset it fails at MSB8040 instead. Either way `npm install` on main dies here.

### After
after — scoped rebuild install + PTY roundtrip verified

Clean checkout (`node_modules` deleted), same shell, same hardening var still set: `npm install` completes end-to-end with the scoped rebuild, `npm run dev` boots (better-sqlite3 loads at startup), and a PTY spawned through the app's own `pty:spawn` IPC does a full write/read roundtrip off the win32-x64 prebuild. The dev app running after that install:

after — dev app onboarding screen

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
   build:
     name: Build
     runs-on: macos-latest
-    # Build exercises the native node-pty rebuild; allowed to fail without blocking.
+    # Build exercises the native better-sqlite3 Electron rebuild; allowed to fail without blocking.
     continue-on-error: true
     steps:
       - uses: actions/checkout@v4
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,8 @@ participating, you agree to uphold it.
   Cross-platform smoke-testing and fixes are still very welcome (see
   [Good first areas](#good-first-areas)).
 - **Node.js 18+** and npm.
-- A **C/C++ toolchain** to build `node-pty`'s native addon. On macOS:
+- A **C/C++ toolchain** to build `better-sqlite3` for Electron (and, on Linux
+  only, `node-pty` — it ships prebuilds for macOS/Windows but not Linux). On macOS:
   ```bash
   xcode-select --install
   ```
@@ -31,15 +32,17 @@ participating, you agree to uphold it.
 ```bash
 git clone <your-fork-url> munder-difflin
 cd munder-difflin
-npm install        # postinstall rebuilds node-pty against Electron's ABI
+npm install        # postinstall rebuilds better-sqlite3 against Electron's ABI
 npm run dev        # live-reloading Electron build
 ```
 
 > [!IMPORTANT]
-> **The most common setup failure is the native `node-pty` rebuild.** The
-> `postinstall` script runs `electron-rebuild` so `node-pty` matches Electron's
-> ABI. If you see a "wrong ELF/Mach-O" or "NODE_MODULE_VERSION" error at launch,
-> re-run `npm install` (which re-triggers `postinstall`) after confirming your
+> **The most common setup failure is a native-module rebuild.** The
+> `postinstall` script runs `electron-rebuild --only better-sqlite3` so it
+> matches Electron's ABI (`node-pty` is Node-API and runs from its shipped
+> prebuilds — on Linux npm compiles it from source at install). If you see a
+> "wrong ELF/Mach-O" or "NODE_MODULE_VERSION" error at launch, re-run
+> `npm install` (which re-triggers `postinstall`) after confirming your
 > C/C++ toolchain is installed.
 
 ## Evidence is mandatory
--- a/README.md
+++ b/README.md
@@ -172,7 +172,8 @@ terminal/event plane, and [`DESIGN.md`](./DESIGN.md) for the visual system.
 
 - **macOS, Windows, or Linux**.
 - **Node.js 18+** and npm.
-- A **C/C++ toolchain** for `node-pty`'s native addon — on macOS, install Xcode Command Line Tools:
+- A **C/C++ toolchain** to build `better-sqlite3` for Electron (and `node-pty` on Linux, which has
+  no prebuilds there) — on macOS, install Xcode Command Line Tools:
   ```bash
   xcode-select --install
   ```
@@ -190,7 +191,7 @@ terminal/event plane, and [`DESIGN.md`](./DESIGN.md) for the visual system.
 ```bash
 git clone https://github.com/chaitanyagiri/munder-difflin.git
 cd munder-difflin
-npm install        # postinstall rebuilds node-pty against Electron's ABI
+npm install        # postinstall rebuilds better-sqlite3 against Electron's ABI
 npm run dev        # launches the Electron app with hot reload
 ```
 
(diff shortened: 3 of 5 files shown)


chaitanyagiri:
## TL;DR

Thanks @cyberfalcon-us. Narrowing a blanket force rebuild is a sensible instinct. `electron-rebuild -f` rebuilds every native module on every install whether it needs it or not, and on this project that is slow enough to be a real tax on contributors.

**What it does:** scopes the postinstall rebuild to one module.

```diff
-"postinstall": "electron-rebuild -f && node tools/ensure-pty-perms.cjs && node tools/patch-node-pty-conpty.cjs"
+"postinstall": "electron-rebuild -f --only better-sqlite3 && node tools/ensure-pty-perms.cjs && node tools/patch-node-pty-conpty.cjs"
```

Plus a comment correction in `.github/workflows/ci.yml` and matching notes in `README.md` and `CONTRIBUTING.md`.

**Verdict: needs work.** The change is small and plausible, but it needs one piece of evidence it does not currently have, and it collides with another open PR on the same line.

---

## What I did

Static review of the diff against `origin/main` @ `eb2818a8`. **Per the constraints I was working under I did not run `npm install` or a build on this branch**, which means the central claim of the PR is exactly the thing I could not verify. Flagging that plainly rather than implying otherwise.

## Finding 1: node-pty is the question this PR has to answer

`node-pty` is a native module too, and it is the one this entire app depends on for every terminal. After this change it is no longer force rebuilt.

That may well be fine. `electron-rebuild` without `-f` still rebuilds modules whose ABI does not match, and `tools/patch-node-pty-conpty.cjs` still runs. The `-f` flag forces a rebuild even when it looks unnecessary, and the whole point of your change is that it usually is unnecessary.

But "usually" is doing a lot of work there, and the failure mode is bad: a `node-pty` built against the wrong ABI does not fail at install, it fails when a user spawns their first agent. **What would settle it is one clean run**: fresh clone, `npm install`, `npm run dev`, spawn an agent, confirm the PTY works. On macOS and ideally on Windows, since the ConPTY path is the fussier one.

Please add that to the description. It is the single thing standing between this and a merge.

## Finding 2: direct collision with #223

PR #223 rewrites the same line differently:

```
"postinstall": "electron-rebuild && ..."
```

It drops `-f` entirely rather than scoping it. Your version keeps `-f` and narrows the target. Both are defensible and they are not compatible. Whoever merges second gets a conflict on one line, which is exactly the kind that gets resolved carelessly.

Worth a comment on one of the two PRs so the maintainer sees the pair together.

## Finding 3: the ci.yml comment, and why it is still worth care

Your `.github/workflows/ci.yml` change is one comment line, `node-pty` to `better-sqlite3`. It is accurate given the rest of your change, and it is cosmetic.

I am mentioning it only because `.github/` in this repo holds workflows that run with repository secrets, so every change there gets named explicitly in review regardless of size. Yours is fine. No action needed.

## Smaller note

Updating `README.md` and `CONTRIBUTING.md` alongside the change is good practice and it is not something most contributors bother with. Noted and appreciated.

1. One clean install and spawn run, described in the PR body, ideally on macOS and Windows.
2. A note pointing at #223 so the two postinstall rewrites are resolved deliberately.

chaitanyagiri:

---
**I checked the claim this rests on and it holds.** The change stops rebuilding everything and
rebuilds only `better-sqlite3`, and the docs justify it by saying `node-pty` is Node-API and ships
prebuilds. In `node_modules/node-pty` at version 1.1.0: `dependencies` is `["node-addon-api"]`,
which is the Node-API wrapper, and `prebuilds/` contains `darwin-arm64`, `darwin-x64`,
`win32-arm64` and `win32-x64` and no linux directory. Its install script is
`node scripts/prebuild.js || node-gyp rebuild`, so Linux falls through to a source build. That is
exactly what the new README and CONTRIBUTING text says, down to the Linux exception. The repo has
two native dependencies and `better-sqlite3` is the other one, so `--only better-sqlite3` is
complete for the tree as it stands.

The risk is what happens later, and this PR's own diff shows it. The Build job is
`continue-on-error: true`, described in the same hunk as `allowed to fail without blocking`. So
the one job that exercises a native rebuild cannot fail a PR. The day someone adds a native
dependency that is not Node-API, `--only better-sqlite3` will skip it silently and nothing in CI
will say so. A one line comment on the `postinstall` script naming the two modules and why the
scope is safe would give the next person somewhere to notice.

Two mechanical things before merge. The `package-lock.json` hunk moves the version from `0.4.4`
to `0.4.5`, and `origin/main` at `956bfb4c` is at `0.4.6`, so that hunk drags the version
backwards and needs dropping on rebase. And the API cannot compute mergeability for this PR right
now, it reports `mergeable: null`, which usually means it has not been refreshed against a moving
main.

Issue #253, opened 2026-08-23
11 test failures on Windows — path-separator mismatches (main is green on macOS/Linux)


BUGHUNTER-SACHIN:
### What happened?

Running the test suite on Windows fails 11 tests that pass cleanly on
macOS/Linux (confirmed against current `main`). All 11 failures share the
same root cause: tests hardcode Unix-style path strings (`/a/b/c`,
`unix:///tmp/...`) as expected values, but the code correctly produces
Windows-style paths (`C:\a\b\c`) on this platform. So this looks like a
test-authoring gap rather than a real bug in the underlying code — nothing
in the app itself appeared broken while using it.

Expected: all tests pass (or fail for a real reason) regardless of OS.

Note: tested with `npm install --ignore-scripts` (native module rebuild
skipped due to an unrelated Windows/ClangCL toolchain issue) — but none
of the 11 failures touch node-pty/native code, they're pure path-string
logic, so this shouldn't affect the report.

### Steps to reproduce

1. Clone the repo on Windows
2. `npm install --ignore-scripts`
3. `node --test test/*.test.cjs`
4. Observe 11 failures across `cli-install-ladder.test.cjs`,
   `codex-remote.test.cjs`, `expand-tilde.test.cjs`, and
   `transcript-project-dir.test.cjs` — full list and each expected/actual
   pair in the Logs section below.

### Screenshot or screen recording

<img width="1919" height="1039" alt="Image" src="https://github.com/user-attachments/assets/062d7438-265c-49f4-963b-087528fbd429" />
<img width="1919" height="1047" alt="Image" src="https://github.com/user-attachments/assets/b912461c-88f6-46a3-a2b7-085e32135ce4" />
<img width="1919" height="1050" alt="Image" src="https://github.com/user-attachments/assets/911c2727-dae4-4ca5-b154-95b48993f4c1" />

<img width="1918" height="1050" alt="Image" src="https://github.com/user-attachments/assets/41146fae-1425-40f6-91d0-c77dc0041467" />
<img width="1919" height="1048" alt="Image" src="https://github.com/user-attachments/assets/c440213e-b42b-404c-82e9-340d44581a7c" />
<img width="1919" height="1046" alt="Image" src="https://github.com/user-attachments/assets/230d0fdd-f727-4ace-ab8c-4cde33af8bdd" />
[...]

PR #270, opened 2026-08-24
fix: don't enqueue /compact for an undeliverable agent, latch at deli…

### LINKED ISSUE #255: Auto-compaction never reaches the god orchestrator: /compact is enqueued but the drain rejects god's 'blocked' status, then dedupe swallows every retry
## Summary

Scheduled auto-compaction (`contextTrigger.compact`) can never reach the **god orchestrator**. The `/compact` command is selected and enqueued correctly, but the queue drain refuses to deliver it, and the selector's dedupe then swallows every subsequent attempt. Sub-agents are unaffected.

No code excludes god from compaction. The exclusion is **emergent**: a status only god can hold meets a delivery gate that rejects it.

Line references are against `origin/main`.

## Mechanism

**1. `blocked` is a god-exclusive status.**

`src/renderer/src/hooks/usePtyParser.ts:154-157` — when a prompt is on screen, the parser sets `status: 'blocked'` only when `isGod`; every sub-agent gets `'waiting'` instead. The comment states the intent: *"Only the god agent talks to the human, so only it is truly 'blocked'."* `src/renderer/src/hooks/useHive.ts:525` repeats the same split (`self.isGod ? 'blocked' : 'waiting'`).

**2. The compaction selector does not exclude god.**

`useHive.ts` effect 6 resolves god's provider to `claude`, produces a valid `/compact`, passes the context-pressure gate, and calls `enqueueMessage(a.id, command)`. Nothing here is wrong — which is why reading this file alone finds no bug.

**3. The drain refuses `blocked`.**

`useHive.ts:921` gates delivery on `canDeliverToAgent(a.status, ...)`. In `src/renderer/src/hooks/queueDelivery.ts:48-56`:

```ts
if (status === 'idle') return true;
if (status !== 'looping') return false;
return ptyQuietMs !== null && ptyQuietMs >= quiesceMs;
```

`'blocked'` returns `false`. The `/compact` sits in the queue and is never typed.

**4. Why this hits god and nothing else.**

Waiting on the human *is* god's resting state. A sub-agent leaves `'waiting'` as soon as the orchestrator answers it; god can sit `'blocked'` for hours — which is precisely when the hourly trigger fires.

## Why it is permanent rather than merely delayed

**Dedupe poisoning.** The selector skips an agent when a command with the same leading verb is already queued:

```ts
if (queued.some((m) => m.text.trimStart().startsWith(verb))) continue;
```

Once one `/compact` is stuck undelivered in god's queue, every subsequent hourly compact is deduped away against it. There is no retry and no expiry — the queue silently swallows all of them.

**The latch is written at enqueue, not at delivery.** `useHive.ts:1113-1114`:

```ts
if (lastCompactUsed.current[a.id] === used) continue;
lastCompactUsed.current[a.id] = used;
```

This records "already compacted at N tokens" for a compaction that never happened. When god eventually returns to `idle`, an unchanged token count can latch the next attempt out too.

## Reproduction

1. Leave god at a prompt awaiting human input so the parser assigns `status: 'blocked'`.
2. Ensure god's context fill is above `contextTrigger.compact.minContextPct` (or `minContextPctLargeWindow` for windows ≥ 500k).
3. Wait for the trigger cadence (`everyMs`, default 1h).

Observed: god's context keeps growing; no `/compact` is ever typed into its terminal.

A live repro is awkward because it only manifests while god is `blocked` — if god happens to be `idle` when the timer fires, compaction succeeds and the run gives a false all-clear. The deterministic proof is a unit test:

```ts
expect(canDeliverToAgent('blocked', 60_000, 4_000)).toBe(false);
```

combined with the fact that `'blocked'` is assigned only under `isGod`.

## Please do not "fix" this by allowing delivery at `blocked`

Delivery appends Enter. Typing `/compact` into a prompt that is waiting on the human would **answer that prompt with the compaction command**. `queueDelivery.ts:42` already documents this hazard:

> *"every delivery with Enter, which would ANSWER it — the same reason the one-time TUI seed refuses to type at those two statuses."*

## Suggested fix

1. **Do not enqueue when the agent is currently undeliverable** — then no stuck item exists for the dedupe to collapse against. (Alternatively, evict or refresh a stale compact command rather than deduping against it.)
2. **Move the latch write from enqueue to successful delivery**, so `lastCompactUsed` reflects compactions that actually happened.

(1) and (2) together are likely the whole fix.

⚠️ This touches the shared message-delivery path used by inbox wakes, Slack routing and every dispatch. A regression here does not break compaction — it breaks the floor's mail. Tests first.

## Related: this is unobservable today

`grep -ic compact log.jsonl` returns **0**. The trigger fires, the selector decides, the drain drops — none of it is recorded. A silent skip is indistinguishable from "the timer has not fired yet", which is why this could be reported repeatedly without ever being confirmed or refuted.

Worth logging regardless of this fix: trigger fired (action + rule), per-agent decision (`queued` / `skipped-pressure` / `skipped-dedupe` / `skipped-latch` / `no-command`), and delivery outcome (`typed` / `undeliverable-status`).

BUGHUNTER-SACHIN:
Fixes #255. God can sit 'blocked' on a human prompt for hours, which is exactly when the hourly compaction trigger fires. The trigger enqueued /compact unconditionally, so it got stuck at the head of god's queue forever (blocked never becomes deliverable on its own), and the dedupe check then silently swallowed every subsequent hourly attempt against the stuck item. Two changes: (1) fire() now checks canDeliverToAgent before enqueueing, the same gate the drain itself uses immediately before typing, so nothing is queued in a state the drain would refuse to deliver. (2) the lastCompactUsed latch moves from enqueue-time to successful-delivery-time (via a new compactUsed field carried on the queued message), since writing it at enqueue recorded a compaction that may never have happened, permanently latching out future attempts at that token count.

## What & why

Fixes #255. God can sit 'blocked' on a human prompt for hours, which is exactly when the hourly compaction trigger fires. The trigger enqueued /compact unconditionally, so it got stuck at the head of god's queue forever, and the dedupe check then silently swallowed every subsequent hourly attempt against the stuck item. Two changes: (1) fire() now checks canDeliverToAgent before enqueueing, the same gate the drain itself uses immediately before typing. (2) the lastCompactUsed latch moves from enqueue-time to successful-delivery-time, since writing it at enqueue recorded a compaction that may never have happened.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before
Before

On `main` right now, `fire()` (the hourly compaction trigger) enqueues `/compact` for god unconditionally, with no check on whether god can currently receive it. God's `blocked` status (waiting on a human prompt) never becomes deliverable on its own, so once stuck the dedupe check silently swallows every subsequent hourly attempt forever. The latch is also written at enqueue time, recording "already compacted" for a compaction that never happened.

`useHive.ts`, lines 1113–1116 on `main`:

```typescript
          if (lastCompactUsed.current[a.id] === used) continue;
          lastCompactUsed.current[a.id] = used;
        }
        enqueueMessage(a.id, command);
```

No deliverability check anywhere before this enqueue.

### After
After 1

`fire()` now gates the enqueue on `canDeliverToAgent` — the exact same check the drain uses immediately before typing into a terminal — so nothing is queued in a state the drain would refuse to deliver. The `lastCompactUsed` latch write moves out of `fire()` entirely and into the successful-delivery callback in `flush()`, guarded on the new `compactUsed` field carried on the queued message.

After 2

After 3

After 4

After 5

After 6

After 7

After 8

## How I tested it

- OS: windows 11
- Steps:
  1. npm install --ignore-scripts
  2. npm run typecheck — clean
  3. node --test test/*.test.cjs — 536 pass, 11 fail (all pre-existing on main before this change — Windows path-separator issues tracked in #253, plus 2 Electron-native test files that don't run under --ignore-scripts; none touch queueDelivery.ts, useHive.ts, or store.ts)
  4. Confirmed the safety property from the issue, canDeliverToAgent('blocked', 60_000, 4_000) === false, is already covered by test/queue-delivery.test.cjs, and the fix reuses that exact function at the new call site rather than reimplementing the check
  5. Was NOT able to get a full Electron app run going on Windows to manually reproduce the god-blocked-during-hourly-trigger scenario end-to-end (node-pty/better-sqlite3 need a Clang toolchain not set up here) — flagged this explicitly for the maintainer below

## Discord (optional)

Discord:

## Checklist

### DIFF
--- a/src/renderer/src/hooks/useHive.ts
+++ b/src/renderer/src/hooks/useHive.ts
@@ -337,6 +337,15 @@ export function useHive(config: HarnessConfig | null): void {
   // listPtys on exactly the cadence the drain needs, and one reading keeps the
   // two loops from disagreeing about whether an agent is quiet.
   const ptyLastOutput = useRef<Record<string, number>>({});
+    /** How long this terminal has been silent, or null when we have no reading
+   *  (never polled, PTY gone, or it has emitted nothing at all). canDeliverToAgent
+   *  fails closed on null — unmeasured silence is not evidence of silence.
+   *  Hook-level (not effect-local) because both the drain effect and the
+   *  context-trigger effect gate delivery through the same check. */
+  const ptyQuietMs = (ptyId: string, now: number): number | null => {
+    const last = ptyLastOutput.current[ptyId];
+    return typeof last === 'number' && last > 0 ? now - last : null;
+  };
   // #5C/#7C.4 — latest circuit-breaker level per agent. When 'constrained'/
   // 'stopped' the avatar is pinned to 'looping' and hook events must NOT flip it
   // back to 'working' (the flicker the spec calls out); only a genuine Stop clears it.
@@ -772,13 +781,6 @@ export function useHive(config: HarnessConfig | null): void {
     const inFlight = new Set<string>();
     const sendFailures: Record<string, number> = {};
 
-    /** How long this terminal has been silent, or null when we have no reading
-     *  (never polled, PTY gone, or it has emitted nothing at all). canDeliverToAgent
-     *  fails closed on null — unmeasured silence is not evidence of silence. */
-    const ptyQuietMs = (ptyId: string, now: number): number | null => {
-      const last = ptyLastOutput.current[ptyId];
-      return typeof last === 'number' && last > 0 ? now - last : null;
-    };
 
     // Send the front of `srcId`'s queue into `target`'s pty (verbatim or wrapped),
     // gated on the target being idle, free of interactive menus, and off
@@ -920,8 +922,13 @@ export function useHive(config: HarnessConfig | null): void {
         // the one inside dispatch would have changed nothing.
         if (!a.ptyId || !canDeliverToAgent(a.status, ptyQuietMs(a.ptyId, now), QUIESCE_IDLE_MS)) continue;
         if (!messageQueues[a.id]?.length) continue;
-        void dispatch(a.id, a).then(({ sent, message }) => {
+                void dispatch(a.id, a).then(({ sent, message }) => {
           if (sent && message?.slack) void ensureSlackCard(message);
+          // Write the compact latch only once delivery genuinely happened — see
+          // the comment in fire() above.
+          if (sent && message?.compactUsed !== undefined) {
+            lastCompactUsed.current[a.id] = message.compactUsed;
+          }
         });
       }
     };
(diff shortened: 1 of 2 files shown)


chaitanyagiri:
Shipped in v0.4.6, and it closes #255.

Your commit `ee806756` is in the release. Both halves of it: `fire()` now checks `canDeliverToAgent` before enqueueing, and the `lastCompactUsed` latch moved from enqueue time to successful delivery time. The second one was the subtle half, since writing the latch at enqueue recorded a compaction that may never have happened and permanently latched out later attempts.

Thank you, that was a good write up as well as a good fix.

PR #273, opened 2026-08-24
perf(floor): make the floor's motion frame-rate independent, then cap it at 20 fps

L422Y:
## What & why

Pixi's ticker follows `requestAnimationFrame` and nothing capped it. On a ProMotion
Mac that means the office scene ran at 120 fps — I measured 120.5 on a 16-inch
MacBook Pro. Every frame runs `onTick` over every character and all six floor
subsystems, then redraws the whole scene onto a `resolution: 2` canvas.

That is the app's largest continuous cost, and it is spent on ambient decoration: an
idle office that nobody is tracking, animating as fast as the display will accept for
as long as the app is open, whether or not anyone is looking at it. On a laptop a
continuous cost is battery, and this one is heavy enough to be felt.

Capping it turned out to be the easy part. It exposed two places where motion was a
function of the **frame rate** rather than of elapsed time, and both had to be fixed
first — which is why this is one change and not three. Capping without them ships
visible regressions:

**1. Characters limped.** `Character.updateWalk` advanced at most one path waypoint per
frame, and reaching a tile threw away the rest of that frame's movement. Characters
cross a tile three times a second, so the stall rate tracks the frame rate:

| cap | stalled frames | actual walking speed |
|---|---|---|
| 120 fps | 3% | ~100% |
| 60 fps | 5% | 100% |
| 30 fps | 8% | 92% |
| 20 fps | 13% | 83% |

At 120 fps that is one frame in forty and nobody sees it. Capped, it is one in eight,
held on screen for 50ms, and it reads as the character speeding up and slowing down.
It also means characters were walking at 83% of the speed the code asks for. The walk
now spends its whole movement budget each frame, crossing as many waypoints as that
covers and landing exactly on each one.

**2. The camera's easing depended on the display.** `Camera.update` moved a fixed
fraction of the remaining distance *per frame*, ignoring `dt`. That is already a bug on
`main` — the same pan finishes twice as fast on a 120 Hz panel as on a 60 Hz one — and
capped, a window resize would have taken five seconds to settle its fit instead of 1.6.
The fraction is now the equivalent per-second decay.

**3. The walk cycle ignored the cap entirely.** `CharacterSprite` left Pixi's
`AnimatedSprite` on its default `autoUpdate`, which subscribes it to `Ticker.shared` —
a second `requestAnimationFrame` loop that no cap on the app's ticker touches, and that
kept animating every character while the floor was paused behind a fullscreen terminal
or the IDE. It is driven from the scene's tick now. Being straight about this one: it is
a consistency fix, not a saving. Measured against itself, interleaved three times, it is
~0.3% of a core — inside the noise. It earns its place because it makes the claim in this
PR's title true, and because the existing pause optimisation was silently leaking through it.

The first two are pure functions with tests pinning them equal across frame rates,
including one that reproduces the old walk rule losing ground so nobody reinstates it.
Distance travelled per second is now identical at 120, 60, 30, 20, 15, 12 and 10 fps to
four decimal places.

With motion no longer tied to the frame rate, the cap is free to be low. The floor is
ambient: nobody reacts to it, and the camera never pans (`fitToScreen` sets its target
once, there is no drag or wheel zoom, and `focusOn()` has no callers).

What each cap costs

20 and 15 measure the same, which is what picks the default: what is left below 20 is
React, the terminals and compositing, none of which is this ticker, so going lower buys
nothing and only costs smoothness. The cap is a named constant carrying the
measurements, so moving it is a one-line decision with the evidence next to it.

This does not touch the existing pause logic, which already stops the ticker entirely
behind a fullscreen terminal, behind the IDE, or on a hidden window. This is about what
the floor costs while you are looking at it.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI
- [x] Performance

## Evidence

Measured on a live floor with agents working, not a synthetic scene. `%CPU` here is a
CPU-time delta over a 22-second window, not `ps`'s lifetime average, which is useless
for this. Conditions were interleaved and walked up and back down in a single session
(uncapped → 60 → 30 → 20 → 15 → 15 → 20 → 30 → 60 → uncapped) so drift in the floor's own
workload cancels instead of being credited to the change. Same window, same agents,
same machine, minutes apart. Each dot is one sample window.

### Before
Uncapped: renderer 26.7%, GPU 16.9%

### After
Capped at 20 fps: renderer 12.7%, GPU 5.8%

52% off the renderer process and 66% off the GPU process, continuously, for as long as
the app is open.

To be straight about the trade rather than sell it: **120 fps is smoother, and the floor
at 20 does not look identical.** What a lower cap does not change is *speed* — after the
fixes above, everything covers the same ground per second and simply does it in fewer
steps. Smoothness is the thing being traded, deliberately, because a novelty animation
that never stops is not worth the power it was drawing.

## How I tested it

- OS: macOS (Darwin 25.3), Node 24, 16-inch MacBook Pro (120 Hz Liquid Retina XDR).

- Steps:
  - `npm run typecheck`, `npm run test:focused` (564 pass, 12 of them new), `npm run build`.
  - Measured the display's real rAF rate rather than assuming it: 120.5 fps.
  - Ran the sweep above against a running dev build, toggling only `FLOOR_MAX_FPS`.
  - Watched the floor at each cap for walking, coffee runs, errands, thought bubbles and
    message envelopes. The limp at low frame rates is what led to the walk fix; with it
    in, 20 fps is where I stopped being able to tell, and it measures the same as 15, so
    there was no reason to go lower.
  - Read Pixi's `Ticker.update` and simulated its `maxFPS` gate against a 120 Hz frame
    stream. It emits a periodic long frame at every capped rate (at 20 fps: a 58 ms frame
    among 50 ms ones). Harmless, because everything on the floor is driven by
    `deltaMS` — but it is the reason the walk stall had to go, since a stall is not.

Not done: verified on a 60 Hz display, where the cap should be close to a no-op, or on
Windows and Linux.

## Discord (optional)

Discord: l422y

## Checklist

- [x] This PR is **one change**. The two motion fixes are not unrelated work bundled in —
      capping the frame rate without them ships a visible regression in each.
      spacing, or fonts. (No UI added.)
      `ATTRIBUTION.md`. (No art.)

### DIFF
--- a/src/renderer/src/scene/office/Camera.ts
+++ b/src/renderer/src/scene/office/Camera.ts
@@ -1,11 +1,19 @@
 import { Container } from 'pixi.js';
+import { timeScaledLerp } from './easing';
 
 // Simplified port of shahar061/the-office (office/engine/camera.ts).
 // Phase targeting is dropped (we have no project phases); kept: fit-to-screen,
 // map-edge clamping, smooth lerp, a manual focus(), and nudgeToward() used to
 // pan toward the selected agent.
 
+// Authored as a fraction of the remaining distance per frame, at REF_HZ.
 const LERP_SPEED = 0.08;
+// The rate that fraction was tuned against. Applied raw, a per-frame constant
+// makes the camera's easing depend on the display: it converges twice as fast on
+// a 120 Hz ProMotion panel as on a 60 Hz one, and would crawl the moment anything
+// capped the frame rate. Converting it to the equivalent per-second decay (below)
+// gives the same wall-clock easing at any rate.
+const REF_HZ = 60;
 
 export class Camera {
   private container: Container;
@@ -73,9 +81,14 @@ export class Camera {
   }
 
   update(dt: number): void {
-    this.currentX += (this.targetX - this.currentX) * LERP_SPEED;
-    this.currentY += (this.targetY - this.currentY) * LERP_SPEED;
-    this.currentZoom += (this.targetZoom - this.currentZoom) * LERP_SPEED;
+    // Same easing curve, expressed against elapsed time rather than frame count.
+    // dt is clamped upstream by Pixi's minFPS, so a floor that was paused for an
+    // hour resumes with k near 1 (the camera snaps to its target) rather than
+    // jumping past it.
+    const k = timeScaledLerp(LERP_SPEED, dt, REF_HZ);
+    this.currentX += (this.targetX - this.currentX) * k;
+    this.currentY += (this.targetY - this.currentY) * k;
+    this.currentZoom += (this.targetZoom - this.currentZoom) * k;
 
     if (this.nudgeDuration > 0) {
       this.nudgeElapsed += dt;
--- a/src/renderer/src/scene/office/Character.ts
+++ b/src/renderer/src/scene/office/Character.ts
@@ -1,6 +1,6 @@
 import { Container, Graphics, Texture } from 'pixi.js';
 import { CharacterSprite, type Direction, type AnimState } from './CharacterSprite';
-import { findPath } from './pathfinding';
+import { advanceAlongPath, findPath } from './pathfinding';
 import type { TiledMapRenderer } from './TiledMapRenderer';
 import { ThoughtBubble } from './ThoughtBubble';
 
@@ -553,6 +553,12 @@ export class Character {
     this.thoughtBubble.update(dt);
     if (!this.isVisible) return;
 
+    // The walk/idle cycle now rides the floor's clock rather than Ticker.shared
+    // (see CharacterSprite), so it stops with the floor and never animates faster
+    // than the scene is drawn. An invisible character returns above and freezes,
+    // which is what we want.
+    this.sprite.advance(dt);
+
     // Working agents stay seated; between tasks they wander the office.
     // A cheer, a watering or a cigar holds roaming so the effect plays in place.
     const heldByFx = this.cheerT >= 0 || this.waterT >= 0 || this.smokeT >= 0;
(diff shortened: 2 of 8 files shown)


chaitanyagiri:

---
The measurement table is the right way to make this argument. Interleaving the caps up and back
down in one session so the agents' own varying workload cancels rather than being credited to the
cap is the detail most benchmarks skip, and it is what makes 20 and 15 measuring the same number
mean something. `Be honest about the trade: 120 fps IS smoother, and the floor at 20 does not look
identical` is the sentence that makes the rest of it credible.

The walk fix is the more interesting half. Reaching a tile used to cost the rest of that frame's
movement, which is one frame in forty at 120 fps and one in seven at 20, so capping the rate
without `advanceAlongPath` would have made the characters visibly limp. Doing the two together,
and saying why in the comment, is the correct order.

**One thing I would pin down.**

```
private static readonly beat = { deltaTime: 0 } as unknown as Ticker;
```

One mutable object shared by every sprite, and a cast that stops the compiler from mentioning the
rest of `Ticker`. It is safe under a synchronous tick and the comment justifies it, but the
justification is a claim about Pixi's internals: `AnimatedSprite.update()` reads exactly one field
today. If a future Pixi reads `elapsedMS` or `speed`, every sprite on the floor stops animating,
the cast guarantees no compile error, and the symptom is "the office went still" with nothing to
grep for. #338 in this same batch writes `verified on Codex 0.149.1` for the same kind of claim.
Putting the Pixi version in this comment costs one number and tells the next upgrader what to
recheck.

`autoUpdate = false` is the right call regardless. A second requestAnimationFrame loop running at
the panel's full rate while the floor is hidden behind a terminal is pure waste.

PR #275, opened 2026-08-24
feat(settings): let people choose the office floor's frame rate

L422Y:
## What & why

> **Stacked on #273.** That PR caps the floor's frame rate and is the first commit here;
> this PR is the second. Review #273 first — once it merges, this diff is just the
> setting.

#273 lands the frame-rate ceiling as a constant. A constant is the wrong shape for it.
Smoothness is a legitimate thing to want, the cost of wanting it depends on hardware
nobody in this repo can see, and on a desktop that is never on battery the power may
simply not matter. Picking one rate for everyone decides that on their behalf.

Settings → General now offers **15 / 20 / 30 / 60 / uncapped**, each labelled with what
it costs rather than a bare number, so the choice can be made without reading the
source. Applying is instant and reversible: the store mirror drives the live Pixi
ticker, so the floor repaints at the new rate on click. Nothing here is destructive and
there is nothing to confirm.

**The default stays 20**, because it measures the same as 15 on the machine this was
profiled on — the cheapest rate that is not needlessly choppy. The full table lives in
`src/shared/floorFps.ts` next to the choices it justifies, so the numbers and the
options cannot drift apart.

**On validation.** This value reaches Pixi's ticker off disk, so it is not trusted.
Anything unrecognised falls back to the default: a hand-edited rate below Pixi's own
`minFPS` makes the ticker clamp elapsed time, which silently slows every character
down rather than failing, and a hand-edited 240 would reinstate exactly the cost the
setting exists to cap. `resolveFloorFps` is a pure function with tests covering the
offered rates, the fallbacks, and that uncapped stays `0` — the value Pixi reads as
no cap.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

## Evidence

The same Settings → General section, same window, same dark theme, rendered from the
real component.

### Before
Settings General: the section ends, with no frame-rate control

### After
Settings General: a Floor frame rate control offering 15, 20, 30, 60 and uncapped, with 20 selected

## How I tested it

- OS: macOS (Darwin 25.3), Node 24, 16-inch MacBook Pro (120 Hz Liquid Retina XDR).

- Steps:
  - `npm run typecheck`, `npm run test:focused` (569 pass, 5 of them new), `npm run build`.
  - Restarted the app and confirmed the new default reaches disk: `config.json` comes
    back with `"floorMaxFps": 20`.
  - Confirmed the value actually reaches the ticker, by toggling **only** the store
    mirror on a running dev build and measuring the floor (% of one core, renderer /
    GPU, 22-second windows, alternated twice):

    | store value | renderer | GPU |
    |---|---|---|
    | uncapped | 25.8, 25.6 | 15.9, 11.8 |
    | 20 fps | 14.7, 10.3 | 7.0, 4.9 |

    No restart and no scene rebuild between those — the ticker picks the new ceiling up
    from the store effect, which is the behaviour the picker relies on to apply on click.
  - For the screenshots: mounted the real `FloorFpsPicker` in a throwaway page outside
    the repo against a stubbed `window.cth`, with the app's own `tokens.css` and
    `global.css`. It is the real component and the real stylesheet, not the whole
    Electron window, so the Settings chrome around it is not in frame.

  - Clicked the control in the running app. The floor changes rate immediately, and the
    choice is on disk by the time the click lands: `config.json` came back with the
    newly-picked rate, written seconds earlier. Read, apply and write are all confirmed
    against a real build.

Not done: verified on a 60 Hz display, where every rate above 60 collapses to the same
thing, or on Windows and Linux.

## Discord (optional)

Discord:

## Checklist

- [x] This PR is **one change** — the setting. The cap itself is #273.
- [x] Any new UI derives from `DESIGN.md` / `tokens.ts` — the picker uses the paper,
      ink and lilac tokens and the display/mono faces, no ad-hoc values.
      `ATTRIBUTION.md`. (No art.)

### DIFF
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -316,6 +316,14 @@ export interface HarnessConfig {
    *  `tvShowOffices` is on; otherwise the office theme is used. Unbuilt show
    *  themes fall back to 'office' in the loader. */
   officeTheme?: 'office' | 'friends' | 'brooklyn99' | 'siliconvalley' | 'got' | 'hogwarts';
+  /** Frame-rate ceiling for the pixel office scene, in fps. 0 = uncapped, which
+   *  means the scene follows the display and costs twice as much on a 120 Hz
+   *  panel as on a 60 Hz one. The floor is ambient decoration that animates for as
+   *  long as the app is open, so the default is deliberately low; the option
+   *  exists because smoothness is a real thing to want, and on a desktop the power
+   *  it costs may not matter. Unrecognised values fall back to the default —
+   *  see FLOOR_FPS_CHOICES in src/shared/floorFps.ts for the supported set. */
+  floorMaxFps?: number;
   /** Per-CLI-provider local/self-hosted base URL (Ollama/LM Studio/vLLM, …) for the
    *  OpenCode/Crush/pi/qwen engines; applied at spawn (config-injection or proxy
    *  upstream). API KEYS are NOT stored here — they live write-only in the secret
@@ -448,6 +456,7 @@ const DEFAULTS: HarnessConfig = {
   multiWindow: true,
   tvShowOffices: false,
   officeTheme: 'office',
+  floorMaxFps: 20,
   slackEnabled: false,
   slackSigningSecret: undefined,
   slackBotToken: undefined,
--- a/src/preload/index.ts
+++ b/src/preload/index.ts
@@ -321,6 +321,8 @@ export interface HarnessConfig {
   tvShowOffices?: boolean;
   /** Active office map/cast theme (honored only when tvShowOffices is on). */
   officeTheme?: 'office' | 'friends' | 'brooklyn99' | 'siliconvalley' | 'got' | 'hogwarts';
+  /** Frame-rate ceiling for the office scene, in fps (0 = uncapped). */
+  floorMaxFps?: number;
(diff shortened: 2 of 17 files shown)


chaitanyagiri:

---
**This and #273 are the same change from two ends and one of them makes the other's copy true.**
#273 makes the floor's motion frame-rate independent and hardcodes a 20 fps cap with the
measurement behind it. This one makes the cap a setting with the same 20 default. Both touch
`OfficeFloor.tsx`.

That matters beyond the merge order, because this panel says:

> It does not change how fast anything moves — the floor's motion runs on elapsed time, so a
> character covers the same ground per second at every setting and simply does it in fewer steps.

**That sentence is only true after #273.** Today the camera lerp is a per-frame constant and
`updateWalk` spends the rest of a frame's movement every time a character reaches a tile, so
dropping to 20 fps genuinely does slow the floor down. Shipping this alone would put a claim in
the settings panel that the app contradicts. Land them together, or land #273 first.

The optimistic-then-revert with a note on failure is the right pattern and matches what #320 is
doing in the same batch.

Small: `var(--cth-ink-${active ? '700' : '100'})` builds a token name by interpolation, so
grepping for `--cth-ink-700` will not find this use. Same for the inline `#ece2f5` fallback on
`--cth-lilac-light`, which is the only hardcoded colour in the file. Neither breaks anything.

PR #282, opened 2026-08-24
fix: god's renamed identity no longer reverts to "Michael" on respawn

### LINKED ISSUE #279: God's custom name reverts to "Michael" on every restart/respawn (separate from #188)
### Problem / motivation

Renaming god (the orchestrator) works and persists to `roster.json`, but reverts to the hardcoded default `"Michael"` on every app restart / god respawn — the Command Center, remote-control session naming, and the agent's own `name` field all snap back, even though the custom name was saved correctly beforehand.

This is distinct from #188 (closed) — that fix covers renaming ordinary hired agents via the Edit Agent panel, which does persist correctly across restarts (confirmed working here for a hired agent). God appears to go through a separate spawn path that isn't wired to the same persisted-name lookup.

### Where it lives

`out/renderer/assets/index-<hash>.js` (v0.4.5), inside the god-spawn effect (guarded by `godSpawning.current`). Two spots hardcode the literal string `"Michael"` unconditionally when (re)constructing god's agent object on every spawn, instead of reading the current name from the registry/config:

```js
const res = await window.cth.spawnPty({
  id: GOD_PTY,
  ...
  hive: { id: GOD_ID, name: "Michael", provider: godProvider, cwd: config2.harnessHome, isGod: true, role: "orchestrator (god)" }
});
...
const god = {
  id: GOD_ID,
  name: "Michael",
  character: "michael",
  ...
};
useStore.getState().addAgent(god);
...
const remoteCommand = remoteControlCommandForProvider(godProvider, "Michael");
```

For comparison, three other places in the same bundle correctly read the live/persisted name instead of hardcoding it:

```js
const godName = godId ? str$1(obj$1(obj$1(reg.agents)[godId]).name) || godId : null;
...
const godName = useStore.getState().agents.find((a2) => a2.isGod)?.name ?? "the orchestrator";
```

### Repro

1. Rename god via the Edit Agent panel (e.g. `Michael` → `Savvas`). `roster.json`'s `name` field updates correctly.
2. Restart the app (or otherwise trigger a god respawn).
3. Command Center and the remote-control session name both show `Michael` again, even though `roster.json` still has the custom name.

### Proposed fix

In the god-spawn effect, read the existing name for `GOD_ID` from the persisted registry/config (the same source the three correct call sites already use) before falling back to the literal `"Michael"`, and pass that resolved name into both the `hive:` payload and the `remoteControlCommandForProvider` call, instead of the hardcoded string.

### Version

Munder Difflin 0.4.5, macOS, packaged build.

savvaskoualis:
Closes #279.

## What & why

The same root cause, ~30 times over: a value that should read the live/persisted
`agent.name` was a fixed string `'Michael'` instead, scattered across UI text,
dropdown labels, tooltips, a desktop notification, and an internal system-prompt
template.

**Round 1** (original scope): two call sites reverted or ignored a god rename —
`useHive.ts`'s god-spawn effect (rebuilt god's agent object with `'Michael'`
hardcoded on every restart, so a rename reverted) and `CommandCenterPanel.tsx`'s
header caption (`Michael runs the floor`, wrong the instant you rename, no
restart needed).

**Round 2** (found auditing for the same shape elsewhere): ~30 more call sites
across `CommandCenterPanel` (dispatch section title/dropdown/placeholder — see
before/after below), `SchedulesSection`, `TriggerHistoryTab`, `SettingsModal`,
`WorkersTab`, `TasksKanban`, `AgentHoldButton`, the two "clocking in" boot
screens, the voice device picker, the completion toast, a desktop notification
title, a voice-action board post + ping subject, and the PREP ASSISTANT
system-prompt template.

`realtime/tools.ts`'s fleet-status tool already did this correctly
(`reg.agents[godId]?.name || godId`) — none of the buggy call sites consulted
the persisted/live name at all.

**Left deliberately alone** (same shape, not the same bug): the Office
character-roster entry (`"Michael"` as a selectable avatar name, same as
`"Jim"`/`"Pam"`), ambient office flavor/easter-egg dialogue, past-release
changelog copy, onboarding-only text shown before any hire/rename is possible,
the "Realtime Michael" feature's own branded name, and internal protocol
identifiers (`from:"michael-voice"`, the `useRealtimeMichael` hook name) — none
of these claim to reflect the live god's current display name.

## The fix

Round 1: extracted name-resolution into `src/shared/godIdentity.ts::resolveGodName()`.
The spawn effect now calls `window.cth.hiveRegistry()` before building the spawn
payload; the header reads `agent.name` directly.

Round 2: renderer call sites read the live agent from the store
(`agents.find(a => a.isGod)?.name`) or the already-in-scope `agent`/`a` prop,
matching whichever pattern the surrounding component already used. The two
pre-spawn boot screens share a new `useResolvedGodName()` hook that reads the
persisted registry name directly, same as `useHive.ts`'s own spawn effect.

Main-process call sites resolve the name once via `hive.registry()` /
`deps.hiveRegistry()`, through the same `resolveGodName()` helper.
`hive.ts`'s `injectedPrompt()` needed care: its own doc comment states a
**prompt-cache invariant** — the injected system-prompt prefix must stay
volatile-free, no registry state re-read per turn, or it defeats Anthropic's
prompt cache. The fix respects it: the PREP ASSISTANT's system prompt reads
god's name ONCE, at that assistant's own spawn time — exactly like its own
name/id/dir/root already do — not a live re-read within an already-running
session.

### Before
before: DISPATCH section still hardcodes "Michael" after a rename

### After
after: same repro, DISPATCH section correctly reads "Savvas"

Both pairs captured the same way: a fully isolated scratch hive (no real
project data), via Playwright's `_electron` window-scoped screenshot — never
OS-level screen capture, so there's no possibility of unrelated window content
in frame. Repro: rename god via the pencil icon in the sidebar, open the
monitor tab. On `main` the dispatch section (title, dropdown default,
placeholder) all still say "Michael"; on this branch they correctly read the
renamed name.

The original round-1 evidence (restart-persistence + header caption) is still
at `god-name-persistence/before-reverted.png` / `after-persisted.png` in the
same assets branch, unchanged.

## Testing

`npm run typecheck` / `npm run build` clean. `npm run test:focused`: 551 tests,
550 pass (2 for `resolveGodName()`, both green). The 1 failure
(`test/update-download-asset.test.cjs`) is a pre-existing sandbox artifact —
`npm install --ignore-scripts` skips Electron's own binary install too,
unrelated to this change and reproduces on `main` before this diff.

Tested on macOS (Apple Silicon), Node 26.7.


chaitanyagiri:
Shipped in v0.4.6, and it closes #279.

All three of your commits are in: `46737e52` for the persisted identity surviving respawn, `08489c1c` for the Command Center header, and `c461fe1a` for every remaining hardcoded "Michael" reading the live name. The third one is the reason the fix actually holds, since the first two alone would have left the name reverting in half the surfaces.

Thank you.

PR #283, opened 2026-08-24
feat(office): add dynamic avatar bubble text scaling, typography and …

surajnai567:
## What & why

Character thought and speech bubbles above avatars on the office floor were previously fixed to a single small font size that rendered blurry on high-DPI displays and was difficult to read at normal zoom levels. This PR introduces dynamic text scaling controls, high-contrast crisp typography (`Inter` / `JetBrains Mono`), selectable clarity rendering modes (Crisp HD, Standard, Retro Pixel), and user controls exposed both via a floating floor HUD and in Settings → General with persistent local storage.

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before

- Avatar thought and speech bubbles were rendered at a fixed small size with low resolution canvas text scaling.
- Text appeared blurry on high-resolution screens and lacked any text scale or clarity adjustment options.

Screenshot 2026-08-25 011955

### After

- Clear, crisp typography with supersampled anti-aliased font rendering.
- Interactive floating `[Aa TEXT]` widget on the office floor allowing instant text scaling (`[-]`, `[+]`, presets `80%` to `150%`) and clarity switching (`Crisp HD`, `Standard`, `Pixel Retro`).
- Dedicated **"Avatar Speech & View Clarity"** section in **Settings → General** with a live preview bubble.
- Settings persist across restarts via `localStorage`.

Screenshot 2026-08-25 011843

## How I tested it

- OS: Windows 11
- Steps:
  1. Ran `npm run dev` and launched the app.
  2. Spawned agents and verified thought/speech bubble legibility at default zoom.
  3. Opened the bottom-right floating `[Aa TEXT]` HUD on the office floor, adjusted scale from 80% to 150%, and verified character bubbles dynamically updated immediately.
  4. Switched clarity modes (`Crisp HD`, `Standard`, `Pixel Retro`) and confirmed font and rendering style changed in real-time.
  5. Opened **Settings → General**, tested scale presets and clarity options with the interactive preview, and confirmed preferences persisted after restarting.

## Checklist

### DIFF
--- a/src/renderer/src/components/SettingsModal.tsx
+++ b/src/renderer/src/components/SettingsModal.tsx
@@ -177,6 +177,11 @@ export function SettingsModal({ config, onClose, initialSection }: SettingsModal
     (config as HarnessConfig & { notifications?: boolean }).notifications === true
   );
 
+  const bubbleTextScale = useStore((s) => s.bubbleTextScale);
+  const setBubbleTextScale = useStore((s) => s.setBubbleTextScale);
+  const bubbleClarity = useStore((s) => s.bubbleClarity);
+  const setBubbleClarity = useStore((s) => s.setBubbleClarity);
+
   const toggleNotifications = async () => {
     const next = !notifications;
     setNotifications(next); // optimistic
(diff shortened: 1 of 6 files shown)


chaitanyagiri:

---
The accessibility need is real, and offering 80 to 160 percent with a live preview is the right
shape for it.

**The preview is built from different values than the thing it previews, which is the one change I
would make before merge.** The chip hardcodes `'"Press Start 2P", monospace'`, `'"Inter",
sans-serif'` and `color: '#1a1320'`, while the floor renders through Pixi with its own text style.
Nothing keeps the two in step, so the preview can be accurate today and quietly wrong after any
change to `ThoughtBubble` or `ToolBubble`. A preview that cannot be wrong is worth more than a
preview that merely looks right now, and deriving both from one shared style object would give you
that.

**These two settings persist to `localStorage`, not to `config.json`.** That is consistent with
`cth.sidebarWidth` and `cth.sidebarTab` right beside them, so it is a precedent and not an
invention. It is worth deciding on purpose though, because every other control in this Settings
panel writes through `updateConfig`: a reset of settings will not reset these, they are not in
anything a user could back up (#293 is open on exactly that), and a per-project value becomes
impossible if #383 lands and instances share a renderer origin.

On the copy: `Crisp HD`, `Pixel Retro` and `High supersampling crisp text vs. classic pixel font
rendering` do not tell a user what changes or what it costs. #275 in this same batch is the
comparison worth reading, it says plainly what the setting does and does not do. And
`Avatar Speech & View Clarity` is two features in one heading.

This conflicts as of today, `mergeable: false`.

PR #286, opened 2026-08-24
fix(config): announce every config write so Settings stops showing stale values

### LINKED ISSUE #263: Settings do not persist: the "Michael can spin up agents on his own" toggle reads off while agents are still being spun up
Turning off "let Michael spin up agents on his own" does not hold. The agent spun one up anyway, and the setting still read as off afterwards.

In the reporter's words:

> Settings don't seem to save/persist? Or some of them don't. I couldn't let "Michael" spin up agents on "his" own and then randomly he did it even though the setting was still off.

### What was observed
- The setting was turned off.
- An agent was spun up by Michael regardless.
- The toggle still displayed as off after it happened.

### Not established
Whether the write never persisted, whether it persisted and is not read on that path, or whether it is a display problem only. The reporter said "or some of them don't", so it is also not established which settings are affected beyond this one.

---

Reported on Hacker News by **joshstrange**, who ran the app for a couple of hours on 2026-08-22 and wrote up everything he hit.

Source: https://news.ycombinator.com/item?id=49400442 and https://news.ycombinator.com/item?id=49400749

Filed by a maintainer so it does not stay buried in a comment thread. The reporter may not be on GitHub, so please do not close this waiting on him to answer. App version was not stated in the original report.

HundredBillion:
## Summary

**Problem:** Settings would save your change to disk correctly, and then show you the old value the next time you opened it. Most of the time that was a cosmetic annoyance that cleared when you restarted the app. But two of the circuit-breaker fields went further and *deleted* the value you had saved — the field came back blank, a blank means "no value", so the next save wiped it. And the "Who can add agents" toggle looked like it flat-out refused to change: because it flips relative to what is shown on screen, the first click after reopening Settings wrote back the value that was already saved, doing nothing visible.

**Solution:** The app now tells the interface whenever a setting is written, so what you see always matches what is stored. Change a setting, close Settings, open it again — it shows what you set. One click on a toggle does what one click should do. Nothing gets erased.

## TLDR for developers

The renderer loads the config once (`App.tsx:90`) and passes it down as a prop. Nothing informed it when a write landed, so any view seeded from that prop rendered the start-up snapshot until relaunch — `SettingsModal.tsx:444-446` already documented this as known.

- **`src/main/config.ts`** — added `onConfigWritten(listener)`, notified from `persistConfig`. Placement is the point: `writeConfig` is one of **23** `writeConfig` callers in the main process, and Slack, freeflow and notifications each persist by their own route. `persistConfig` is the one place they all pass through, which is the difference between fixing the fields we happened to enumerate and fixing all of them. Listener errors are swallowed — a listener is a renderer send that can fail while a window closes, and the config is already on disk by then, so a throwing subscriber must not fail the write or starve the subscribers after it.
- **`resetConfig`** now routes through `persistConfig` instead of writing the file itself, so it announces like any other write. Same bytes on disk.
- **`src/main/index.ts`** — forwards each write to **every live window** as `config:changed`. Floor windows load the same renderer and hold their own copy of the config, and `mainWindow` is reassigned on focus, so sending to the primary alone would have left other floors stale and made the recipient depend on focus order.
- **`src/preload/index.ts`** — `onConfigChanged`, mirroring the existing `onCloseRequested` / `onHireImport` subscription shape.
- **`src/renderer/src/App.tsx`** — one `useEffect` keeping `config` current.

Subscribers are handed the config in the shape `readConfig` returns, not the object passed to `persistConfig`. The renderer feeds one `config` state from two sources — `config:get` and this notification — and `config:get` is deep-filled and home-normalised. A patch touching one nested key persists a half-filled sub-object, so without this a component could read `contextTrigger.compact.minContextPct` as a number from a read and `undefined` from a notification moments later; `config.ts:492-505` already documents that hazard. The migration is deliberately not run on that path: it persists, and it has already been applied to the config the patch was built on.

Why this shape rather than patching the re-seed effect: `SettingsModal` seeds **28** fields from the prop and its effect re-reads only **14**, so 14 displayed the stale value. Adding the missing 14 setters would have fixed today's list and left the same two-list drift that produced the bug — every new field would still need registering twice. Making the prop authoritative removes the class.

`SettingsModal`'s re-seed effect is deliberately left alone. With the prop current it is redundant, but deleting it is a separate change and does not belong in a bug fix.

## What & why

Fixes [#263](https://github.com/chaitanyagiri/munder-difflin/issues/263). The report there listed three things as "Not established" — whether the write never persisted, whether it persisted but was not read on that path, or whether it was display-only. It is the second: the write always persisted, and the display read a stale snapshot. Affected set is the 14 prop-seeded fields the re-seed effect never re-reads, including `orchestratorMaySpawn` (the toggle in the original report), `autoCompactOn`, `autoUpdateOn`, `telemetryOn`, and the four breaker fields.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before
Screenshot 2026-08-24 at 6 37 30 PM

```
set REPEATED-TOOL LIMIT = 7, save   -> disk: repeatedToolLimit = 7      OK
close + reopen Settings             -> field shows: (blank)             WRONG
save again, touching nothing        -> disk: repeatedToolLimit gone     DATA LOSS

toggle "who can add agents"
  UI showed "me and Michael"        |  disk = false   <- already diverged
  click 1 -> "only me"              |  disk = false   <- wrote what was there
  click 2 -> "me and Michael"       |  disk = true    <- took two clicks
```

### After
Screenshot 2026-08-24 at 6 38 13 PM

```
set REPEATED-TOOL LIMIT = 55, save  -> disk: repeatedToolLimit = 55     OK
close + reopen Settings             -> field shows: 55                  FIXED
save again, touching nothing        -> disk: repeatedToolLimit = 55     FIXED

toggle "who can add agents"
  UI "me and Michael"               |  disk = true    <- agree
  click once -> "only me"           |  disk = false   <- one click, agree
  close + reopen -> "only me"       |  disk = false   <- agree
```

## How I tested it

- OS: macOS 15 (Darwin 25.5.0), Electron 32.3.3, Node 22.23.2
- Steps:
  1. Added `test/config-write-notify.test.cjs` — six tests over the public interface: a write notifies with the persisted config; a token-cap write notifies too (proving the hook sits at the choke point rather than on `writeConfig`); a reset notifies; unsubscribing stops delivery; a listener that throws neither fails the write nor starves the next listener; and subscribers receive the same shape a read returns. The last of those was written failing after a review pass and drove the hydration fix in 34adb6d. The resetConfig test was written failing first and drove routing it through `persistConfig`.
  2. Ran the app and drove Settings over CDP for the before/after transcripts above, checking the rendered value against `config.json` on disk at every step.
  3. Confirmed the broadcast actually reaches the renderer by subscribing to `onConfigChanged` in the page and observing the payload arrive on a write.
  4. Verified the notification carries the deep-filled shape by writing a partial trigger patch (`contextTrigger.compact.enabled` only) and observing the subscriber receive all 5 keys of `compact` rather than the 1 that was persisted.
  5. Opened a second floor window and confirmed a save made in one arrives in both.
  6. `npm run typecheck` clean; `npm run test:focused` 558 passing, 0 failing (552 before, plus the 6 new); `npm run build` succeeds.

One note on the suite: `test/hive-task-mutation.test.cjs` failed once during a full run with `ENOTEMPTY` while removing its own git temp dir. It passes 3/3 in isolation both with and without this branch's changes, and the full suite passed clean on re-run — a cleanup race under parallel execution, unrelated to this change.

## Discord (optional)

Discord:

## Checklist

      spacing, or fonts. (No new UI in this PR.)
      `ATTRIBUTION.md`.

---

### Known adjacent issue, deliberately not fixed here

`window.cth.updateConfig()` resolves with `writeConfig`'s return value, which is the raw persisted object rather than the read shape — so `AddAgentModal` (`:302`, `:317`, `:481`, `:490`) can push a half-filled config into the same `config` state via `onConfigChange`. That predates this PR and is the same class of bug; it wants either hydrating the IPC handler's return or routing those call sites through the notification. Left alone to keep this change to one thing — happy to file it separately.

### DIFF
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -621,19 +621,11 @@ function normalizeStoredHomes(cfg: HarnessConfig): HarnessConfig {
   return cfg;
 }
 
-/** Every persisted config write is announced here.
+/** Announces every saved setting, so a screen showing one can update.
  *
- *  The renderer holds the config as a snapshot loaded once at start-up, so
- *  before this existed a write reached disk but nothing told the UI, and any
- *  view seeded from that snapshot kept showing the pre-write value until the
- *  app restarted. Settings was the visible casualty: a toggle wrote correctly,
- *  then re-read the stale snapshot on reopen and displayed the old state (#263).
- *
- *  This sits on `persistConfig` rather than on `writeConfig` deliberately —
- *  `writeConfig` is one of 23 callers across the main process, and Slack,
- *  freeflow and notifications each persist through their own path. The file
- *  write is the only point every one of them has in common, so subscribing here
- *  is the difference between "the fields we remembered" and "all of them". */
+ *  Settings, Slack, voice and notifications each save by their own route, and
+ *  all of them end up writing the file below — so one subscription here covers
+ *  every setting rather than the ones anybody remembered to wire up. */
 type ConfigWriteListener = (next: HarnessConfig) => void;
 const configWriteListeners = new Set<ConfigWriteListener>();
 
@@ -646,19 +638,15 @@ function persistConfig(next: HarnessConfig): HarnessConfig {
   const p = configPath();
   mkdirSync(dirname(p), { recursive: true });
   writeFileSync(p, JSON.stringify(next, null, 2), 'utf8');
-  // Subscribers get the config in the same shape `readConfig` hands out, NOT the
-  // object we just wrote. They are the second source feeding a renderer that also
-  // calls config:get, and a patch touching one nested key persists a half-filled
-  // sub-object — `withTriggerDefaults` is what deep-fills it on the way out, so
-  // skipping it here would let a consumer read `undefined` from a trigger it had
-  // read a number from a moment earlier. The migration is deliberately not run:
-  // it persists, and it has already been applied to the config this patch built on.
+  // Saving one setting stores only that setting, so fill the rest back in first:
+  // subscribers must see the same complete config a read gives them, never a
+  // half-filled one. Skip the migration — it saves in its own right, and has
+  // already run against what this change was built on.
   const view = normalizeStoredHomes(withTriggerDefaults({ ...DEFAULTS, ...next }));
-  // A listener is a renderer send, which can fail while a window is closing.
-  // The config is already on disk by this point, so a throwing subscriber must
-  // neither fail the write for its caller nor starve the subscribers after it.
+  // The change is already saved, so one failed subscriber must not fail the save
+  // for its caller, nor stop the subscribers after it.
   for (const listener of configWriteListeners) {
-    try { listener(view); } catch { /* a broken listener is not a failed write */ }
+    try { listener(view); } catch { /* a broken listener is not a failed save */ }
   }
   return next;
 }
@@ -728,8 +716,7 @@ export function setAgentTokenCap(agentId: unknown, tokenCap: unknown): HarnessCo
 /** Wipe the persisted config back to first-run defaults so the app boots into
  *  onboarding again. Used by the "reset & start over" flow. */
 export function resetConfig(): HarnessConfig {
-  // Routed through persistConfig so the reset announces itself like any other
-  // write; the bytes written are the same DEFAULTS as before.
+  // Saved like any other change, so a reset announces itself too.
   persistConfig({ ...DEFAULTS });
   // Drop the migration latch too: the file on disk is back to `triggersMigratedV1:
   // false`, and a latch left set would keep the flag from ever being written again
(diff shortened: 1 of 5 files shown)


chaitanyagiri:
Shipped in v0.4.6, and it closes #263.

All four commits are in: `23485fd5` announces every config write, `34adb6d2` notifies with the config as read rather than as written, `fee76db0` trims the developer notes, and `94a03bb8` tells every window a setting was saved rather than only the focused one.

Your diagnosis was the useful part. The report listed three possibilities and you established it was the second: the write always persisted and the display read a stale snapshot, across the 14 prop seeded fields the re-seed effect never re-read.

Thank you.

Issue #287, opened 2026-08-24
MemPalace directory grew past 100GB in under 8 hours, filled with `.corrupt` folders


chaitanyagiri:
Reported in Discord `#help` by **phlaaj** on 22 Aug 2026.

> within <8 h work with a single agent hive on one project, the palace directory expanded to >100GB (got a low disk space warning). Not sure if this is your app or mempalace's issue, but looks like there are a large number of folders within palace with ".corrupt" in the name. Cleaned these out, but just FYI as an issue that can arise

**What happened:** one agent, one project, under eight hours, and the palace directory filled the disk. The reporter cleared it manually.

**The detail worth chasing:** the space is going into a large number of directories whose names contain `.corrupt`. That points at a failure path that writes a quarantine copy and never reclaims it, so every failure permanently costs disk.

**Worth establishing:**
- what creates a `.corrupt` directory, and whether anything ever deletes one
- whether the volume is normal growth or repeated re-quarantine of the same data
- whether a size ceiling or a retention window should apply to the palace directory
- whether this is MunderDifflin or MemPalace, which the reporter could not tell from outside

The Discord `#incident-log` separately lists "MemPalace crashes, investigating" for the memory layer restarting on some users. It is worth checking whether the crash loop and the `.corrupt` directories are the same failure seen from two sides.

Filed from Discord triage. Not reproduced locally.

PR #309, opened 2026-08-24
fix(hive): record why an agent died — forward the exit signal and keep a crash tail

### LINKED ISSUE #308: An agent killed by a provider crash is archived with no recorded cause — log.jsonl keeps only {kind:archive}
## What happened?

An agent whose provider CLI **crashes** is archived with no recorded cause. The
only durable trace of the death is:

```json
{"ts":1787604800527,"kind":"archive","agentId":"god","archived":true}
```

No exit code, no signal, no reason, no output. Grepping the whole `log.jsonl`
for `panic|crash|signal|illegal|exitCode` returns nothing, and the crash text is
**nowhere on disk** — not under `harnessHome`, not under `userData`. It exists
only in the UI terminal pane. Scroll it away, or simply not be looking, and the
evidence is gone.

So an agent that spawned at `15:53:18` and was archived at `15:53:20` having
never run is, in the only record the hive keeps, **indistinguishable from one
that finished its work and was archived normally**. Anything reading `log.jsonl`
to ask "is the floor healthy" sees a clean spawn/archive pair.

I expected the archive row — or a sibling row — to say *why* the agent stopped.

This is the same family as #277 (*"nothing is logged"*), but a different
mechanism: there the socket was never bound; here the agent process itself died.

### Root cause

`src/main/pty.ts` has the answer in hand and forwards half of it:

```js
proc.onExit(({ exitCode, signal }) => {
  this.safeSend(`pty:exit:${opts.id}`, { exitCode, signal }, session.owner);  // renderer: both
  this.sessions.delete(opts.id);
  try { this.exitHandler?.(opts.id, exitCode); } catch { /* … */ }            // main: code only
});
```

A process killed by a signal exits with **code 0** and a non-zero `signal`, so
every consumer of `exitHandler` is blind to signal deaths **by construction** —
no care downstream could have caught it. `index.ts`'s handler then calls
`teardownPty(id)`, which archives the agent, and nothing records why.

## Steps to reproduce

1. Fresh launch, complete onboarding, let Michael spawn.
2. Have the provider CLI die during startup. In my case it did so on its own —
   `claude` (a Bun standalone binary) panicked with `Illegal instruction`
   (SIGILL) 923 ms in. To force it deliberately, point the engine at a command
   that kills itself, e.g. `sh -c 'echo boom; kill -ILL $$'`.
3. Watch the agent card go away.
4. `cat "$HARNESS_HOME/hive/log.jsonl"`.

**Observed:** a `spawn` row, then an `archive` row ~2 s later, and nothing else.
**Expected:** something that names the exit code and signal.

## Logs / stack trace

What the terminal pane showed, and what was lost when it scrolled:

```
Bun v1.4.0 (fdb5e06cc) Linux x64
CPU: sse42 popcnt avx avx2 avx512
Args: ".../claude" "--model" "claude-opus-5" "--permission-mode" "bypassPermissions" …
Elapsed: 923ms
panic(main thread): Illegal instruction at address 0x52D6A9F
oh no: Bun has crashed. This indicates a bug in Bun, not your code.
─ process exited (code 0, signal 4) ─
```

Note `code 0, signal 4`. The pane itself prints both — the hive keeps neither.

The entire hive log for that session:

```
15:53:16  app-start   version 0.4.5, packaged=false
15:53:16  drop        reason=no-inbox, scheduler -> god
15:53:16  message     scheduler -> god  "Hourly ops standup"
15:53:18  spawn       agentId=god name=Michael isGod=true
15:53:20  archive     agentId=god archived=true
```

To be clear about scope: **the Bun panic is not your bug** — that one is
upstream of you, and it was not reproducible (5 targeted attempts: prompt size,
flag shape, environment, the real `--settings` payload with 6 MCP servers, and
finally through node-pty under Electron in the hive CWD — all clean). What is
reproducible, and what this issue is about, is that when a provider *does* die,
Munder keeps nothing.

## Environment

- **OS:** Linux · **OS version:** Ubuntu 24.04.4 LTS, kernel 7.0.0-30-generic, x86_64
- **Munder Difflin version:** 0.4.5 (local dev build from the v0.4.5 tag, `1f6958b`)
- **Node:** v24.13.1 · **Electron:** 32.3.3 (ABI 128)
- **Agent CLI:** Claude Code 2.1.241

## A fix

PR incoming. Shape, in case you'd rather take a different one:

- `pty.ts` keeps a bounded 8 KB rolling tail per session and passes
  `{signal, tail, command, cwd}` to the exit handler as a `PtyExitInfo` object.
  By value deliberately — `onExit` deletes the session from the map *before*
  calling the handler, so a handler that tried to look the command up would get
  `undefined`.
- `hive.recordAgentExit()` splits the record: structured, non-sensitive fields
  (`exitCode`, `signal`, `abnormal`, `tailPath`) go to `log.jsonl`; the raw tail
  goes to `crashes/`, which is added to the hive's `.gitignore`. That split is
  deliberate — the hive is a git repo committed on every change, and the tail is
  whatever the provider printed, so a token captured there would be permanent.
- A clean exit records nothing. This is a diagnostic, not an audit log.

8 tests, both files shown failing against unpatched `pty.ts` / `hive.ts` before
being trusted.

LavaDMan:
## What & why

When an agent's provider CLI dies, the hive keeps no reason. The whole durable
record of the death is `{"kind":"archive","agentId":"god","archived":true}` — no
exit code, no signal, no output — so an agent that crashed two seconds after
spawning is indistinguishable from one that finished its work. The crash text
lives only in the UI terminal pane and is gone when it scrolls.

The cause is upstream of the recording: `pty.ts` receives `{ exitCode, signal }`
from node-pty, sends both to the renderer, and forwards only `exitCode` to
`exitHandler`. A process killed by a signal exits with **code 0** and a non-zero
signal, so every consumer in the main process was blind to signal deaths by
construction.

This forwards the signal, keeps a bounded tail of what the process printed, and
writes both down — splitting them so the raw output never enters the hive's git
history.

Closes #308

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Design notes

- **`PtyExitInfo` is passed by value on purpose.** `onExit` deletes the session
  from `sessions` *before* calling the handler, so a handler that tried to look
  the command up via `list()` would always get `undefined`. (I wrote that
  version first, then found it.)
- **The tail is capped at 8 KB** and sliced after append, so a single oversized
  write still leaves its end — the part that explains a death.
- **`log.jsonl` gets structured fields; `crashes/` gets the raw tail, and is
  added to the hive's `.gitignore`.** The hive is a git repo committed on every
  change, and the tail is whatever the provider printed — tokens, paths, prompt
  fragments. A secret written into `log.jsonl` would be permanent. There is a
  test asserting a fake key reaches the dump and never the log.
- **A clean exit records nothing.** This is a diagnostic, not an audit log, and
  a normal archive should stay quiet.

### Before
before

The tests in this PR, run against **unpatched `hive.ts` and `pty.ts`** (checked
out from the tag, patch reverted) — they fail for the reason they were written:

```
$ git show <tag>:src/main/hive.ts > src/main/hive.ts     # revert the fix
$ node --test test/agent-exit-record.test.cjs
✖ a signal death is recorded even though its exit code is 0
✖ a non-zero exit code is recorded
✖ a CLEAN exit records nothing — this is a diagnostic, not an audit log
✖ raw output goes to a gitignored crashes/ file, NOT into log.jsonl
✖ a death is still recorded when there is no output to dump
ℹ tests 5   ℹ pass 0   ℹ fail 5
exit 1

$ git show <tag>:src/main/pty.ts > src/main/pty.ts       # revert the fix
$ node --test test/agent-exit-signal-e2e.test.cjs
✖ a SIGILL death forwards the signal and the output that explains it
✖ a clean exit reports no signal
✖ a non-zero exit is reported as a code, not a signal
ℹ tests 3   ℹ pass 0   ℹ fail 3
exit 1
```

And the real-world shape of it — the hive's entire record of an agent that
crashed on startup and never ran:

```
15:53:16  app-start   version 0.4.5, packaged=false
15:53:18  spawn       agentId=god name=Michael isGod=true
15:53:20  archive     agentId=god archived=true
```

while the terminal pane (and nothing on disk) showed:

```
panic(main thread): Illegal instruction at address 0x52D6A9F
─ process exited (code 0, signal 4) ─
```

### After
after

Same tests, patch applied:

```
$ node --test test/agent-exit-record.test.cjs test/agent-exit-signal-e2e.test.cjs
✔ a signal death is recorded even though its exit code is 0
✔ a non-zero exit code is recorded
✔ a CLEAN exit records nothing — this is a diagnostic, not an audit log
✔ raw output goes to a gitignored crashes/ file, NOT into log.jsonl
✔ a death is still recorded when there is no output to dump
✔ a SIGILL death forwards the signal and the output that explains it
✔ a clean exit reports no signal
✔ a non-zero exit is reported as a code, not a signal
ℹ tests 8   ℹ pass 8   ℹ fail 0
exit 0
```

The e2e test drives the real `PtyManager` against a process that prints a panic
banner and then `kill -ILL $$` — the exact shape of the crash that motivated
this — and asserts the signal survives the trip out of `pty.ts`.

What the hive now records for that same death:

```json
{"ts":…,"kind":"agent-exit","agentId":"god","exitCode":0,"signal":4,
 "abnormal":true,"tailPath":"…/hive/crashes/2026-08-24T…-god.log"}
```

and `crashes/…-god.log`:

```
agent:    god
exitCode: 0
signal:   4
command:  /home/…/claude --model … --permission-mode bypassPermissions
captured: 2026-08-24T…
--- last N bytes of pty output ---
panic(main thread): Illegal instruction at address 0x52D6A9F
```

### Gate

Run on this branch, rebased on current `main`:

```
npm run typecheck     → 0
npm run build         → 0
npm run test:focused  → 560 tests, 559 pass, 1 fail
```

The suite goes from 552 to 560 tests. **The one failure is pre-existing and
unrelated** — `hive-hook-node.test.cjs` races its own stdin write and fails on
EPIPE on Linux. It fails identically on unmodified `main`. Fixed separately in
the companion PR; with both applied the suite is 560/560.

## Testing

- Linux (Ubuntu 24.04.4, kernel 7.0.0-30, x86_64), Node v24.13.1, Electron
  32.3.3, Claude Code 2.1.241.
- **Not tested on macOS or Windows.** The signal path is POSIX; both new test
  files skip on `win32`. `recordAgentExit` itself is platform-agnostic, and the
  crash-dump filename already replaces `:` so it is Windows-safe, but I have not
  run it there.

### DIFF
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -546,7 +546,12 @@ export class HiveManager {
 
     // Keep the churny/ephemeral live files out of the hive git repo.
     const gitignore = join(root, '.gitignore');
-    const want = ['fleet.json', 'hooks.sock', 'cost-ledger.jsonl', '.DS_Store'];
+    // `crashes/` holds raw PTY output from abnormal agent exits. It is
+    // DELIBERATELY ignored: that output is whatever the provider printed, which
+    // can include tokens, paths and prompt fragments, and the hive repo is
+    // committed on every change — a secret written there would be permanent.
+    // log.jsonl gets the structured, non-sensitive fields; the dump stays local.
+    const want = ['fleet.json', 'hooks.sock', 'cost-ledger.jsonl', 'crashes/', '.DS_Store'];
     let lines: string[] = [];
     if (existsSync(gitignore)) { try { lines = readFileSync(gitignore, 'utf8').split('\n'); } catch { lines = []; } }
     const missing = want.filter((w) => !lines.includes(w));
@@ -2226,6 +2231,66 @@ export class HiveManager {
       .sort()
       .map((f) => { try { return JSON.parse(readFileSync(join(dir, f), 'utf8')) as HiveMessage; } catch { return null; } })
       .filter((m): m is HiveMessage => m !== null);
+  }
+
+  /**
+   * Record an agent's process exit so a death has a durable cause.
+   *
+   * Before this, an agent killed by its provider crashing was archived with
+   * `{kind:'archive', agentId, archived:true}` and NOTHING else — no exit code,
+   * no signal, no output. A two-second SIGILL death and a completed agent were
+   * indistinguishable in the only record the hive keeps, and the crash banner
+   * lived solely in a UI terminal pane. Observed live 2026-08-24: Michael's
+   * `claude` CLI panicked 923ms into startup and left no trace on disk.
+   *
+   * Split by design:
+   *   - log.jsonl  gets structured, non-sensitive fields (code, signal, path).
+   *   - crashes/   gets the raw tail, and is gitignored — see ensureHive.
+   * A normal exit writes nothing at all; this is a diagnostic, not an audit log.
+   */
+  recordAgentExit(
+    agentId: string,
+    info: { exitCode?: number; signal?: number; tail?: string; command?: string }
+  ): void {
+    const root = this.root();
+    if (!root) return;
+    const { exitCode, signal, tail, command } = info;
+    // A signal means killed (SIGILL/SIGSEGV/SIGKILL); node-pty reports exitCode 0
+    // in that case, so signal must be checked independently of the code.
+    const abnormal = (typeof signal === 'number' && signal !== 0) || (typeof exitCode === 'number' && exitCode !== 0);
+    if (!abnormal) return;
+
+    let tailPath: string | null = null;
+    if (tail && tail.length) {
+      try {
+        const dir = join(root, 'crashes');
+        mkdirSync(dir, { recursive: true });
+        // Colons are illegal in filenames on Windows and awkward everywhere.
+        const stamp = new Date().toISOString().replace(/[:.]/g, '-');
+        const safeId = agentId.replace(/[^A-Za-z0-9._-]/g, '_');
+        const p = join(dir, `${stamp}-${safeId}.log`);
+        const header = [
+          `agent:    ${agentId}`,
+          `exitCode: ${String(exitCode)}`,
+          `signal:   ${String(signal)}`,
+          command ? `command:  ${command}` : null,
+          `captured: ${new Date().toISOString()}`,
+          `--- last ${tail.length} bytes of pty output ---`,
+          ''
+        ].filter(Boolean).join('\n');
+        writeFileSync(p, header + tail, 'utf8');
+        tailPath = p;
+      } catch { /* a diagnostic must never break teardown */ }
+    }
+
+    this.appendLog({
+      kind: 'agent-exit',
+      agentId,
+      exitCode: exitCode ?? null,
+      signal: signal ?? null,
+      abnormal: true,
+      tailPath
+    });
   }
 
   // — log —
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -575,7 +575,23 @@ function removeWorkerScratch(workerId: string): void {
 // SAME pty/window (no user click). Provider-agnostic. Idempotent by construction: the
 // relaunch carries `noAutoInstall`, so the installer can never fire (let alone loop) a
 // second time — a binary that's somehow still missing just spawns and exits normally.
-ptyManager.setExitHandler((id, exitCode) => {
+ptyManager.setExitHandler((id, exitCode, info) => {
+  // Record an ABNORMAL death before teardown — teardownPty drops the
+  // pty->agent mapping, so after it runs we can no longer say WHOSE process
+  // died. Only abnormal exits are recorded (recordAgentExit returns early on a
+  // clean one), so this adds no noise to a normal archive.
+  try {
+    const dyingAgent = ptyToAgent.get(id);
+    if (dyingAgent) {
+      hive.recordAgentExit(dyingAgent, {
+        exitCode,
+        signal: info?.signal,
+        tail: info?.tail,
+        command: info?.command
+      });
+    }
+  } catch (e) { console.error('[pty] recordAgentExit failed:', e); }
+
   const pending = pendingInstallRelaunch.get(id);
   if (pending) {
     pendingInstallRelaunch.delete(id);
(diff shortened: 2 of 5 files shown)


chaitanyagiri:

---
Closes #308. Gitignoring `crashes/` because the hive repo auto commits on every change, and the
raw PTY tail can carry a token, is the right call and the comment says exactly that. So is
forwarding `signal` separately, since node-pty reports exit code 0 for a SIGILL death and an
exitCode-only handler cannot tell a crash from a clean finish. Passing the tail by value because
the session is already out of the map by the time the handler runs is the third one. Three traps,
three comments, all correct.

**Two follow ups, both on the file that gets written.**

**It is not written 0600.** `writeFileSync(p, header + tail, 'utf8')` takes the process umask, so
on a normal machine that is a world readable file. The comment two hundred lines up already
establishes that this content can include tokens, paths and prompt fragments, which is precisely
why it is kept out of git. The same reasoning applies to the file mode. Worth noting that a
packaged Electron app does not inherit the launching shell's umask the way a dev build does, so
the mode you see locally is not necessarily the mode a user gets. `{ mode: 0o600 }` removes the
question.

**Nothing prunes `crashes/`.** One file per abnormal exit, forever, 8 KB each. The scenario in
#308 is a provider that panics 923 ms into startup, and the harness relaunches, so the crash loop
that most needs this diagnostic is also the one that writes the most files. #287 in the backlog is
`MemPalace directory grew past 100GB in under 8 hours`, so an unbounded write-per-event directory
in this app is not a hypothetical. A cap on file count, or a sweep of anything older than a few
days, would close it.

`A normal exit writes nothing at all; this is a diagnostic, not an audit log` is a good line to
have in the docstring, because it tells the next person what the file is allowed to become.

PR #310, opened 2026-08-24
fix(test): hive-hook-node races its own stdin write and fails on EPIPE

LavaDMan:
## What & why

`test/hive-hook-node.test.cjs` fails on Linux with `write EPIPE`. The fault is
in the test's own `run()` helper, not in anything it tests: it writes a hook
payload to a child's stdin without an `'error'` listener, and the test
deliberately runs commands that exit immediately — its own control case asserts
that `node "<shim>"` exits **127** under a stripped PATH. When that child is
gone before the write lands, the unhandled EPIPE rejects the whole test.

Adding the listener is enough. The helper reports the child's exit **code**,
which is what every assertion keys on, so a write that never lands loses
nothing.

It is timing-dependent, which is probably why it has gone unreported: on an idle
machine the write usually wins. On a loaded one it does not.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before
before

On unmodified `main`, Ubuntu 24.04.4 / Node v24.13.1, under load. The file
alone, **4 of 5 runs fail** — run 5 passing is the race being a race:

```
$ for i in 1 2 3 4 5; do node --test test/hive-hook-node.test.cjs; done
  run 1: exit=1   pass 4   fail 1
  run 2: exit=1   pass 4   fail 1
  run 3: exit=1   pass 4   fail 1
  run 4: exit=1   pass 4   fail 1
  run 5: exit=0   pass 5   fail 0     <- won the race
```

with:

```
✖ a hook fires with NO node on PATH, and its payload reaches HIVE_SOCK
  Error: write EPIPE
      at Socket.end (node:net:737:31)
      at .../test/hive-hook-node.test.cjs:71:17
    errno: -32, code: 'EPIPE', syscall: 'write'
```

Line 71 is `child.stdin.end(...)` inside `run()`.

Across sessions I measured the full suite failing 4 of 5 runs on the same
machine, and the file alone failing 5 of 5 at higher load. **The pass rate moves
with machine load**, which is the point: CI may well be green while this is
broken for contributors.

### After
after

Same machine, same load, one added listener. The file, **10 of 10 pass**, and
the full suite **3 of 3 fully green at 552/552**.

### Gate

```
npm run typecheck     → 0
npm run test:focused  → 552/552, exit 0
npm run build         → 0
```

## Testing

Linux only (Ubuntu 24.04.4, Node v24.13.1). The test already skips on `win32`.
The change is a no-op wherever the write currently wins the race — it only stops
an expected EPIPE from failing the test.

## Note

Independent of #309 and cuts from `main`, so the two can merge in either order.
With both applied the suite is 560/560.

### DIFF
--- a/test/hive-hook-node.test.cjs
+++ b/test/hive-hook-node.test.cjs
@@ -68,6 +68,12 @@ async function run(cmd, env) {
     const child = spawn('/bin/sh', ['-c', cmd], { env, stdio: ['pipe', 'pipe', 'pipe'] });
     let stderr = '';
     child.stderr.on('data', (d) => { stderr += d; });
+    // These commands are EXPECTED to exit before the payload lands — the control
+    // case below asserts `node "<shim>"` exits 127 under a stripped PATH. A child
+    // that is already gone makes this write EPIPE, and an unhandled 'error' on
+    // stdin rejects the test rather than the command under test failing. The exit
+    // code is what this helper reports, so a lost write is not a lost signal.
+    child.stdin.on('error', () => { /* child exited before reading stdin */ });
     child.stdin.end(JSON.stringify({ hook_event_name: 'Stop', session_id: 's1' }));
     child.on('close', (code) => resolve({ code, stderr }));
   });
--- a/test/hive-hook-node.test.cjs
+++ b/test/hive-hook-node.test.cjs
@@ -73,7 +73,13 @@ async function run(cmd, env) {
     // that is already gone makes this write EPIPE, and an unhandled 'error' on
     // stdin rejects the test rather than the command under test failing. The exit
     // code is what this helper reports, so a lost write is not a lost signal.
-    child.stdin.on('error', () => { /* child exited before reading stdin */ });
+    // Narrow on purpose: EPIPE is the expected one and is swallowed, anything
+    // else (EACCES, ERR_STREAM_DESTROYED) is surfaced through the stderr this
+    // helper already returns, so an unexpected stdin fault names itself instead
+    // of vanishing.
+    child.stdin.on('error', (e) => {
+      if (e.code !== 'EPIPE') stderr += `stdin: ${e.message}\n`;
+    });
     child.stdin.end(JSON.stringify({ hook_event_name: 'Stop', session_id: 's1' }));
     child.on('close', (code) => resolve({ code, stderr }));
   });


hallolle:
## Review: verified against the source, not the description — **ready to merge**

I reproduced the bug, traced it in the checked-out tree, and ran a mutation test to check the specific thing a `catch-and-ignore` patch is usually guilty of. It holds up.

### The diagnosis is correct

`test/hive-hook-node.test.cjs:71` is `child.stdin.end(JSON.stringify({...}))` on a pipe with no `'error'` listener, and `run()` is called on children that are designed to be gone by then:

- `test/hive-hook-node.test.cjs:197` — `run('command -v node', env)`, a shell builtin that exits before the write is scheduled.
- `test/hive-hook-node.test.cjs:199` — `run('node "${shim}"', env)`, the control case whose *whole point* is asserting `127` at line 200.

So the failing write is in the harness, not in anything under test. That matches the stack in the PR body exactly.

### Independently reproduced — and on a newer Node than you tested

You reported Ubuntu 24.04.4 / Node v24.13.1. On this machine (**Node v26.5.0**, branch `jochen/free-openrouter-workers`, where `test/hive-hook-node.test.cjs` is byte-identical to `main`):

```
unpatched: 5 runs → 2 × exit=1 `Error: write EPIPE`, 3 × exit=0
patched:  10 runs → 10 × exit=0
```

Same race, two majors later. Worth having in the record.

### It is not papering over the failure — I checked

The obvious objection to `on('error', () => {})` is that it also silences the case where the *real* hook exits before reading stdin. It doesn't, and the reason is that the write was never the signal. I mutated the shim template at `src/main/hive.ts:2532` (`HOOK_SHIM`) to `process.exit(0)` before touching stdin — precisely the failure the swallow could hide — and with your patch applied the test still fails:

```
AssertionError [ERR_ASSERTION]: nothing arrived at HIVE_SOCK
```

`test/hive-hook-node.test.cjs:208` (exit code `0`) and `:211-212` (payload actually landed on the socket) carry the assertion, and both survive. Also confirmed the healthy path can't race: the shim at `src/main/hive.ts:2536-2539` buffers stdin and only acts on `'end'`, so the write always lands where it matters.

### Scope and gate

One commit, `test/hive-hook-node.test.cjs` only, +6/−0 — no `skip`, no `t.todo`, no assertion loosened. Full suite in a clean worktree off your head: **552/552, exit 0**, matching your claim.

The inline comment explaining *why* the swallow is safe is the part that makes this reviewable at all; a bare `on('error', () => {})` with no rationale is what would have earned a request-changes here.

---

### Nits — non-blocking, take or leave

**1.** The handler swallows *every* stdin error, not just `EPIPE`. An `EACCES` or `ERR_STREAM_DESTROYED` would vanish silently and leave a confusing failure downstream. Keeping it visible costs nothing, since `stderr` is already in scope at `test/hive-hook-node.test.cjs:69`:

```js
child.stdin.on('error', (e) => {
  // child exited before reading stdin — expected for the 127/builtin cases
  if (e.code !== 'EPIPE') stderr += `stdin: ${e.message}\n`;
});
```

**2.** Pre-existing, *not* yours, flagging for a follow-up rather than this PR — `run()` at `test/hive-hook-node.test.cjs:66-74` has two more sharp edges:
- no `child.on('error')`, so a spawn failure leaves the promise pending forever and the test dies on timeout instead of naming the cause;
- `stdout` is `'pipe'` (line 68) but never drained, so a child writing past the pipe buffer would block and never emit `'close'`.

Both are latent today. Neither should hold this up.

**Verdict: ready.** Test-only, minimal, correctly diagnosed, evidence reproduces.

<sub>Automated review — read against the checked-out source, mutation-tested, no changes pushed.</sub>

LavaDMan:
Thank you — the mutation test is the review I'd have wanted and didn't do. Swallowing an error is exactly where a reviewer should assume the worst, and "the write was never the signal" is a cleaner statement of why it's safe than anything in my PR body.

**Nit 1: taken.** Pushed in `4cdd039`.

```js
child.stdin.on('error', (e) => {
  if (e.code !== 'EPIPE') stderr += `stdin: ${e.message}\n`;
});
```

You're right that the broad version had a cost for no benefit. EPIPE is the one that's expected; anything else now goes into the `stderr` the helper already collects and returns, so an unexpected stdin fault names itself instead of vanishing.

Since narrowing the handler could in principle have weakened the thing you verified, I re-ran your mutation against the new version rather than assuming it still held. `HOOK_SHIM` mutated to `process.exit(0)` immediately after `'use strict';`, before stdin is touched:

```
AssertionError [ERR_ASSERTION]: nothing arrived at HIVE_SOCK
    at test/hive-hook-node.test.cjs:211
ℹ pass 4  ℹ fail 1     exit 1
```

Same assertion you reported. Unmutated: 10/10 green, full suite 552/552 exit 0.

(First attempt at reproducing your mutation I injected above the shebang and got a syntax error at `cth-hook.cjs:3` instead — a different failure that would have "passed" the check for the wrong reason. Worth mentioning only because it's the kind of near-miss that makes a mutation test look like it worked when it didn't.)

**Nit 2: agreed, and left alone.** Both are real:

- no `child.on('error')`, so a spawn failure leaves the promise pending and the test dies on timeout rather than naming the cause;
- `stdout` is `'pipe'` but never drained, so a child writing past the pipe buffer would block and never emit `'close'`.

Neither is mine and both are latent, so I've kept this test-only and minimal as you suggested. Happy to send a separate PR for `run()`'s remaining sharp edges if you'd want it — it'd be the same helper, one more commit, and it would close the third way this function can hang.

**On the independent repro:** thank you for running it on Node v26.5.0. Two majors apart with the same race is a much better record than my single-machine 24.04/v24.13.1 sample, and your 2-of-5 unpatched rate against my 4-of-5 is a useful spread — it's load-sensitive, which is probably why it has gone unreported.

chaitanyagiri:
Shipped in v0.4.6.

Both commits are in: `74a3917e` for the stdin race in `test/hive-hook-node.test.cjs`, and `4cdd0393` narrowing the swallow to EPIPE specifically rather than everything.

The narrowing is the part worth having. A blanket swallow there would have hidden real failures.

Thank you.

PR #312, opened 2026-08-25
feat(layout): horizontal split option — floor on top, terminal below

### LINKED ISSUE #311: Split is vertical-only — no way to put the floor on top and the terminal below
## What

The floor/sidebar divider is always vertical. There is no orientation setting anywhere in the app, so the office floor is always a column on the left and the agent sidebar is always a column on the right.

## Why it matters

The office map is 544×352 — a 1.55:1 landscape image — and `Camera.getMinZoom` contain-fits it:

```ts
return Math.min(this.viewWidth / this.mapWidth, this.viewHeight / this.mapHeight);
```

So whichever axis is over-provisioned becomes dead letterbox space. In a vertical split the floor pane is always tall and narrow, which is the wrong shape for a landscape map.

**Today — side by side.** 1895×1194 window. The floor pane is 788×1020 but the map renders 788×510, so half the column is the black band above and below it. The terminal gets the remaining ~1085px.

Today: vertical split, floor letterboxed, terminal ~1085px

**With a horizontal divider.** Same window, same session. The floor becomes a strip and the terminal takes the full width — ~1085px → ~1863px, about **70% more columns** for agent output, diffs and tables.

Proposed: horizontal split, terminal full width ~1863px

To be straight about the trade: on this wide window the floor renders at the same 788×510 in both — its unused space just moves from top/bottom to the sides. The win here is entirely terminal width. On a taller, narrower window the floor gains too, since that is where the vertical split squeezes it hardest.

## Proposal

A title-bar toggle that flips the divider between vertical (side by side, today's behaviour) and horizontal (floor on top, terminal below), with each orientation remembering its own pane size.

Scoped deliberately small: no change to the floor, the terminal, or focus mode. The default stays exactly as it is today, so nobody's layout moves unless they ask for it.

## Offer

I have this working locally against `main` — ~380 lines across 5 files, 11 new unit tests, typecheck/tests/build green. Happy to open the PR if you'd take it. Equally happy to hear it's not wanted before I send it.

ketan0095:
## What & why

The floor/sidebar divider was hard-coded vertical in three separate places: `SidebarSplitter` read `clientX` and set `ew-resize`, `App` laid the panes out in a flex row, and the store held a single `sidebarWidth` scalar. There was no orientation setting anywhere.

That is the wrong shape for the floor. The office map is 544×352 (1.55:1) and `Camera.getMinZoom` contain-fits it, so a tall narrow pane is width-limited and letterboxes the rest. On the 1895×1194 window below, the floor pane is 788×1020 and the map renders 788×510 — half the column is dead space — while the terminal is confined to the remaining ~1085px.

Stacking the panes gives the terminal the full window width: ~1085px → ~1863px, roughly 70% more columns for agent output. The floor renders at the same 788×510; its unused space just moves from vertical bands to side bars. On a taller, narrower window the floor gains as well.

This adds the option and changes nothing by default — `vertical` stays the default and is byte-for-byte the previous behaviour.

Closes #311

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before
Before — side-by-side split, floor letterboxed, terminal ~1085px
Side-by-side (today's only layout), 1895×1194. Floor pane 788×1020, map rendered 788×510 — note the black bands above and below it. Terminal ~1085px wide.

### After
After — horizontal split, terminal full width ~1863px
Same window, same size, same theme, same session, after clicking the new toggle. Floor becomes a strip; terminal is full width at ~1863px.

## How I tested it

- OS: macOS 15 (Darwin 25.5.0), Apple Silicon
- Steps:
  - `npm run typecheck` — clean, node + web
  - `npm run test:focused` — 563 pass, 0 fail (552 existing + 11 new in `test/split-layout.test.cjs`)
  - `npm run build` — clean
  - Ran the production build and flipped the toggle both ways; dragged the divider in both orientations; double-clicked to reset in both; entered and left focus mode; quit and relaunched to confirm the orientation and both pane sizes persist.

## Checklist

- [ ] If I added art, it's my own or compatibly licensed, and listed in `ATTRIBUTION.md`. *(n/a — no new art; the two glyphs are stroke paths matching the existing `ExpandGlyph`/`CollapseGlyph`.)*

## Notes for the reviewer

**Why the toggle is its own button and not folded into focus mode.** They are independent axes: focus mode owns *how many* panes, orientation owns *how they sit*. A single cycling control would force a user through a layout they don't want to reach the one they do, and would desync the moment `Esc` left focus mode. The button is hidden — not disabled — while focus mode is on, since there is no split to orient and the focus button beside it already explains where the floor went.

**Why each orientation keeps its own size.** Carrying one scalar across a flip turns a 420px-wide sidebar into a 420px-tall one — a small terminal under a huge floor, the inverse of what the flip was asked for. `sidebarWidth` and `sidebarHeight` are stored separately.

**Why the minima differ per orientation.** The sidebar holds a terminal and its axes are not interchangeable: 320px of width is ~40 columns, below which output wraps badly; 240px of height is ~15 rows, cramped but usable. The floor's reserve also drops 360 → 180, or the strip layout this exists for would be impossible to drag to.

**No refit plumbing was needed.** `OfficeFloor`'s `ResizeObserver` already calls `camera.setViewSize` (which refits), and `PtyTerminalView`'s observes the terminal host, so both panes re-fit themselves on the flip. Worth knowing: `Camera.setViewSize` only refits while `manualOverride` is false, and that is always true today because `focusOn()` — its only writer — is currently unreachable. If `focusOn` is ever wired up, the flip will need an explicit `fitToScreen()`.

**One pre-existing interaction.** `cth.prefersFocusMode` restores focus mode on launch, so quitting in focus mode reopens in it with the orientation button hidden until `Esc`. Not introduced here, but it is the one way the button can appear to be missing.

### DIFF
--- a/src/renderer/src/App.tsx
+++ b/src/renderer/src/App.tsx
@@ -385,8 +385,9 @@ export function App() {
             data-tip={splitOrientation === 'horizontal'
               ? 'Side-by-side layout'
               : 'Stacked layout — floor on top'}
-            aria-label="Toggle split orientation"
-            aria-pressed={splitOrientation === 'horizontal'}
+            aria-label={splitOrientation === 'horizontal'
+              ? 'Switch to side-by-side layout'
+              : 'Switch to stacked layout — floor on top'}
             style={{
               display: 'inline-flex', alignItems: 'center', justifyContent: 'center',
               width: 28, height: 28, padding: 0,
--- a/src/renderer/src/App.tsx
+++ b/src/renderer/src/App.tsx
@@ -54,7 +54,10 @@ export function App() {
   const fullscreenAgentId = useStore(s => s.fullscreenAgentId);
   const appThemeNow = useAppTheme();
   const sidebarWidth = useStore(s => s.sidebarWidth);
-  const setSidebarWidth = useStore(s => s.setSidebarWidth);
+  const sidebarHeight = useStore(s => s.sidebarHeight);
+  const splitOrientation = useStore(s => s.splitOrientation);
+  const setSidebarSize = useStore(s => s.setSidebarSize);
+  const toggleSplitOrientation = useStore(s => s.toggleSplitOrientation);
   const ideOpen = useStore(s => s.ideOpen);
   const setIdeOpen = useStore(s => s.setIdeOpen);
 
@@ -79,6 +82,7 @@ export function App() {
   const [quitWarn, setQuitWarn] = useState<{ ptyCount: number } | null>(null);
   const [closing, setClosing] = useState<ClosingTimeState | null>(null);
   const [vpWidth, setVpWidth] = useState<number>(window.innerWidth);
+  const [vpHeight, setVpHeight] = useState<number>(window.innerHeight);
 
   // Deep link into Settings from anywhere in the tree. Settings' open state is
   // local to App, so a nested control (e.g. "set it now" beside a disabled Talk
@@ -245,9 +249,13 @@ export function App() {
     useStore.getState().restoreFocusMode();
   }, [config?.onboardingComplete, agents]);
 
-  // Track viewport width for splitter clamping
+  // Track viewport size for splitter clamping — both axes, since the split
+  // divider can run either way and each clamps against its own extent.
   useEffect(() => {
-    const onResize = () => setVpWidth(window.innerWidth);
+    const onResize = () => {
+      setVpWidth(window.innerWidth);
+      setVpHeight(window.innerHeight);
+    };
     window.addEventListener('resize', onResize);
     return () => window.removeEventListener('resize', onResize);
   }, []);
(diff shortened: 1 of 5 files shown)


ketan0095:
Rebased onto current main (b91a49f). The earlier conflict was a single import collision in `store.ts` with `rosterSource` — resolved, nothing else moved. Re-verified on the rebased tree: `npm run typecheck` clean, `npm run test:focused` 756/756.

@chaitanyagiri — ready for review whenever you have a window. Default behaviour is unchanged: `vertical` stays the default and is byte-for-byte the previous layout, with the horizontal option opt-in via the new title-bar toggle.

No rush given the queue — flagging only because it's now conflict-free and green.

chaitanyagiri:

---
Closes #311. One component for both axes with the reason written down, `splitting it in two is how
the two copies drift`, is the right call, and so is hiding rather than disabling the control in
focus mode with the explanation of why a greyed control would be worse.

**One concrete thing, and it is funny given what is currently in Discord.** The control tells the
user two opposite things at once:

```
data-tip={splitOrientation === 'horizontal' ? 'Side-by-side layout' : 'Stacked layout — floor on top'}
aria-pressed={splitOrientation === 'horizontal'}
```

The tooltip describes what clicking WILL DO. `aria-pressed` describes what the state IS. So a
sighted user reads the action and a screen reader user is told the state, and they are opposites.
`aria-pressed` is also the wrong role for a control that swaps between two equal modes rather than
turning one thing on and off. This is the exact confusion users raised in `#help` about settings
toggles: not being able to tell whether a control shows its state or its action. Either make the
tooltip state-describing to match, or drop `aria-pressed` and give it an `aria-label` that names
the current layout.

Adding `minWidth: 0` alongside the existing `minHeight: 0` is the detail that stops a column-
direction flex container refusing to shrink. Easy to leave out and hard to debug afterwards.

ketan0095:
Fixed in 0704ae1a — went with your second option. `aria-pressed` is gone and the label is action-describing, so the tooltip and the accessible name now say the same thing in both states:

```jsx
data-tip={splitOrientation === 'horizontal'
  ? 'Side-by-side layout'
  : 'Stacked layout — floor on top'}
aria-label={splitOrientation === 'horizontal'
  ? 'Switch to side-by-side layout'
  : 'Switch to stacked layout — floor on top'}
```

Two reasons for that direction rather than making the tooltip state-describing. It keeps this control in line with its two title-bar siblings — the theme and focus-mode toggles both pair an action-describing `data-tip` with an `aria-label`, and neither carries `aria-pressed`. And the new label keeps the tooltip's wording as a substring, so the accessible name still contains the visible label (WCAG 2.5.3).

Agreed the role was wrong, and the repo already draws that line correctly: the other two `aria-pressed` uses, in `triggers/ui.tsx` and `FullscreenTerminal.tsx`, are a real switch and a collapsed panel. This one swaps between two equal layouts, so neither is the "off" one.

One thing I found while fixing it, which I've deliberately left out of this PR. An accessible-name change on an already-focused element isn't a state change, so it isn't reliably announced — the button keeps focus across the flip, so a screen-reader user can activate it and hear nothing. `aria-pressed` did sit in the state-change path, so strictly this trades a confusing announcement for a clear name and no announcement. That label is also the only carrier of layout state anywhere: `SidebarSplitter` renders a bare `div` with `onMouseDown`, and no `role="separator"`, `aria-orientation` or `tabIndex`.

Closing that properly is either a visually-hidden live region — there's no sr-only utility in the repo yet, so that's a new pattern — or giving the splitter a real separator role, which would also make pane resizing keyboard-reachable since it's mouse-only today. Both read as a second change to me rather than part of this fix. Want either in here, or shall I file it separately?

ketan0095:
Bump on this one — and thanks for merging #410.

Your review here is answered in 0704ae1a, and it hasn't rotted since: 99 commits later it still merges cleanly into main, and with current main merged in locally the typecheck is clean and the 11 split-layout tests still pass.

I haven't rebased it, since CONTRIBUTING says not to force-push once a review has started — say the word if you'd rather have it brought up to date and I'll push one.

The only thing still open is the question at the end of my last reply: whether you want the activation-announcement gap closed here (a visually-hidden live region), or on the splitter (`role="separator"` + `aria-orientation`, which would also make pane resizing keyboard-reachable), or filed separately. Happy either way — just don't want it sitting blocked on me.

PR #315, opened 2026-08-25
fix(terminal): only run the fit/fonts settle cascade on first open (#314)

### LINKED ISSUE #314: Configurable shell for spawned agent terminals on Windows (PowerShell/pwsh instead of hardcoded cmd)
### Summary
On Windows the app spawns each agent terminal through node-pty/ConPTY, wrapping the launched command in `cmd /d /s /c "..."`. There is currently no setting to change this. I'd like an option (in Settings and/or `config.json`) to select the shell used for spawned terminals - e.g. `powershell.exe` or `pwsh.exe` - instead of the hardcoded `cmd` wrapper.

### Motivation
The embedded terminal has a cosmetic input-rendering glitch under the current `cmd`-wrapped ConPTY path: when typing into the prompt, the **first character shifts left by one column, leaving a gap in the first word**. It's purely visual - pressing `Ctrl+L` (clear/redraw) fixes it, and pressing `Enter` submits the correct text - but it's persistently distracting during normal use.

Empirically, running the same CLI directly under a native PowerShell host does not exhibit this glitch. Being able to point the app's spawned terminals at PowerShell/pwsh would work around it, and is also just a reasonable choice to expose to Windows users who prefer PowerShell.

### Environment
- Munder Difflin v0.4.5 (packaged), Windows 10 Pro (19045)
- Terminals spawned via node-pty ConPTY; Windows command path wrapped in `cmd /d /s /c "..."`

### Proposed behaviour
- A configurable shell for spawned agent terminals, e.g. a `windowsShell` / `terminalShell` key in `config.json` (values like `cmd`, `powershell`, `pwsh`) with a matching Settings control.
- Default remains `cmd` for backward compatibility; users can opt into PowerShell/pwsh.

### Notes
The underlying input is always correct (Enter submits fine), so this reads as an xterm.js/ConPTY redraw issue in the terminal widget rather than a shell-semantics problem. A configurable shell is the pragmatic ask; separately, the first-character column-shift on redraw may be worth fixing directly in the terminal component.

QuintinHumphreys:
## Summary

Fixes #314. Two cosmetic terminal bugs that share a single root cause in the pooled-terminal reattach path.

## Root cause

`PtyTerminalView`'s attach effect runs a "settle cascade" on every mount: a double-`requestAnimationFrame` fit, retries at 60ms and 240ms, and a `document.fonts.ready` re-measure (reapply `fontFamily`/`fontSize`, `clearTextureAtlas`, refit) - each of which also forces `scrollLines(-1)` / `scrollToBottom()` when `scrollToEnd` is true.

Because terminals are pooled (`terminalPool.ts`) and views reparent the same host `div` on every tab switch / fullscreen toggle, this whole cascade re-runs on **every reattach**, even though cell/font metrics were already correct after the terminal's first open:

- **First typed character shifts one column (gap in the first word):** a keystroke landing during the reattach `clearTextureAtlas()` + refit race paints at stale cell metrics. `Ctrl+L` forces a clean repaint (which is why it "fixes" it), and `Enter` is unaffected because the underlying buffer was never wrong - it is purely a paint bug.
- **No usable scrollback:** every reattach yanks scroll to the bottom up to ~4x over a ~240ms window, so a user who scrolls up shortly after switching tabs keeps getting snapped back down - reading as "cannot scroll above the top."

## Fix

- Add `settled: boolean` to `TerminalEntry` (`terminalPool.ts`), initialized `false` on creation.
- In `PtyTerminalView`'s attach effect, latch `isFirstAttach = !entry.settled; entry.settled = true` **before** scheduling async work, and gate the 60ms/240ms retries + `document.fonts.ready` re-measure cascade behind `isFirstAttach`. On reattach the effect keeps only the single double-rAF `fit(scrollToEnd=true)` plus the existing synchronous `scrollToBottom()` - preserving "switching tabs snaps to the latest output" while removing the repeated texture-atlas-clear race and the repeated scroll yank.

## Verification

- `npm run typecheck` (node + web): clean.
- `npm run build` (electron-vite): succeeds.
- Verified by typecheck/build + code reasoning. A live GUI pass on a running build is recommended by a reviewer, as the fix concerns on-screen paint/scroll behavior.

### DIFF
--- a/src/renderer/src/components/PtyTerminalView.tsx
+++ b/src/renderer/src/components/PtyTerminalView.tsx
@@ -147,6 +147,11 @@ export function PtyTerminalView({ ptyId, onStreamData, onUserPrompt, onToggleFul
     entry.term.options.theme = THEMES[ptyThemeRef.current];
     entry.term.options.fontSize = fontSizeRef.current;
     attachTerminal(entry, container);
+    // Only the terminal's true first open needs the settle cascade below (the
+    // delayed refits + web-font re-measure). Latched before scheduling so a
+    // reattach can never replay it — see TerminalEntry.settled.
+    const isFirstAttach = !entry.settled;
+    entry.settled = true;
     entry.onData = (chunk) => onStreamDataRef.current?.(chunk);
     entry.onPrompt = (text) => onUserPromptRef.current?.(text);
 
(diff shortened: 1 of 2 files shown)


QuintinHumphreys:
Closing this for now - I'd like to verify the fix on a local build before putting it up for review. Will reopen once confirmed.

PR #317, opened 2026-08-25
fix: reset usage when a PTY exits

aaroncoville:
## What & why

Tearing down a PTY forgets the agent's circuit-breaker state, but leaves its telemetry sessions in place. Live usage is summed across *every* retained session for an agent id, so a replacement agent that reuses that id starts with its dead predecessor's tokens already counted against it.

A per-agent **token cap can trip on spend that belongs to a session that no longer exists**, throttling or stopping a brand-new agent for something it never did.

This forgets the agent's live telemetry sessions alongside the breaker state. Only the in-memory live accumulation is cleared — the on-disk cost ledger and the transcript fallback are untouched, so lifetime spending is still reported.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before
CleanShot 2026-08-25 at 11 06 10@2x

Post 900 output tokens for a session, let the PTY die, then post 10 for a new session under the **same agent id**. The replacement inherits the dead session's spend:

```
not ok 1 - forgetAgent resets usage before the same agent id is respawned
  expected: 10
  actual: 910
  operator: 'strictEqual'
not ok 2 - forgetAgent keeps usage for other agents
```

### After
CleanShot 2026-08-25 at 11 07 18@2x

The same scenario with the reset in place — the replacement reports only its own usage, and other agents' counters are left alone:

```
# tests 555
# pass 555
# fail 0
```

```
$ npm run typecheck
0 errors

$ npm run build
build_exit=0
```

## Known limitations

Two edges are deliberately left unguarded:

**A metrics batch already in flight when the PTY dies can land after the reset** and recreate the dead session's entries, so one replacement may briefly see the pre-fix behaviour. The window is milliseconds and the next teardown clears it; carrying a tombstone to close it seemed disproportionate to the risk.

**A resumed session keeps its session id**, so until fresh live telemetry arrives its usage is served from the transcript fallback at the session's full historical spend, then drops to post-restart-only once live data lands. That spend is genuinely that session's, and live-over-fallback precedence is the existing contract, so this is arguably correct rather than a defect — but it is visible.

## How I tested it

- OS: macOS (darwin 25.6.0), Munder Difflin 0.4.5
- Steps:
  1. Reproduced against the real `TelemetryCollector`: posted 900 output tokens for a dead session and 10 for a respawn session under one agent id, and read back `output: 910`.
  2. Ran `npm run test:focused` on a branch cut from `main`: 555/555 (baseline on `main` is 552; this adds 3).
  3. Replaced the `forgetAgent` body with a no-op and re-ran — 2 tests fail, including the `10` vs `910` assertion above.
  4. Commented out the call site and re-ran — the wiring test fails, confirming it pins the call rather than matching its own comment text.
  5. `npm run typecheck` — 0 errors. `npm run build` — succeeds.

Scope was checked explicitly: `forgetAgent` runs only inside the `if (agentId)` branch of `teardownPty`, which is the single funnel for every teardown path (natural exit, explicit kill, and the reap/lifecycle sweep), and it clears only the dying agent's entries — a test pins that other agents' usage survives.

## Checklist

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -448,6 +448,8 @@ function teardownPty(id: string): void {
     try { workerWake.forget(agentId, id); } catch { /* best-effort */ }
     // Drop breaker state so a dead agent can't leak/zombie a tripped level.
     try { breaker.forget(agentId); } catch { /* best-effort */ }
+    // A replacement using this id needs a new usage counter, not the dead PTY's.
+    try { telemetry.forgetAgent(agentId); } catch { /* best-effort */ }
     // W1 — kill this agent's proxy-bridge sidecar (qwen), if any, so a dead
     // PTY never leaves an orphan loopback listener. No-op for non-proxy agents.
     try { hive.stopProxyBridge(agentId); } catch (e) { console.error('[hive] stopProxyBridge failed:', e); }
--- a/src/main/telemetry.ts
+++ b/src/main/telemetry.ts
@@ -206,6 +206,25 @@ export class TelemetryCollector {
     return () => this.apiErrorSubs.delete(cb);
   }
 
+  /** Drop an agent's live usage so a respawn starts with a fresh counter.
+   *  Clears the in-memory accumulation only — the on-disk cost ledger and the
+   *  transcript fallback are untouched, so lifetime spending still reports. The
+   *  span ring goes too: it is keyed by agent id, so a replacement would
+   *  otherwise inherit the dead agent's tool waterfall.
+   *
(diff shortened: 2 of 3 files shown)


chaitanyagiri:
Shipped in v0.4.6.

Your commit `2b96d5d0` is in the release: usage resets when a PTY exits, with `test/telemetry-forget-agent.test.cjs` covering it.

Thank you.

PR #320, opened 2026-08-25
fix: refresh MCP defaults after toggling

aaroncoville:
## What & why

Enabling an MCP server writes the grant to disk, but the toggle keeps rendering the value the app was launched with. 

The toggle's on/off state reads `config.mcpDefaults[id].enabled` from the `HarnessConfig` prop, which is loaded once at start-up and never refreshed. The green status line is local state, set after the write resolves to the value that was *requested*, never to a read-back. So the status line reports the attempt and the control reports start-up.

This reads the persisted configuration on mount and again after each write, so the control renders what the configuration says actually landed. A failed read keeps the prop-seeded value rather than guessing, and a failed write leaves the control showing the unchanged grant.

This holds regardless of whether config props are ever broadcast-refreshed (as #286 proposes): a consent control should render the persisted grant read back from disk, not a prop that may or may not have been refreshed. The two changes compose as defense in depth.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before
CleanShot 2026-08-25 at 16 27 55@2x

Enable an MCP server in Connections. The green status line reports it as enabled while the toggle stays **off**, in the same view, at the same moment. Leaving Settings and returning still shows **off**, though the on-disk configuration is correct throughout.

### After
CleanShot 2026-08-25 at 16 29 09@2x

The toggle flips at the same moment the status line appears, and re-entering the section shows the state read back from disk rather than the start-up snapshot.

## Known limitations

The control reflects what the configuration file contains after the write. It does not verify that a server subsequently starts, or that a granted tool is reachable.

### DIFF
--- a/src/renderer/src/components/McpDefaultsSettings.tsx
+++ b/src/renderer/src/components/McpDefaultsSettings.tsx
@@ -1,6 +1,7 @@
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import type { HarnessConfig } from '@/store/config';
 import { MCP_CATALOG, type McpTier } from '@shared/mcpCatalog';
+import { applyToggle, resolveEnabledFor } from './mcpToggleLogic';
 
 export interface McpDefaultsSettingsProps {
   config: HarnessConfig;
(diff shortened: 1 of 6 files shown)


chaitanyagiri:

---
**The comment at the top of `mcp-toggle-component.test.cjs` is the best justification for a test
in any of these 34 PRs:**

```
Comment out the `setMcpDefaults(await applyToggle(...))` line and every one of those tests stays
green while the stale-prop bug returns in full.
```

That is the question worth asking of every test and it is almost never written down. Same for
writing `github-token` out literally rather than importing an id from the catalog, because a test
that asks the implementation which id to check cannot notice the id changing.

**The stale READ is fixed and the stale WRITE is still there.**

```
await deps.updateConfig({ mcpDefaults: { ...(currentDefaults ?? {}), [id]: { enabled: next } } });
```

`currentDefaults` is the component's local state, seeded at mount. The patch replaces
`mcpDefaults` wholesale, which your own test asserts. So if anything changed `mcpDefaults` after
this panel mounted, another window, another instance, an agent, this write puts the mount-time map
back and silently drops that change. It is the same bug the PR is about, on the other verb.
Reading fresh config first and merging the toggle onto THAT before writing is the same shape as
the read fix and closes it.

One note on the mount-only `useEffect`: it fixes the reported case, which is a remount after
switching settings sections. A panel left open while something else writes is still stale. If
`onConfigWritten` can reach the renderer, subscribing would cover both, but the remount case is
the one people hit and this does cover it.

This conflicts as of today, `mergeable: false`.

PR #322, opened 2026-08-25
docs: add a permanent /download fallback page

chaitanyagiri:
## What

A permanent `/download.html` fallback page: a stable, always-current place to install or reinstall Munder Difflin for macOS, Windows, and Linux, **independent of the in-app updater**.

This is the highest-leverage item in the updater work because it is the one recovery path that does not depend on the app updating itself. Every other fix ships *through* the updater we are trying to repair; this one does not, so it works even for a client whose auto-update is stuck.

## How it stays "always correct" with no per-release edits

- Every platform button points at **`releases/latest`** (the release page), not a version-stamped asset URL. Installer filenames carry the version (`...-0.4.5-mac-universal.dmg`), so a direct latest-download link to them would 404 the moment 0.4.6 ships. Pointing at `releases/latest` never goes stale.
- The verify link uses **`releases/latest/download/SHA256SUMS.txt`** — that asset name has no version in it, so this URL is both permanent and always the current release's checksums.
- No version number is hard-coded anywhere on the page.

## Contents

- macOS (universal, signed + notarized), Windows (x64 setup), Linux (x86_64 AppImage), each with the exact filename to pick and a button to the latest release.
- An optional "Verify your download" panel with the `shasum` / `Get-FileHash` commands.
- An "Already installed?" panel explaining the app self-updates and when to use this page instead.

## Scope / safety

- New file only: `docs/download.html`. Nothing else changed — no pricing, no wall data, no `index.html`.
- Reuses the site's existing design tokens, nav, footer, and theme toggle (mirrors `wall.html`). No red, no dashes, tags balanced.
- Not yet linked from the landing nav — its primary use is as a stable URL to hand out (Discord, the app's update-error state, new-machine setup). Say the word if you also want it in the top nav; that is a one-line change to `index.html`.

## ⚠️ Do not merge or deploy without the founder

This is a live public page. Opened for review. Making it live is a merge + Pages deploy, which I am leaving to you / the founder rather than doing unilaterally.

### DIFF
--- a/docs/download.html
+++ b/docs/download.html
@@ -1,0 +1,305 @@
+<!doctype html>
+<html lang="en">
+<head>
+<meta charset="UTF-8" />
+<meta name="viewport" content="width=device-width, initial-scale=1.0" />
+<title>Download Munder Difflin</title>
+<meta name="description" content="Download or reinstall Munder Difflin for macOS, Windows, and Linux. Always the latest release, straight from GitHub, with checksums to verify it." />
+<link rel="canonical" href="https://munderdiffl.in/download.html" />
+<link rel="icon" type="image/svg+xml" href="./logo.svg" />
+<link rel="icon" type="image/png" sizes="32x32" href="./favicon-32.png" />
+<link rel="icon" type="image/png" sizes="512x512" href="./logo.png" />
+<link rel="apple-touch-icon" href="./apple-touch-icon.png" />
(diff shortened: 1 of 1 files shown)


chaitanyagiri:
Shipped in v0.4.6 as `4788ae54`. The permanent /download fallback page is live. Closing, the work is in the release.

PR #323, opened 2026-08-25
fix: release the WebGL context when a terminal's renderer is torn down

aaroncoville:
## What & why

Switching between agent terminals slowly kills the office floor's rendering. Each switch tears down one terminal's WebGL renderer and leases another, but `@xterm/addon-webgl`'s `dispose()` never releases the underlying WebGL2 context (xtermjs/xterm.js#6068, still present in the latest stable). The orphaned contexts accumulate — measured climbing ~2 per switch — until Chromium's live-context cap (32) is reached and the browser starts evicting the oldest context, which on a busy floor is the scene's own context: the floor goes black.

The fix releases the context ourselves at both teardown sites in `terminalPool.ts`: capture the terminal's `webgl2` context before `dispose()`, then call `getExtension('WEBGL_lose_context').loseContext()` after. The lookup is scoped to that terminal's own canvases, and the call is guarded and idempotent — a no-op when the canvas, extension, or live context is absent.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

## Evidence

Instrumented probe (real `@xterm/xterm` + `@xterm/addon-webgl` in Electron, tracking every `webgl2` context) running 30 lease/dispose cycles — the exact per-switch churn the app performs.

### Before
CleanShot 2026-08-25 at 17 12 38@2x

Live contexts climb 3, 5, 7, … 31 and plateau at 32 — Chromium's cap, where eviction of the oldest context begins.

### After
CleanShot 2026-08-25 at 17 13 10@2x

Live contexts stay at 0 across all 30 cycles; nothing accumulates and the cap is never approached.

## How I tested it

- OS: macOS
- Reproduction probe: hidden Electron with remote debugging, instrumented `HTMLCanvasElement.getContext`, 30 renderer lease/dispose cycles; unpatched curve climbs to the 32-context cap, patched curve stays flat at zero.
- `npm run typecheck` (clean), production build (fix present in bundle), `npm run test:focused` (552/552 — renderer GPU behavior is covered by the probe, which the node suite cannot exercise).

### DIFF
--- a/src/renderer/src/components/terminalPool.ts
+++ b/src/renderer/src/components/terminalPool.ts
@@ -586,6 +586,30 @@ function leaseWebglRenderer(entry: TerminalEntry): void {
   }
 }
 
+/** Find the live WebGL2 context an addon renderer is drawing into, so teardown
+ *  can explicitly release it. @xterm/addon-webgl's dispose() tears down its
+ *  renderer and removes its canvases but never calls loseContext(), so the
+ *  underlying context lingers past the addon that owned it (xterm/xterm.js#6068).
+ *  On a floor where agents are switched repeatedly, each switch disposes one
+ *  renderer and leases another, so these orphan contexts pile up until Chromium
+ *  hits its live-context cap (~16) and evicts an older one — often the office
+ *  floor's own context — blacking out a terminal or the scene. Losing the context
+ *  ourselves frees it immediately instead of waiting on GC.
+ *
+ *  Scoped to THIS terminal's element and matched by an active webgl2 context, so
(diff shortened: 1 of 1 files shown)


chaitanyagiri:
Shipped in v0.4.6.

Your commit `e6c936ac` is in: `src/renderer/src/components/terminalPool.ts` releases the WebGL context when a terminal's renderer is torn down.

It landed alongside #248, which retries a floor rebuild that cannot get a context. Yours stops the leak, that one recovers from it.

Thank you.

PR #324, opened 2026-08-25
fix(updater): guard restart re-entry and report a failed quit

chaitanyagiri:
## What

Fixes the recurring restart-to-install wedge: `The command is disabled and cannot be executed`, seen **twice** in `updater.log` (13 Aug and 22 Aug), where the 22 Aug case took six restart clicks before one install finally landed. This is the founder-visible "I click restart and nothing happens" symptom.

## Root cause (two, both here)

1. **Re-entry.** The handler aborted any in-flight restart and fired a fresh `quitAndInstall()` on **every** click. The first call starts Squirrel's install/relaunch, which disables further install commands; a second click's `quitAndInstall()` hits that now-disabled command and Squirrel emits `The command is disabled and cannot be executed`. A guard now refuses a duplicate while a restart is already pending.

2. **Silent failure.** `quitAndInstall` reports a refusal through the autoUpdater **error event**, not a throw, so the handler's `await` never settled: the button spun forever with no way back, which is what drove the repeated clicking in the first place. A failed quit now settles the pending restart (`failPendingRestart`) so the renderer reports the error and restores its notice instead of hanging.

## Change

- `pendingRestart` resolver now carries an outcome `{ ok, error }`, so both the cancel path (`abortPendingRestart`) and the new failure path (`failPendingRestart`) report the truth back to the renderer.
- `abortPendingRestart` keeps its exact signature; `index.ts` `app:cancelClose` still calls it unchanged.
- The autoUpdater `error` handler settles a pending restart (no-op when none is pending, so an ordinary background check error is unaffected).

## Tests

- `test/restart-cancel.test.cjs` extended with the re-entry-guard and settle-on-error contracts. **All 5 assertions green.** `typecheck:node` clean.
- The existing three assertions (await-an-outcome, cancel wiring, renderer restore) still pass unchanged.

## Scope

- One concern only: the restart path. `src/main/updater.ts` + its test. No renderer change (the existing `if (!res.ok)` restore in UpdateToast already recovers now that the handler actually returns on failure).
- The **native menu item** and the **error-state download link** (the other two GO items) are deliberately separate follow-up PRs, per single-concern hygiene.

## Not for merge yet

Part of the 0.4.6 updater set. Opened for review; merge/release is the founder's call.

### DIFF
--- a/src/main/updater.ts
+++ b/src/main/updater.ts
@@ -71,7 +71,7 @@ let lastStatus: UpdateStatus | null = null;
  * left to tell), or the user cancels and `abortPendingRestart()` settles it and
  * the handler reports the truth.
  */
-let pendingRestart: (() => void) | null = null;
+let pendingRestart: ((outcome: { ok: boolean; error?: string }) => void) | null = null;
 
 /**
  * The user backed out of the quit that a restart-to-install asked for.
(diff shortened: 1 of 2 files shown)


chaitanyagiri:
Shipped in v0.4.6 as `942fb741`. Restart re-entry is guarded and a failed quit is reported. Part of the updater repair set that this release proves end to end. Closing.

PR #326, opened 2026-08-25
feat(updater): badge acknowledges a check and drives the real auto-update

chaitanyagiri:
## What

One coherent badge behaviour for 0.4.6, two parts:

### 1. A successful check is now visible (the acknowledgement)

On the latest release a manual check succeeded, found nothing, and settled silently to a grey `latest` chip. A working check and a dead click were indistinguishable, so the button read as broken. Now, after a **manual** no-update check, the badge flashes a positive popover ("You are on the latest version. vX is the newest release. Checked just now.") that auto-dismisses. It fires only for the manual no-update result (read back from the settled status), not for background 6h checks, and is `role="status" aria-live="polite"`.

### 2. The badge does the real auto-update (Option B)

The badge only ever offered a **manual** download (a DMG + a drag-to-Applications card), never the native download/restart the Settings pane runs. So when 0.4.6 lands, clicking the badge would hand the user a DMG and they would rightly say auto-update still does not work. Now, when the native updater has staged an update, the badge drives the same path Settings does:

- `available` → action `download` ("vX · update"): kick the native download.
- `downloaded` → action `restart` ("vX · restart"): quit and install.
- `available-manual` → action `manual` ("vX · download"): the notify-only fallback ONLY, where the native updater could not fetch it. The drag-to-Applications hover card now shows only in this case.

This is the last code change before the 0.4.6 freeze, and it is the behaviour the rc rehearsal exists to exercise; rehearsing without it would test the old badge, a build that never ships.

## Tested vs rc-only (the reporting standard)

- **Verified now**, without a real release: `describeUpdate`'s mapping for every state (unit tests, all 17 green: `available`→download, `downloaded`→restart, `available-manual`→manual), the badge's click-wiring, the acknowledgement (2 tests green), and `typecheck:web` clean. In dev these states can also be driven visually through the existing `update:simulate` handler.
- **Only a real signed release can exercise** (the rc, `0.4.6-rc.1` → `0.4.7-rc.1`): `downloadUpdate`, `quitAndInstall`, and the Squirrel install itself, i.e. that clicking `restart` actually installs and relaunches. The unit tests prove the badge *asks* for the right action in each state; they cannot prove the native install, because 0.4.5 is the latest release and no update exists to drive it. Do not read the green here as covering the install path.

## Scope

`src/shared/updateState.ts`, `src/renderer/src/components/UpdateBadge.tsx`, and `test/update-state.test.cjs`. Part of the frozen 0.4.6 set (#322, #324, #325, #326). The error-state download link and the menu item were cut to 0.5.0. Founder's merge.

### DIFF
--- a/src/renderer/src/components/UpdateBadge.tsx
+++ b/src/renderer/src/components/UpdateBadge.tsx
@@ -25,6 +25,11 @@ export function UpdateBadge() {
   /** The version whose download was just started, for the "now replace the
    *  app" notice. Local state: it is a one-off explanation, not an update state. */
   const [started, setStarted] = useState<string | null>(null);
+  /** Brief, positive "checked, you are current" flash after a MANUAL check that
+   *  found no update. Without it a successful check settles silently back to the
+   *  grey "latest" chip, which is indistinguishable from a click that did
+   *  nothing, and that is exactly why the badge read as broken. */
+  const [checkedOk, setCheckedOk] = useState(false);
 
   useEffect(() => {
     // Subscribe first, then pull — main may have emitted before this window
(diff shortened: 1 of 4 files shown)


chaitanyagiri:
Shipped in v0.4.6 as `1ca01b09` and `2096c1d3`. The badge acknowledges a check and drives the real auto-update rather than a manual download. This is the change the release notes lead on with "updates install themselves". Closing.

PR #329, opened 2026-08-25
feat(telemetry): instrument the activation funnel (launch → first agent)

chaitanyagiri:
## Why

The 6 existing events are fine; the problem is **coverage, not volume**. There is zero instrumentation between `app_launched` and `agent_spawned`, which is exactly where activation happens and where people fall out, and **a failed spawn currently sends nothing** — so Mode 2 (an engine CLI with no unattended install path) and "the user changed their mind" are the same non-event. Jim's numbers make this the priority: day-one return is 9.2% with zero agents and 39.9% with one, and the path to that first agent was dark.

Founder brief: the most detailed **non-content** metrics we can get, with the policy moved to match the code in the same change.

## The five new events, and the question each answers

| Event | Props | The question it exists to answer |
| --- | --- | --- |
| `onboarding_completed` | `provider` | What share of installs finish setup, and which engine do they choose? |
| `agent_spawn_attempted` | `provider` | How many people try to start an agent? `attempted − spawned` is the activation fallout. |
| `agent_spawn_failed` | `provider`, `reason` | **Why** don't agents start? `cli_missing` (Mode 2) vs `cwd_missing` vs `already_running` vs `spawn_error`. |
| `agent_install_started` | `provider`, `rung` | How often does the auto-installer run, via which rung (`npm`/`node-then-npm`/`native`)? |
| `agent_install_finished` | `provider`, `rung`, `outcome` | Does the auto-installer actually complete? `outcome: install_failed` is the Linux-cannot-finish signal that used to be silent. |

The funnel now reads end to end:
```
app_launched → onboarding_completed → agent_spawn_attempted
   ├─ agent_spawned                         binary present, real agent
   ├─ agent_spawn_failed (reason)           cwd/already/error, or cli_missing = Mode 2
   └─ agent_install_started (rung)          binary missing, installer running
        └─ agent_install_finished (outcome)
             ├─ agent_launched → (relaunch) agent_spawned
             └─ install_failed              Mode 2: installer ran, could not complete
```

## Wiring (four main-process sites, no renderer changes)

- `spawnAgentCore` top → `agent_spawn_attempted`, **gated on `!noAutoInstall`** so the missing-CLI relaunch (the only re-entry) never double-counts one attempt.
- The real `ptyManager.spawn` → `agent_spawned` on ok, `agent_spawn_failed` otherwise.
- The missing-CLI branch → `agent_install_started` (installer running), `agent_spawn_failed:cli_missing` (manual rung = Mode 2), or `agent_spawn_failed:<reason>` (install PTY failed to spawn).
- The install-PTY exit handler → `agent_install_finished` with `agent_launched` (clean exit, relaunching) or `install_failed` (non-zero exit).
- `config:update` on the `onboardingComplete` false→true transition → `onboarding_completed`.

## Docs move with the code (the #238 lesson: change the promise, then collect)

`TELEMETRY.md` and `docs/privacy.html` both gain the five events. `privacy.html` **also picks up `update_applied`**, which it was missing — closing a pre-existing doc/code drift so the page mirrors the allowlist exactly.

## Anonymity unchanged

Every added property is a closed CLI name or a fixed enum; the hard allowlist in `analytics.ts` drops anything else (test proves an unknown key is dropped); events stay `$process_person_profile: false`. Opt-out (`telemetryEnabled` + `DO_NOT_TRACK` + build-from-source) untouched.

## The one decision NOT taken here: CONTENT

Content (prompts, agent output, file paths, repo names, hostnames) is deliberately **excluded**. Users type API keys and paste proprietary code into these terminals, so collecting content means holding other people's secrets — a breach-liability posture, not a policy toggle, and GDPR/DPDP-loaded (~52% India traffic). That is the founder's call, raised with cost in `hive/shared/eng/TELEMETRY-ACTIVATION-FUNNEL-kevin.md`, and this PR ships none of it either way.

## Verified / rc-only

- **Verified now:** typecheck (node + web) clean; the allowlist accepts the five events and drops an un-allowlisted key; the events stay anonymous; and source-string assertions pin all four wiring sites (8/8 in `test/telemetry-funnel.test.cjs`, and the 43 existing analytics tests still pass).
- **Not exercised here (needs a packaged build with a live PostHog key):** that PostHog actually receives the events; the real missing-CLI install path end to end on a bare machine.

## Boundaries

Target **0.4.7**. 0.4.6 stays frozen — this touches nothing in it and does not touch the updater. Non-content only; docs in lockstep with the allowlist.

### DIFF
--- a/TELEMETRY.md
+++ b/TELEMETRY.md
@@ -1,8 +1,9 @@
 # Telemetry
 
 Munder Difflin collects a small set of **anonymous** usage events so we can
-understand adoption (how many people launch the app, which features get used)
-and make the product better. This document is the complete, authoritative
+understand adoption (how many people launch the app, whether they get a first
+agent running, which features get used) and make the product better. This
+document is the complete, authoritative
 contract: **if an event or property is not listed here, the app does not send
 it.** The implementation lives in [`src/main/analytics.ts`](src/main/analytics.ts)
 and enforces this list as a hard allowlist — the code and this file are kept in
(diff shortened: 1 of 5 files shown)


chaitanyagiri:
Shipped in v0.4.6 as `5e2e40b6`. The activation funnel from launch to first agent is instrumented. Note this one went in as a direct commit with no merge commit naming the PR. Closing.

PR #330, opened 2026-08-26
fix(codex): use the verified Codex default

aaroncoville:
## What & why

The Codex preset recommends `gpt-5-codex`, but a fresh `model/list` probe under ChatGPT login does not offer that slug. The CLI supports legacy-model access, so list absence alone does not establish rejection; separately, the recorded ChatGPT-login first-message error rejects `gpt-5-codex`. A new orchestrator can therefore fail after launch.

This updates the preset to `gpt-5.6-sol`, the same probe's default, and refreshes the curated picker and add-agent example to its observed catalog. The regression test independently pins the recommendation and verifies that the picker offers it, so a future catalog change is visible in tests rather than silently leaving a stale default.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before
CleanShot 2026-08-25 at 15 53 58@2x

The preset and add-agent example use `gpt-5-codex`. It is not offered by the fresh ChatGPT-login `model/list` output, and the recorded first-message error rejects it under that login.

### After
CleanShot 2026-08-25 at 15 54 51@2x

The preset, picker, and example use the observed default `gpt-5.6-sol`; the picker now includes the complete observed catalog.

## How I tested it

- Ran the zero-spend `model/list` probe under ChatGPT login on 2026-08-25.
- `npm run typecheck`
- `npm run test:focused` (553 passing)
- `npm run build`

## Known limitations

The catalog result is scoped to a ChatGPT login on 2026-08-25. Codex's model catalog is provider-controlled and can include legacy models outside the picker; refresh `model/list` before changing these curated recommendations.

### DIFF
--- a/src/renderer/src/components/AddAgentModal.tsx
+++ b/src/renderer/src/components/AddAgentModal.tsx
@@ -99,7 +99,7 @@ Return EXACTLY this shape (omit optional fields you don't need; keep the spec st
 }
 
 Rules:
-- "provider" MUST be one of: cursor | claude | codex | antigravity. "model" must be a real model id for that provider (e.g. gpt-5.6-luna-high, claude-opus-4-8[1m], gpt-5-codex, "Gemini 3.1 Pro (High)").
+- "provider" MUST be one of: cursor | claude | codex | antigravity. "model" must be a real model id for that provider (e.g. gpt-5.6-luna-high, claude-opus-4-8[1m], gpt-5.6-sol, "Gemini 3.1 Pro (High)").
 - Do NOT include shell commands or any flags beyond these fields.
 - Make "description" + "goal" concrete enough that the agent knows exactly what to do on its first turn.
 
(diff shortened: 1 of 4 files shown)


chaitanyagiri:

---
The lockstep test is the good part. `codex recommends the verified ChatGPT default` asserts both
the value and that the picker actually offers it, so the comment in `agentProvider.ts` saying
`Keep this in lockstep with CODEX_MODELS` is now enforced rather than hoped for. The previous
value `gpt-5-codex` was never in `CODEX_MODELS` at all, so an agent created off the recommended
path stored a model its own picker did not list. That is worth saying in the PR body, because it
makes this a correctness fix rather than a version bump.

The question I cannot answer from the diff: what happens to agents already persisted with
`gpt-5-codex`. Adding four models and changing the recommendation does not migrate anything. If
the picker renders a stored value that is not in its option list, does it show blank, fall back
to CLI default, or keep the stored id? Whichever it is, someone with an existing codex agent will
hit it on first open after upgrading. I do not know what that code does, so this is a question
and not a finding.

This conflicts as of today, the API reports `mergeable: false`.

PR #331, opened 2026-08-26
Per-agent Talk: live voice sessions for worker agents

aarjav:
## What this adds

**Per-agent Talk** — the live "Talk" voice button, until now available only on the orchestrator (god/Michael) card, now works on **worker agent cards** too. Clicking Talk on a worker card opens an OpenAI-realtime, speech-to-speech session that adopts **that agent's** identity: its name, its registry role, and a digest of its own `memory.md`.

- **Distinct voice per role** — worker sessions use `marin`; god keeps `cedar`, so you can hear which agent picked up.
- **Self-scoped, read-only authority** — worker voice sessions get tools scoped to *themselves* (`ask_my_session`, `my_latest_reply`, `check_my_session`, `read_my_memory`, `get_my_tasks`). They **cannot** hire, kill, pause, archive, dispatch other agents, or change settings — that stays god's console.
- **god's Talk path is byte-identical** — `session.ts` only reaches into the new per-agent module for non-god targets.

## How the voice ↔ session relay works

The voice session is the phone line to a working agent, not a separate brain. `ask_my_session` relays what the operator says **into the agent's live Claude Code session** over the two paths the app already trusts — the renderer message queue for an idle agent, a steer note for a busy one (chosen deterministically in code, never by the model) — then reads the session's reply back and speaks it.

## Bug caught in live QA and fixed here (A3)

The first live test surfaced a real gap: the operator's question reached the agent's session (visible in the terminal), but the session's **answer never returned to the voice** — the voice layer polled `recentAssistantText` / `recentTextTs`, but **nothing in the live pipeline ever wrote those fields** (only mock data did, which is why unit tests were green while the live path was dead).

The fix (commit `2522336`) adds the missing producer:
- `main/transcript.ts` surfaces each agent's **latest assistant message** from its transcript;
- exposed over IPC (`index.ts` / `preload`);
- written into the store from `useHive`'s per-agent poll (~3s), so `recentAssistantText` now has a **live writer**, not just mock;
- `RELAY_WAIT_MS` raised to 90s (real Claude Code turns routinely exceed 25s);
- a **non-mock** transcript → store → relay end-to-end test added so this can't regress green again.

### Before
Before — worker cards (Jim, Pam) have no Talk button; only the orchestrator card does

_Pre-feature build: the **Talk** button appears **only on the orchestrator (BOSS / god) card**. The worker cards — **Jim** and **Pam** — have no Talk button, so there is no way to hold a live voice conversation with a worker agent._

### After
After — worker cards Jim and Pam now show the Talk button; a live worker voice session is active

_This build: worker cards — **Jim** and **Pam** — now show the **Talk** button, and a live worker voice session is active (god shown "listening"). Clicking Talk on a worker and asking "what are you working on?" returns the agent's answer **spoken back in its own voice** (`marin`) within ~90s, sourced from its live session. god's Talk is unchanged (`cedar`)._

## Testing

- `559 / 559` tests pass, including the new non-mock live-path test.
- Operator ran the live 4-step voice QA (A1 god unchanged/cedar; A2 worker greets as itself in marin; A3 ask-your-work answer spoken back; A4 mid-session card switch) — **passing** after the A3 fix.

## Known issues

- **Worker voices are not yet distinct from one another.** All worker agents currently share a single voice (`marin`) — so Jim and Pam sound **identical to each other**, though both remain clearly distinct from Michael/god (`cedar`). This is confirmed in code: `AGENT_VOICE` is one shared constant rather than a per-agent assignment. Per-worker voice differentiation is a planned follow-up, not part of this PR.

## Commits

- `bd7b1f2` — feat(realtime): per-agent Talk — voice sessions for worker cards
- `2522336` — fix(realtime): per-agent Talk A3 — write the session's reply back to the store

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -32,7 +32,7 @@ import { MemoryManager } from './memory';
 import { KnowledgeManager } from './knowledge';
 import { MemoryReflector, type ReflectSettings } from './reflect';
 import { PersistStore } from './db';
-import { readAgentUsage, readContextTokens, seedSessionTranscript, resolveSessionCwd } from './transcript';
+import { readAgentUsage, readContextTokens, readLatestAssistantText, seedSessionTranscript, resolveSessionCwd } from './transcript';
 import { listIssues, listCIRuns } from './github';
 import { SlackWebhookServer, SlackReplyServer, postSlackReply, type SlackEventFile } from './slack';
 import {
@@ -3711,6 +3711,19 @@ ipcMain.handle('hive:agentContext', (_evt, agentId: unknown) => {
   const tp = hookServer.transcriptPath(agentId);
   if (!tp) return null;
   return readContextTokens(tp) ?? 0;
+});
+
+// The latest thing an agent's session actually SAID — the live producer behind
+// per-agent Talk's relay (renderer writes it into the store as
+// recentAssistantText/recentTextTs; agentVoice's ask_my_session waits on it).
+// Same transcript-path source as hive:agentContext, so it is available for
+// exactly the agents whose hooks have fired. Null until then, or until the
+// session has spoken.
+ipcMain.handle('hive:agentLatestText', (_evt, agentId: unknown) => {
+  if (typeof agentId !== 'string') return null;
+  const tp = hookServer.transcriptPath(agentId);
+  if (!tp) return null;
+  return readLatestAssistantText(tp);
 });
 
 // A consolidated, NON-SENSITIVE per-agent directory for the voice read-layer
--- a/src/main/transcript.ts
+++ b/src/main/transcript.ts
@@ -339,3 +339,97 @@ export function readContextTokens(transcriptPath: string): number | null {
     return null;
   }
 }
+
+/** The latest thing an agent's live session actually SAID, plus when it said it.
+ *  `ts` is epoch ms (the record's own `timestamp`, or the transcript's mtime when
+ *  a record carries none) — the renderer stores it as `recentTextTs`, which is
+ *  what the per-agent Talk relay watches to know the session has answered. */
+export interface LatestAssistantText {
+  text: string;
+  ts: number;
+}
+
+/** Tail window for the text scan. Same reasoning as CONTEXT_TAIL_BYTES: the
+ *  record we want is always at the END of an append-only transcript, and these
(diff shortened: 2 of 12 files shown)


aarjav:
Before and after screenshots. If a video is ABSOLUTELY needed I can upload one.

talk-evidence-AFTER
talk-evidence-BEFORE
AFTER-2Screenshot 2026-08-25 at 7 20 07 PM
AFTER4-Screenshot 2026-08-25 at 7 18 05 PM
AFTER-3Screenshot 2026-08-25 at 7 17 25 PM

chaitanyagiri:

---
**Scope: I read `src/main/transcript.ts` and the preload contract. I did not review
`renderer/src/realtime/session.ts` or `agentVoice.ts`, which is where most of this PR lives. Treat
everything below as a comment on the producer half only.**

`readLatestAssistantText` is careful in the places that matter. Scanning backwards from a 256 KB
tail, tolerating a first line cut mid-record, skipping `isSidechain` records because subagent
chatter is not the session speaking, and dropping thinking and `tool_use` blocks so only what the
agent actually said is returned. The size plus mtime memo means the 3 second poll is a `stat` in
the steady state. All of that is right.

**Two questions, and the second is the one I would want answered in the PR body.**

**What bounds the length?** `assistantText` joins every text block of the record with no cap, and
the relay speaks the result. An agent that returns a long answer, and they routinely do, produces
a voice that talks for minutes. If a cap exists downstream, say where.

**Where does the user agree to this text leaving the machine?** This returns the agent's own words
to the renderer so they can be spoken, and if that goes through a hosted realtime voice API then
the agent's output is transmitted to a third party. Agent output is exactly what this repo is
careful about elsewhere: #309 in this same batch gitignores the crash dumps because PTY output
`can include tokens, paths and prompt fragments`, and that is a local file. **I am asking rather
than asserting, because `RealtimeMichaelToggle` already exists**, so this may simply be widening a
path the user has already opted into for Michael, in which case saying so in the PR closes it. If
it is not, then per-agent Talk sends every worker's replies somewhere Michael's setting never
covered.

Housekeeping: this conflicts today, `mergeable: false`. It is the largest PR in the batch at
roughly 1,070 added lines, and it does NOT collide with #335 despite both touching
`transcript.ts` — #335 rewrites lines 155 to 320 and this appends past 339.

Issue #296, opened 2026-08-24
Isolated pi agent dir has no models.json, so the first session dies ~415ms after every hire


chaitanyagiri:
Reported in Discord `#help` by **dirkolbrich** on 25 Aug 2026. He asked his own orchestrator to investigate why agent restore was failing for him and posted the findings. This is the first of four, filed separately because they are four separate fixes. He listed them in priority order and this was his first.

> The first session died ~415 ms after each hire. The fresh isolated `PI_CODING_AGENT_DIR` had no `models.json`, so pi couldn't resolve the assigned omlx model and exited. The log shows all three hired at 11:59/12:03/12:07 and archived ~0.4 s later, and the pi sessions folders are empty.

**What happens:** hiring with isolation creates a fresh agent directory that contains no `models.json`. pi cannot resolve the assigned model, exits immediately, and the agent is archived under half a second after being hired. The pi session folders are left empty.

**The user visible result** is an agent that appears to be hired and is dead before you look at it.

**His proposed fix:**
> Seed the isolated pi-agent dir with `~/.pi/agent/models.json` on first spawn (and/or allow choosing the local LLM at hire)

**His workaround today** is copying `models.json` out of his local `~/.pi/agent/` folder by hand. He confirmed that after doing this plus a manual `roster.json` fix, all four isolated directories had the model available and the model half of the problem was cured. Restore still failed afterwards for a different reason, which is filed separately.

He classed this one as a missing feature rather than a bug: there is no local LLM choice at hire, and nothing copies the existing config into the isolated directory.

Filed from Discord triage. Reporter's own diagnosis, not independently reproduced.

chaitanyagiri:
**Restore family.** These four came from one investigation by dirkolbrich and are best read together, in this order:

- #296 isolated pi agent dir has no `models.json`, first session dies on hire
- #297 PTY exit deletes the named agent's worktree
- #298 worktree origin never persisted, restore falls back to the dead path
- #299 pi `session_id` never forwarded, restored agent starts fresh

#297 destroys the worktree, #298 is why nothing can find it afterwards, and #299 means even a fixed restore comes back with no memory. #296 is separate from the chain but is what kills the agent at hire time.

chaitanyagiri:
_Posted from @chaitanyagiri by an agent on the Munder Difflin hive floor. First person statements below are the agent's. Replies here are read._

---

Linking this to work that already exists, because neither side references the other.

[...]

PR #332, opened 2026-08-26
fix(pi): copy models.json and models-store.json to .pi-agent directory

reschandreas:
## What & why

I use pi with local models and wanted to test munder difflin, however I noticed that in the agent directories my custom config with my models was missing and I could not use them, see the attached gif that it does not see any models.
With the two json files we bring the user's desired config into the agent's directory so that the models work.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

## Evidence
The agents using pi can now use the locally configured models, evidence is the absence of the warning.

### Before
https://github.com/user-attachments/assets/87bfaad0-2482-408d-8b4c-a2d8a9eb6d60

Note the warning that no models are available.
### After
https://github.com/user-attachments/assets/a00472d6-21a7-434e-bc0a-4a27975b7af1

The warning is gone.

## How I tested it

- OS: macOS 26.6.2 (25G83)
- Steps:
  - use pi with locally configured models (I use https://github.com/drumih/turbo-fieldfare )
  - create a new agent with pi and default config
  - observe that the newly spaned agent can now use the models which it could not see before the change.

## Discord (optional)

Discord:

## Checklist

      spacing, or fonts.
      `ATTRIBUTION.md`.

### DIFF
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -2170,12 +2170,16 @@ export class HiveManager {
       // Pi ignores it). Kept minimal and hive-authored.
       const manifest = { name: 'munder-hive-bridge', version: '0.3.1', main: 'extensions/hive-bridge.js', auto: true };
       writeFileSync(join(home, 'extensions.json'), JSON.stringify(manifest, null, 2), 'utf8');
-      // we need to copy models.json and models-store.json from the user's ~/.pi to the per-agent dir so Pi can load its models
+
       const userPiDir = join(homedir(), '.pi', 'agent');
-      const userModelsPath = join(userPiDir, 'models.json');
-      writeFileSync(join(home, 'models.json'), existsSync(userModelsPath) ? readFileSync(userModelsPath, 'utf8') : '{}', 'utf8');
-      const modelsStorePath = join(userPiDir, 'models-store.json');
-      writeFileSync(join(home, 'models-store.json'), existsSync(modelsStorePath) ? readFileSync(modelsStorePath, 'utf8') : '{}', 'utf8');
+      for (const fileName of ['models.json', 'models-store.json'] as const) {
+        try {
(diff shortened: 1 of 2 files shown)


chaitanyagiri:

---
The three tests cover the cases that matter, including the asymmetric one where `models.json`
exists and `models-store.json` does not.

**The `{}` fallback is the thing I would want checked against pi itself.** When the user has no
`~/.pi/agent/models.json`, this writes an explicit `{}` into the per agent dir. Before this
change there was no file there at all. Those are not the same input: a CLI that finds no config
usually falls back to its own defaults, and a CLI that finds an empty object may read it as "the
user configured zero models". The tests assert `{}` is written, which is what the code does, and
nothing establishes that pi is happy with it. If you saw pi start cleanly with an empty
`models.json`, say so in the PR. If not, not writing the file at all when the source is missing
is the safer shape.

**`existsSync` then `readFileSync` on the next line**, twice. If the file disappears in between,
the read throws, and since this is the last block in the `try`, the whole thing lands in
`catch (e) { console.error('[hive] installPiHooks failed:', e) }`. Narrow window and it does log,
so this is minor, but `try { readFileSync } catch { '{}' }` per file removes it and is the same
number of lines.

**The comment.** `// we need to copy models.json and models-store.json from the user's ~/.pi to
the per-agent dir so Pi can load its models` says what the next line already says. Every other
comment in this file says what breaks without it, and #296 has the fact: the first session dies
about 415ms after every hire. That is the sentence worth having there, because it is the thing a
future reader needs in order not to delete this as redundant.

reschandreas:
hi @chaitanyagiri thanks for the feedback, I changed the PR to only create the files in the agent's directory if the user has them.

PR #333, opened 2026-08-26
closed: N/A

Monilnarang:
N/A


Monilnarang:
@chaitanyagiri This has no visual UI change; it only adjusts `cbcode` provider detection and terminal startup arguments. Could you please apply the `no-visual-change` label to waive the required screenshot check?

chaitanyagiri:
Thanks for this, and for the write-up. The three failure modes are laid out clearly and the
table made the change easy to follow.

First, an apology about CI. The `Before / after evidence` failure is our bug, not yours. The
workflow's section matcher uses a multiline `$` in a lazy lookahead, so the text it captures
under each heading is always empty and it reports both as missing no matter what is attached.
It is currently failing on our own PRs too. We are fixing it. Separately, that check only
accepts an image or a video, which is a poor fit for a CLI-only change like this one, so we
will sort out the right path for your PR rather than have you work around it.

On the change itself: I re-checked the isolation claim and it holds, and it is stronger than
you wrote it. For any non-cbcode binary `normalizePermissionMode` returns the same array
reference, so a stock `claude` argv is not merely byte-identical, it is untouched. I also
checked the adversarial case of a real `claude` living in a directory named `cbcode`, and it
correctly does not trigger. The deny-list is the right call and it is complete: the
remote-control opener has exactly one call site in the tree and you patched it. On permissions,
I probed the rewrite across a set of argv shapes and it is one-directional, so it can never
introduce `bypassPermissions` where it was absent, and it correctly leaves a bare
`bypassPermissions` token and `--disallowedTools bypassPermissions` alone.

One thing we do need before merging, and it is the only real blocker.

Mapping to the Claude family gives an agent two separate injections, and they ride on
different flags. `--append-system-prompt` carries the hive protocol text, which your
PASSTHROUGH transcript proves works. `--settings` carries the hook bridge: a generated
settings file wiring SessionStart, Stop, PreToolUse, PostToolUse and friends to a shim that
posts to our `hooks.sock`. That bridge is what actually makes an agent a participant, giving
live status on the floor, Stop-to-inbox drain, session ids and the context gauge. We have no
evidence for it yet.

Your step 6 does not settle it, because the protocol text is itself what tells an agent to
read its inbox at startup. A cbcode that accepted `--settings` and silently ignored it would
look identical from the terminal. That matters more than usual here, because the mapping also
turns OFF our no-hook fallback: `usesHookStandingGoal()` returns true for any `hiveAware`
provider, so an agent's standing goal stops being prepended to deliveries on the assumption
that hooks will carry it. If the hooks are inert, the goal is delivered by neither path and
nothing errors.

We cannot test this ourselves; there is no cbcode on any machine we have. Could you run this
and paste the result?

```bash
cat > /tmp/cb-probe.json <<'JSON'
{ "hooks": {
    "SessionStart": [{ "hooks": [{ "type": "command", "command": "touch /tmp/cb-hook-fired" }] }],
    "Stop":         [{ "hooks": [{ "type": "command", "command": "touch /tmp/cb-hook-fired" }] }] } }
JSON
rm -f /tmp/cb-hook-fired
cbcode --permission-mode auto --settings /tmp/cb-probe.json -p "say hi"; echo "exit=$?"
ls -l /tmp/cb-hook-fired 2>/dev/null && echo "BRIDGE OK" || echo "BRIDGE DEAD"
```

If the file appears, your claim stands as written and we will merge. If `--settings` is
refused, or accepted and ignored, the PR is still worth landing, because running without a
bridge beats dying at launch, but we would want to say so explicitly and give cbcode its own
bridge tier rather than have it inherit one that is not there.

Two smaller things:

1. We still send a Claude-only flag you have not carved out. Every named hive agent gets
   `--remote-control-session-name-prefix <name>` from `spawnAgentCore`. If cbcode drops Remote
   Control the way it drops the slash command, that is the same class of failure you fixed for
   `bypassPermissions`. Your successful app run suggests it is tolerated, but could you confirm
   with `cbcode --permission-mode auto --remote-control-session-name-prefix probe -p "say hi"; echo $?`

2. Detection is by binary leaf name, which means an install aliased as `claude`, a fairly
   natural shape on a managed laptop, is not recognised and still gets `bypassPermissions`.
   Worth adding to your limitations list. Setting the provider explicitly is the workaround.

On the `--agent` limitation: we are happy to ship it as is. Handing Claude flags to a Codex
session fails loudly and immediately rather than silently misbehaving, so it is the same
visible failure this PR fixes for the default case. Parsing `--agent` as a follow-up is the
right shape.

Once the hook probe comes back we will merge. We will land it on `main` for the release after
next rather than the one in flight, since it touches the spawn path every agent goes through
and that branch is mid-rehearsal.

PR #335, opened 2026-08-26
fix(usage): cache transcript usage per file, not per (file, sessionId)

### LINKED ISSUE #334: Usage cache keyed by (file, sessionId) causes O(agents × files) reparsing — CPU pegs / UI freezes under active multi-agent load
## Summary

`readAgentUsage`'s per-file cache is keyed by `dir|file|sessionId`, not just `dir|file`. With N agents sharing a project directory, one agent's transcript growing causes **every other agent** to independently re-read and re-parse the same new bytes on their own next usage check — even though the session filter means only one of them actually needed the data. This scales with active agent count, so the app is fine at idle and increasingly CPU-pegged the more agents are simultaneously doing real work.

On a 10-agent floor this repros as the Electron main process pegging ~100% CPU (confirmed via `sample`/CPU profile — >50% of self-time in `parseUsageLines`) and the whole UI freezing for as long as agents stay active.

## Root cause

`src/main/transcript.ts`:

```ts
function readFileUsage(dir: string, file: string, sessionId: string | undefined): FileUsageEntry | null {
  const key = `${dir}|${file}|${sessionId ?? '*'}`;
  ...
```

Because the cache key includes `sessionId`, the incremental byte-offset parse of a file is duplicated once per distinct sessionId that ever queries it — not shared across agents. `parseUsageLines` also does the session filtering *after* parsing every line, so the wasted work isn't even skipped early.

## Reproduction (isolated, no real transcripts needed)

10 simulated agents sharing one project dir, 30 rounds of "everyone active" (each round: every agent's file gets one new record appended, then every agent's own `readAgentUsage(cwd, {sessionId})` fires — the real ~30s usage-beat access pattern):

| | wall-clock | `JSON.parse()` calls |
|---|---|---|
| current `main` | 180.6ms | **3000** |
| with per-file caching | 65.6ms | **300** |

Exactly 10x fewer parse calls, matching the 10 simulated agents — confirms the mechanism precisely, not just a vague slowdown.

## Fix

PR incoming: cache the parsed result per **file** only, storing both an unfiltered total and a per-`sessionId` breakdown from the same single parse pass. A filtered `readAgentUsage(cwd, {sessionId})` call becomes a Map lookup against already-parsed data instead of triggering its own read+parse. Existing `test/transcript-usage.test.cjs` suite (10 cases, including the exact session-filter-correctness contract) passes unchanged.

savvaskoualis:
Fixes #334.

## The bug

`readFileUsage`'s cache was keyed by `dir|file|sessionId`, not just `dir|file`. Every agent sharing a project directory got its own independent cache entry for the *same physical file* — so when one agent's transcript grew, every other agent's next usage check re-read and re-parsed those same new bytes too, just to find nothing matching their own session. This scales with active agent count: fine at idle, increasingly CPU-heavy the more agents are simultaneously working, until the Electron main process pegs and the UI freezes.

Confirmed via a live CPU profile on a 10-agent floor — `parseUsageLines` was >50% of main-thread self-time while agents were active.

## The fix

Cache per **file**, not per `(file, sessionId)`. `parseUsageLines` now writes each record into both an unfiltered total (`all`) and a `perSession` map in one pass over the file's new bytes. A filtered `readAgentUsage(cwd, {sessionId})` call reads off the already-computed bucket — no extra file I/O or JSON parsing, regardless of how many agents ask.

### Before
https://raw.githubusercontent.com/savvaskoualis/munder-difflin/fix/usage-cache-cross-agent-amplification/docs/evidence/usage-cache-fix-before.mp4

Isolated dummy workload (60 simulated agents sharing a throwaway project dir, 150 rounds of "everyone active" — no real transcripts, no real hive involved), running the pre-fix code. **49.8s wall-clock, 40,770,000 `JSON.parse()` calls, ~89% CPU sustained** for the entire run.

### After
https://raw.githubusercontent.com/savvaskoualis/munder-difflin/fix/usage-cache-cross-agent-amplification/docs/evidence/usage-cache-fix-after.mp4

Identical workload, identical machine, this PR's code. **3.6s wall-clock, 9,000 `JSON.parse()` calls.** Same inputs, ~14x faster, ~4530x fewer redundant parses.

## Verification

**Existing correctness suite**, unchanged, all passing (includes the exact session-filter-correctness contract this touches):

```
$ node test/transcript-usage.test.cjs
  ok  sums usage across records and files
  ok  appended records are picked up (incremental tail parse)
  ok  repeat reads with no change are stable
  ok  a partial trailing line is not counted until completed
  ok  a rewritten (shrunk) file is re-parsed from scratch
  ok  a deleted file drops out of the totals
  ok  sessionId filter sums only that session, incrementally too
  ok  malformed lines and non-assistant records are skipped
  ok  model is the last one seen (per current semantics)
  ok  missing project dir yields zeros
```

Benchmark script included as `test/usage-cache-benchmark.manual.cjs` (env-configurable via `BENCH_AGENTS`/`BENCH_ROUNDS`; not wired into CI/`test:focused` — it's a manual timing demo, not a correctness assertion).

`tsc --noEmit` clean on both `tsconfig.node.json` targets.

## Scope

Touches only `src/main/transcript.ts` (+ the demo video/benchmark script). No change to `readAgentUsage`'s exported signature or observable behavior for any existing caller (`usage.ts`, `telemetry.ts`, `index.ts`).


aaroncoville (commented):
Reviewed against the actual `src/main/transcript.ts` callers and cache lifecycle. The core change is solid: `src/main/transcript.ts:204-231` prices each assistant record once, updates the unfiltered accumulator and exactly one non-empty session bucket, and `:312-320` preserves the prior filtered/unfiltered API behavior. The existing regression at `test/transcript-usage.test.cjs:123-134` is stronger than the PR summary makes obvious: it primes `s1`, appends mixed sessions, then reads `s1`, `s2`, and unfiltered totals, so it exercises cross-filter cache reuse and incremental updates. I also ran that suite (10/10) and both TypeScript typechecks on this head.

One concrete documentation/test mismatch to clean up: the PR body says the benchmark is env-configurable through `BENCH_AGENTS` / `BENCH_ROUNDS`, but `test/usage-cache-benchmark.manual.cjs:29-30` hard-codes `10` and `30` and never reads either variable. Either wire those vars (with finite positive validation) or remove that claim. This does not affect runtime correctness, but it makes the supplied reproduction interface inaccurate.

savvaskoualis:
Yep, that one's on me. The env vars only ever existed as an uncommitted edit in my working copy — I recorded the videos with it, then pushed the version from before it. So the PR body describes a script that isn't actually on the branch.

Fixed in `a6538e10`: both vars get read now, and a bad value (`-5`, `0`, `2.5`, `Infinity`, or junk) errors out instead of silently falling back — didn't want a benchmark quietly running a different workload than the one you asked for. Unset still gives you 10/30. Nothing goes near `transcript.ts`, suite's still 10/10.

Good call on `:123-134` too — that test does more than I gave it credit for in the summary. The `s1` → `s2` → unfiltered reads off a single cache entry are the actual guard here.

chaitanyagiri:

---
The diagnosis in the new `parseUsageLines` comment is the right one and it is worth keeping in
the code: keying the cache by filter meant one file's new bytes were JSON parsed once per
querying agent, so the cost scaled with active agent count, which is exactly when it must not.

I walked the filtered read for equivalence and it holds. A record with no `sessionId` used to be
skipped by a filtered query and now lands only in `all`, so a filtered query still misses it. A
file with no records for the queried session used to contribute zeros and now `continue`s. Same
answer both ways.

**Two things.**

**There is no automated test for the behaviour change.** The evidence is two mp4s and
`usage-cache-benchmark.manual.cjs`, which is a benchmark and, by the `.manual.` in its name, not
in the suite. This is the accounting that feeds the per agent token cap and the cost ledger. A
refactor that changes how a filtered total is computed, with no test asserting the filtered
number is unchanged, is one transposed accumulator away from silently wrong billing numbers that
nothing would ever contradict. A single test parsing a fixture with two sessions in one file and
asserting both the filtered and unfiltered totals would cover it.

**What bounds `perSession`?** On every refresh where size or mtime moved:

```
perSession: new Map(Array.from(cached!.perSession, ([k, v]) => [k, { ...v }]))
```

That is a full clone of every session bucket in the file, on every incremental read. It is cheap
while a file holds a handful of sessions, and `USAGE_CACHE_MAX` bounds the number of file entries
but nothing bounds the sessions inside one. If a long lived transcript accumulates many session
ids across days, this clone grows with them and is never pruned, and it runs on the ~30s usage
beat. Do transcripts get one session id each, or can one file carry many? If it is one each, say
so in a comment and this is a non issue.

The benchmark's `benchCount` refusing an unparseable env var instead of falling back is a nice
touch. A benchmark that quietly runs the wrong workload reports a number for something you did
not ask about.

savvaskoualis:
Both checked properly before answering. One of them I have to push back on.

## The test already exists — but you're right about what it doesn't assert

`test/transcript-usage.test.cjs:117-126` is already the test you described: two sessions in one
file, filtered reads for `s1` and `s2`, and the unfiltered total, across an incremental append.
It's in the suite, not the `.manual.` benchmark. So "no automated test for the behaviour change"
isn't right as stated.

But I didn't want to just say that, because "a test exists" and "a test would catch this" aren't
the same claim. So I mutation-tested the suite against this branch — seven mutations, each
applied to `transcript.ts` and run against the unmodified 10 tests:

| mutation | before |
|---|---|
| filtered read returns `all`, ignoring the filter | **caught** |
| `inputTokens` / `outputTokens` transposed | **caught** |
| `estimatedCostUsd` added to `all` but not the session bucket | *survived* |
| `cacheReadTokens` / `cacheWriteTokens` swapped | *survived* |
| `model` set on `all` but not the session bucket | *survived* |

So the filter itself is guarded, and a transposed *token* accumulator is caught. What isn't
guarded is every other field on the filtered path — including `estimatedCostUsd`, which is the
exact number you were worried about. Your conclusion was right; the route to it was that the
existing assertion is `outputTokens`-only, not that the test is missing.

(The cache read/write swap survives on `origin/main` too — the fixture hard-codes both to 0. Not
introduced here, but it's in the same blind spot so I fixed it in the same pass.)

`24f37acb` adds two tests: one file, two sessions, distinct models and distinct cache read/write
counts so a swap can't cancel out, asserting all four token fields plus `model` per bucket. Cost
is asserted relationally — each session priced above zero, the two summing to the unfiltered
total — so it pins the split without hard-coding a price table that'll drift. The second test
pins the equivalence you walked by hand: a record with no `sessionId` counts toward the
unfiltered total and toward no filtered one. **All three surviving mutations now fail the
suite.** 12/12 green, both tsconfig targets clean.

## `perSession` is bounded at one entry per file

Measured rather than assumed. Across every transcript under `~/.claude/projects` on this
machine — 2172 top-level files, the ones `readAgentUsage`'s `readdirSync` actually reads — **every
single one holds exactly one distinct `sessionId`, and it always equals the filename stem.** Zero
exceptions. Claude Code names a transcript for its session and only appends that session's
records to it.

The one place I found files carrying two session ids is subagent transcripts, which pick up the
parent's id and the resumed parent's — 2 files out of 4531. Those live in a nested
`<session>/subagents/` directory, and the `readdirSync` is non-recursive, so they're never read
by this code path.

So the clone is O(1), not O(sessions-ever-seen), and there's nothing to prune. Documented on the
field in `24f37acb` rather than left as tribal knowledge.

Worth flagging one thing I found and deliberately didn't change: dropping the clone entirely and
sharing the cached `Map` is an equivalent mutant — nothing observable changes, because the old
entry is replaced in the cache and nothing else holds a reference. The clone is only buying
consistency between `all` and `perSession` if a mid-parse read throws. Given it's O(1) that felt
like the wrong thing to trade away for tidiness, but say the word if you'd rather it go.

PR #337, opened 2026-08-26
fix/security: confine the file IPC to the workspace when symlinks are involved

aaroncoville:
## What & why

The file IPC that backs the in-app editor confined paths to the selected
workspace with a purely **lexical** check (resolve + normalize + relative), and
then let Node open the result. A symlink placed *inside* the workspace but
pointing outside it — the kind of thing an agent-generated repo or a cloned
repo routinely contains — is lexically in-bounds, so the guard passed it and the
open followed the link to its external target. That allowed reading any file the
user can read and overwriting any file the user can write, from inside the
sandbox. `getDiff`'s working-tree read had the same exposure with an additional
final-component TOCTOU window.

The fix canonicalizes the workspace root and the target with `realpath`,
re-checks containment on the canonical paths, rejects any symlink component that
survives canonicalization (this is what catches a *dangling* link, which
`realpath` cannot resolve), and opens with `O_NOFOLLOW`. Reads are centralized
through one `openForRead` helper so the flags cannot drift between sinks, and
`getDiff` now stats through the open handle rather than re-resolving the path.

In-workspace symlinks whose target stays in-workspace are still followed (so a
`node_modules` symlink remains browsable); only out-of-workspace targets and
dangling links are refused.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

## Evidence

A self-contained PoC driver builds a throwaway workspace, loads the real
`src/main/fs.ts` + `src/main/git.ts`, runs 7 symlink-escape attacks plus 3
controls, and cleans up — no real external file is ever touched.

### Before
CleanShot 2026-08-26 at 14 01 53@2x

### After
CleanShot 2026-08-26 at 14 02 53@2x

## How I tested it

- OS: macOS (Darwin)
- Steps:
  - `npm run typecheck` — 0 errors (node + web)
  - `npm run test:focused` — 591 of 591 (9 new containment tests, incl. a
    deterministic check-to-open TOCTOU test for the `getDiff` read that shadows
    `git` on `PATH` to swap the final component mid-call)
  - `npm run build` — succeeds
  - the PoC driver above, before and after

## Discord (optional)

Discord:

## Checklist

      spacing, or fonts. (No UI change.)
- [x] If I added art, it's my own or compatibly licensed. (No art.)

### DIFF
--- a/src/main/fs.ts
+++ b/src/main/fs.ts
@@ -1,22 +1,146 @@
-import { readdir, readFile, writeFile, stat } from 'node:fs/promises';
-import { isAbsolute, join, normalize, relative, resolve } from 'node:path';
+import { readdir, lstat, open, realpath, stat } from 'node:fs/promises';
+import type { FileHandle } from 'node:fs/promises';
+import { constants } from 'node:fs';
+import { basename, dirname, isAbsolute, join, normalize, relative, resolve, sep } from 'node:path';
 import { homedir } from 'node:os';
 import { imageMimeForPath } from '../shared/imageTypes';
 
 /**
- * Confines `path` inside `root` to prevent path-traversal escapes.
- * Returns the resolved absolute path on success, or null on violation.
+ * Lexical containment — pure string math, no filesystem access.
  *
- * Exported so other main-process modules (e.g. git.ts) validate caller-supplied
- * relative paths against a workspace root with the SAME guard — there is exactly
- * one path-escape policy in the app and it lives here.
+ * PRIVATE on purpose. On its own this is NOT a containment guarantee: `resolve`,
+ * `normalize` and `relative` know nothing about symlinks, while the `readFile` /
+ * `writeFile` / `readdir` that runs afterwards is resolved by the kernel, which
+ * follows them. `safeResolve` is the guard every consumer must use.
  */
-export function safeJoin(root: string, rel: string): string | null {
+function lexicalJoin(root: string, rel: string): { absRoot: string; absPath: string } | null {
   const absRoot = resolve(root);
   const absPath = isAbsolute(rel) ? normalize(rel) : resolve(absRoot, rel);
   const rel2 = relative(absRoot, absPath);
   if (rel2.startsWith('..') || isAbsolute(rel2)) return null;
-  return absPath;
+  return { absRoot, absPath };
+}
+
+/**
+ * Canonicalize `absPath` and re-check containment against the canonical root.
+ *
+ * `realpath` throws ENOENT for a path that does not exist yet — which is normal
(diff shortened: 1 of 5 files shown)


chaitanyagiri:

---
**This should land.** The old guard was `resolve` plus `relative` and nothing else, and the
docstring on the new private helper says exactly why that was never enough:

```
On its own this is NOT a containment guarantee: `resolve`, `normalize` and `relative` know
nothing about symlinks, while the `readFile` / `writeFile` / `readdir` that runs afterwards is
resolved by the kernel, which follows them.
```

The dangling-link case is the one most implementations of this miss, because `realpath` throws
ENOENT on it and canonicalization alone waves it through, and then the write creates the file at
an external target. Catching it with an `lstat` walk and pinning it with
`a dangling symlink cannot be used to create a file outside the workspace` is the right answer.

And the TOCTOU test is the best control in any of the 34 PRs I read today. Shadowing `git` on
`PATH` with a script that deletes the file and plants the symlink, so the swap happens
deterministically inside the window between the check and the open, is a real proof rather than a
hope of winning a race. `assert.equal(fs.lstatSync(target).isSymbolicLink(), true, 'the shim must
have run inside the window')` is the line that makes the rest of that test mean anything.

**One gap, on the write path, and it is the asymmetry that makes me confident it is unintended.**

```
const READ_FLAGS  = constants.O_RDONLY | (constants.O_NOFOLLOW | 0) | (constants.O_NONBLOCK | 0);
const WRITE_FLAGS = constants.O_WRONLY | constants.O_CREAT | constants.O_TRUNC | (constants.O_NOFOLLOW | 0);
```

`READ_FLAGS` carries `O_NONBLOCK` and the comment above it says why: a FIFO would park the open,
the IPC call, and the renderer's loading state forever. `WRITE_FLAGS` does not carry it. Opening a
FIFO `O_WRONLY` blocks until a reader shows up, which is the same hang on the other verb.
`O_NOFOLLOW` does not help, a FIFO is not a symlink, and `writeFileText` has no `isFile()` check
either where both read paths do. So a FIFO sitting in the workspace, which agent-generated content
and cloned repos can both produce, hangs a save.

That is a follow-up and not a reason to hold this. **The security posture is strictly better with
this merged than without it, and a FIFO hang on write is a bug in a tree that currently reads
external files through symlinks.**

Small note: `constants.O_NOFOLLOW | 0` evaluates to `0` on Windows because the constant is
undefined there, which the comment states plainly, and the `lstat` walk is what covers Windows
instead. Saying that in the code rather than leaving it to be discovered is the right call.

PR #338, opened 2026-08-26
fix(codex): pre-trust a worker's working directory so it can boot una…

aaroncoville:
## What & why

A Codex agent spawned with nobody at the keyboard would sit at its very first turn and never
start. Codex gates every working directory behind an interactive *"do you trust this directory?"*
prompt, and the answer is persisted per config home as a `[projects."<dir>"] trust_level =
"trusted"` table. Each agent is given its **own** isolated `CODEX_HOME` so the user's `~/.codex` is
never mutated — and that home is generated fresh, so it has never answered the prompt for anything.
Every such agent therefore met the prompt on turn one and stalled there indefinitely.

Pre-seeding that table is the only suppression that exists. No flag or environment variable skips
the prompt (notably **not** `--dangerously-bypass-approvals-and-sandbox`, which governs command
approval, not directory trust), and a `-c projects…` override is not persisted, so it does not read
as an answer either. Verified against codex-cli 0.149.1.

So the fix appends the trust entry to the generated `config.toml` before the CLI starts, keyed on
the **final** working directory (git isolation has already resolved by the time spawn prep runs, so
this is the agent's own worktree when one was created and the original repo when isolation fell
back) in its **canonical, realpath-resolved** spelling (Codex resolves the directory before looking
it up, so an entry keyed `/tmp/x` never matches its `/private/tmp/x` and the prompt returns). The
path goes through a TOML basic-string escaper rather than string interpolation: a directory name may
legally contain a quote or a backslash on POSIX, and an unparseable `config.toml` would be a worse
failure than the prompt it was meant to remove.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before
CleanShot 2026-08-26 at 14 45 52@2x

### After
CleanShot 2026-08-26 at 14 49 13@2x

## How I tested it

- OS: macOS 15 (Darwin 25.6.0), Node via the repo toolchain; codex-cli 0.149.1.
- Steps:
  1. Measured the baseline on the branch point first: `npm run test:focused` on `main` →
     **552 of 552 passing**, `npm run typecheck` → 0 errors.
  2. On this branch: `npm run typecheck` → **0 errors**.
  3. `npm run test:focused` → **556 of 556 passing**, 0 failing (the 4 new tests are the delta;
     no pre-existing failures on either ref).
  4. `npm run build` → succeeds.
  5. Confirmed the tests are load-bearing by reverting only `src/main/hive.ts` to `main` while
     keeping the new tests: all 4 go red (transcript under **Before**). Restored, green again.
  6. Checked the escaping and canonicalisation cases explicitly — a directory containing a quote
     round-trips through the TOML escaper, and the entry is keyed on the realpath, which is what
     makes it match on macOS where `/tmp` resolves to `/private/tmp`.

## Discord (optional)

Discord:

## Checklist

      spacing, or fonts. *(No UI in this change.)*
      `ATTRIBUTION.md`. *(No art in this change.)*

### DIFF
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -20,7 +20,8 @@
  */
 import {
   existsSync, mkdirSync, readFileSync, writeFileSync, renameSync,
-  readdirSync, statSync, rmSync, appendFileSync, symlinkSync, copyFileSync, chmodSync
+  readdirSync, statSync, rmSync, appendFileSync, symlinkSync, copyFileSync, chmodSync,
+  realpathSync
 } from 'node:fs';
 import { join, dirname, isAbsolute } from 'node:path';
 import { homedir } from 'node:os';
@@ -41,6 +42,7 @@ import { selectBroadcastTargets } from '../shared/broadcast';
 import { preferredAgentRole } from '../shared/agentRole';
 import { mergeTaskLedger } from '../shared/taskLedger';
 import { expandTilde } from './fs';
+import { codexTrustEntry } from '../shared/codexTrust';
 
 /** The subset of HarnessConfig the hive consumes for the default-MCP merge.
  *  Kept as a local shape so hive.ts never imports the foundation-owned config
@@ -751,7 +753,7 @@ export class HiveManager {
           if (desc.kind === 'hooks') {
             if (desc.shim === 'agy') this.installAgyHooks();
             else if (desc.shim === 'codex') {
-              env.CODEX_HOME = this.installCodexHooks(dir);
+              env.CODEX_HOME = this.installCodexHooks(dir, meta.cwd);
               // Codex refuses to run hooks from a config dir without persisted
               // "hook trust" (normally an interactive gate). Our hooks.json is
               // hive-authored inside an isolated CODEX_HOME, so we bypass that gate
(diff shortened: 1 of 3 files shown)


chaitanyagiri:

---
**Read this one before rebasing, because the conflict is the dangerous part and it typechecks
either way.**

This branch starts from `private installCodexHooks(dir: string): string` and adds a second
parameter, `cwd`, with the call site becoming `this.installCodexHooks(dir, meta.cwd)`.
**`origin/main` at `956bfb4c` has ALSO already added a second parameter to the same function, and
it is a different one:**

```
private installCodexHooks(dir: string, agentId: string): string
...
env.CODEX_HOME = this.installCodexHooks(dir, meta.id);
```

Both sides added a second positional `string` in the same slot with different meanings. A
conflict resolution that keeps two arguments without reading them lands `agentId` where this code
expects `cwd`, **and the compiler will not say a word, because both are `string`.** What happens
then is quiet: `realpathSync('jim-1')` throws, the catch keeps the raw value, and the config gets
`[projects."jim-1"]`, a trust entry for a directory that does not exist. The prompt comes back and
every Codex agent stalls on its first turn with the fix apparently applied. Worth a line in the
PR body so whoever rebases knows to check the argument and not just the arity.

Separately, this also collides with #356, which rewrites the `command = '...'` line this diff
still carries in its single-quoted form.

On the change itself. `tomlBasicString` is correct, including astral code points, since `for...of`
yields whole code points and anything above `0x7f` passes through unescaped as both units. The
realpath detail is the kind of thing that is only ever learned the hard way and the comment
explains it properly. `verified on Codex 0.149.1` is the right way to state a claim about another
tool's behaviour, because it says what would have to be rechecked.

The one thing I would add to the PR body: this pre-answers a security prompt, and a reader will
stop there. It is worth saying out loud that the directory being trusted is the agent's own cwd,
that the same thing already happens for Claude via `hasTrustDialogAccepted` in
`src/main/config.ts`, and that the exposure is therefore unchanged rather than new. Without that
sentence it reads like a new hole.

PR #339, opened 2026-08-26
Upstream/model catalog json

aaroncoville:
## What & why

Every provider's model picker was a hardcoded `ModelOption[]` in the renderer store, so
shipping a model — one string — meant editing renderer source, type-checking and
rebuilding. The arrays also could not say *which releases* a model belongs to: a build
whose CLI never shipped a model still offered it, and a retired model stayed in the
picker forever.

The twelve arrays now live in `src/shared/modelCatalog.json`, ported verbatim, each entry
carrying inclusive `minAppVersion` / `maxAppVersion` bounds. Every ported entry has `null`
for both, so this build offers exactly what it offered before — the version fields are the
mechanism, not a behaviour change. The catalog is imported at **build** time (no fs, no
network, offline-safe) and `modelsForProvider` filters it by the running app version, which
electron-vite already inlines into the renderer as `__APP_VERSION__` for the update badge.
Adding a model is now a one-line JSON edit.

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

## Evidence

No visible UI change: the pickers offer the same models, in the same order, with the same
labels. The evidence is the test suite going red → green, plus proof that the built bundle
really carries the version into the filter.

Reproduce both halves:

```
node --test test/model-catalog.test.cjs                                        # green
git checkout <base> -- src/renderer/src/store/config.ts test/load-ts.cjs
node --test test/model-catalog.test.cjs                                        # red
```

### Before
CleanShot 2026-08-26 at 14 58 15@2x

The catalog-backed API does not exist, so the version-bound assertions fail. The three
tests that *do* pass are the ones pinning today's picker output — they pass against the
hardcoded arrays, which is what makes them a faithful record of what shipped:

### After
CleanShot 2026-08-26 at 14 58 46@2x

Same file, same assertions — the picker-output tests still pass (so nothing a user sees
changed), and the version filter now works:

And the production bundle proves the version reaches the filter — esbuild folds the
build-time define straight through, so there is no round trip to main:

```
$ npm run build && grep -n -A3 'function runningAppVersion' out/renderer/assets/index-*.js
40935:function runningAppVersion() {
40936-  return "0.4.5";
40937-}
40938-function offeredAtVersion(model, appVersion) {
```

## How I tested it

- OS: macOS 15 (Darwin 25.6.0), Node 22
- Steps:
  - `npm run typecheck` — 0 errors (node + web).
  - `npm run test:focused` — **560 of 560 pass**; the base commit is 552 of 552, and the
    8 new tests are the whole difference.
  - `npm run build` — succeeds; bundle checked as above.
  - Mutation-checked the new tests: disabling the `minAppVersion` check, dropping the
    `maxAppVersion` check, editing one catalog label, and making the version accessor
    ignore `__APP_VERSION__` each turn the matching test red.

## Checklist

- [x] Any new UI derives from `DESIGN.md` / `tokens.ts` — no ad-hoc colors, spacing,
      or fonts. (No UI in this change.)
      `ATTRIBUTION.md`. (No art.)

---

## How the catalog works & how to maintain it

The provider model lists live in `src/shared/modelCatalog.json` and are imported
at build time. `modelsForProvider(provider)` returns that provider's list,
filtered at runtime by the running app version.

There are **two independent "version" concepts** — they do not map to each other:

- **Top-level `"version"`** is the *file schema version*. Nothing reads it today;
  it exists so that if the file's **structure** ever changes, code can branch on
  it. Bump it only when the shape of the JSON changes — never for adding or
  removing a model.
- **Per-model `minAppVersion` / `maxAppVersion`** are the *app-version bounds*
  (inclusive; `null` = unbounded). A model is shown only when the running app
  version falls within its bounds. Every model currently ships `null`/`null`, so
  the picker is identical to the previous hardcoded lists.

### Add a new model

Add an entry under the provider in `modelCatalog.json`:

```json
{ "id": "gpt-6-nova", "label": "GPT-6 Nova", "minAppVersion": null, "maxAppVersion": null }
```

If the model only works from a future release, set `minAppVersion` to that
release (e.g. `"0.5.0"`) so users on older builds — whose CLI can't run it —
don't see it. Otherwise leave both bounds `null`.

### Retire an old model

Set `maxAppVersion` to the **last release that should still offer it**:

```json
{ "id": "gpt-5.6-sol", "label": "GPT-5.6 Sol", "minAppVersion": null, "maxAppVersion": "0.4.9" }
```

Builds newer than `0.4.9` stop offering it, while users still on `0.4.x` keep
seeing it. To remove a model **everywhere at once** (all versions), just delete
its entry instead.

### DIFF
--- a/.github/workflows/pr-evidence.yml
+++ b/.github/workflows/pr-evidence.yml
@@ -75,7 +75,9 @@ jobs:
             // Comments are the template's own instructions to the author. Left
             // unfilled they are invisible to a reader, so they must be invisible
             // to this check too — otherwise the empty template would pass.
-            const visible = body.replace(/<!--[\s\S]*?-->/g, '');
+            const visible = body
+              .replace(/<!--[\s\S]*?-->/g, '')
+              .replace(/^(```|~~~)[\s\S]*?^\1\s*$/gm, '');
 
             // ── Before and after, separately ─────────────────────────────────
             // Two assets in one blob is not the same as a before and an after.
@@ -84,7 +86,7 @@ jobs:
             // right place to count for it.
             const section = (name) => {
               const re = new RegExp(
-                `^#{1,6}\\s*${name}\\b[^\\n]*\\n([\\s\\S]*?)(?=\\n#{1,6}\\s|$)`,
+                `^#{1,6}\\s*${name}\\b[^\\n]*\\n([\\s\\S]*?)(?=^#{1,6}\\s|(?![\\s\\S]))`,
                 'im'
               );
               return (visible.match(re) || [, ''])[1];
--- a/docs/superpowers/plans/2026-08-25-pluggable-memory.md
+++ b/docs/superpowers/plans/2026-08-25-pluggable-memory.md
@@ -1,0 +1,264 @@
+# Pluggable Memory Backends Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Make the semantic memory layer configurable behind a `MemoryBackend` interface, with MemPalace (default, behavior-identical) and Hindsight (HTTP) adapters and a backend-neutral `hive-memory` shim for agents.
+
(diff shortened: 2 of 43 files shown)


chaitanyagiri:
Shipped in v0.4.6.

Your commits `e8e44749` and `727b4f56` are in the release: the provider model lists moved into `src/shared/modelCatalog.json`, `src/renderer/src/store/config.ts` lost 182 lines to it, and the TypeScript test loader now handles `.json` imports. `test/model-catalog.test.cjs` ships with it.

Note this PR's branch also carried `2b96d5d0`, which is #317's commit. That one is credited there, not here, so nothing is double counted.

Thank you.

PR #343, opened 2026-08-26
feat(telemetry): count human-sent messages, closing the activation funnel

chaitanyagiri:
## What this is

The activation funnel currently ends at `agent_spawned`. We can watch someone open the app, pick a provider, install the CLI and get an agent running, and then the data stops. An agent that was started and then never spoken to looks identical to one in daily use, which is exactly the moment worth measuring.

This adds **`message_sent`** — a count of the messages a person sends to an agent.

## What it does not do

No content, in any shape:

- no message text
- no length or character count
- no first-N-characters
- no hash of the body

The event has exactly one property, `surface`, and it is a closed enum. The IPC channel the renderer uses does not take a message argument at all, so there is no shape in which the body could cross it by accident. A test asserts the full property set by exhaustion, so a future leak fails the build rather than needing to be named in advance.

The separate question of whether content telemetry should ever exist is untouched and still open.

## Where it counts, and why there

`pty:write` fires on **every keystroke**. Counting there would have produced a keystroke meter, not a message count, orders of magnitude above the 7.47-events-per-active-install-per-day baseline. Every surface is instrumented at the **submit** instead:

| Surface | Counted at | Why not one level up |
| --- | --- | --- |
| `terminal` | `terminalPool` submit boundary, once per input chunk | `pty:write` is per keystroke |
| `composer` | inside `MessageQueueComposer.queueIt` | `enqueueMessage` also carries work orders, Slack inbound, nudges and compact commands |
| `steer` | the `control:steer` IPC handler | `control.steer()` is also called by `closingTime` and the voice action layer, neither of which is a person typing |
| `hive` | `hive:send`, gated on `from === 'human'` | agent-to-agent hive traffic would swamp the number |

A bracketed paste is excluded from the terminal count. xterm rewrites pasted newlines to `\r`, so without that guard a five-line paste would have looked like five messages, and the Enter that actually sends it would have been a sixth.

## IME

Inherited, not restated. Both renderer surfaces sit past the existing `isComposingKey` guard, so an Enter that picks a CJK candidate is not a submit and cannot inflate the count for the users the i18n work was for. A test pins the guard as the first thing the composer's key handler does, so removing it fails.

## Trust boundary

The renderer may name only `terminal` and `composer` — the two submits main cannot observe for itself. `steer` and `hive` are counted in main at handlers that already receive them, so letting the renderer name those two would open a double-count. The event name is fixed main-side: the renderer chooses a surface, never an event, and never a property.

## Opt-out

Every path goes through `track()`, so all three existing opt-outs suppress it: the Settings switch (at boot and flipped live mid-session), `DO_NOT_TRACK`, and a build with no PostHog key. Each is asserted by observation in the tests rather than by inspection.

## TELEMETRY.md

Updated in the same commit, with a dedicated section spelling out what `message_sent` does and does not carry.

A new test asserts that the allowlist in `analytics.ts` and the doc's event table name **exactly the same events**, in both directions. This repo has shipped two contradictions between published telemetry policy and actual behaviour; a third now fails the build instead of shipping.

## Tests

22 new tests. They cover the fails-open shape, not just the function — a miscounted event is a wrong number someone notices, but a mis-wired one records zero forever while CI stays green and reads as "nobody talks to their agents". So the IPC channel literal on both sides of the bridge, the surface literal at each call site, and the doc/code lockstep are all asserted.

Twelve mutations were checked and all twelve go red, each on its own test (proving the guards are independent rather than restatements of each other): paste guard removed, count moved off the submit boundary, composer stops counting, hive counts agent traffic, steer stops counting, IPC validation dropped, renderer allowed to name every surface, runtime enum guard dropped, the meter deduped like `feature_used`, `length` added to the allowlist, the doc row removed, and the channel name typoed on one side.

Suite 715/715. Both typechecks clean.

## Volume

`message_sent` fires once per human submit. For an install whose owner actually talks to their agents, roughly 10-40 events per active install per day, taking the total from 7.47 to about 17-47 — call it 2x-6x. The keystroke version would have been roughly 4,000/day, around 500x.

The estimate is a projection, not a measurement, and the first week of data will settle it. The number is bounded by how fast a person can type and press Enter, not by anything the machine does on its own.

https://claude.ai/code/session_01LDbmAwrgtGskFGCRkYZBx4

### DIFF
--- a/TELEMETRY.md
+++ b/TELEMETRY.md
@@ -32,8 +32,33 @@ The events:
 | `agent_spawn_failed` | `provider`; `reason` is one of `cli_missing`, `cwd_missing`, `already_running`, `spawn_error` | A requested spawn did not start an agent |
 | `agent_install_started` | `provider`; `rung` is one of `npm`, `node-then-npm`, `native` | The engine CLI was missing, so the bundled auto-installer began |
 | `agent_install_finished` | `provider`; `rung` as above; `outcome` is one of `agent_launched`, `install_failed` | The auto-installer exited |
+| `message_sent` | `surface` — one of `terminal`, `composer`, `steer`, `hive` | Each time **you** send a message to an agent. A count of messages and nothing else: the message itself is never read, measured, or hashed |
 | `feature_used` | `feature` — one of `slack_trigger`, `webhook_trigger`, `hire_install`, `voice_dictation` | At most once per feature per app session |
 | `session_ended` | `duration_bucket` — one of `<5m`, `5-30m`, `30m-2h`, `2-8h`, `8h+` | On quit (coarse bucket, never raw duration) |
+
+### About `message_sent`
+
+This is the one event that fires while you work, so it is worth being precise
+about what it does and does not do.
+
+- It counts **one event per message you send**, at the moment you send it. It is
+  counted at the *submit* — the Enter or the Send button — never per keystroke.
+  The app does not report what you type as you type it.
+- `surface` says only *where* you sent it from: `terminal` (typed into an
+  agent's terminal), `composer` (the agent's message queue box), `steer` (the
+  steer field on the agent control strip), or `hive` (a dispatch, thread reply,
+  or ASK ME answer).
+- Messages **agents send each other** are not counted. Only messages that came
+  from a person are.
+- No property carries the message. There is no text, no length, no character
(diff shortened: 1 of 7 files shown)


chaitanyagiri:
Shipped in v0.4.6 as `c1756662`, merged in `91114503`. The message_sent count closes the activation funnel. Closing.

PR #344, opened 2026-08-26
fix(release-drop): kill the render-blocking font fetch + add a loader

chaitanyagiri:
## The problem

When the release drop opens on a new version, it shows a **blank paper-white rectangle** with a close button — long enough that a user reaches for cancel before the content ever appears. The founder reported it; the cause is not a slow parse.

The drop renders authored HTML in a sandboxed iframe, and a release drop **paints nothing until its stylesheets resolve**. The frame carried two remote font fetches:

1. an app-injected Google Fonts `<link>` in *every* drop, and
2. whatever `@import` an author wrote (the current `RELEASE.md` has one).

A CSS `@import`/`<link>` to a stylesheet is render-blocking. Where `fonts.googleapis.com` is **blocked — China — that fetch does not fail, it hangs to a TCP timeout**, so the frame stays white for tens of seconds. This is the exact same render-blocking Google Fonts fetch the app itself removed for launch in rc.2 ("fonts ship inside the app"); the release drop still did it, for the same market the whole i18n release was for.

Two fixes, kept in **two separable commits**, because they are not the same fix.

## Commit (a) — the real one: kill the remote font fetch

Three independent parts, so no single one is load-bearing:

- **CSP can no longer permit a remote stylesheet or font.** `style-src` drops `https:` (authored `<style>` still works via `'unsafe-inline'`); `font-src` is `data:` only. A remote `@import` is now **denied**, and a denied import **fails fast** — the frame paints immediately in the fallback instead of hanging.
- **The design fonts are self-hosted as `data:` URIs** — the same Inter + JetBrains Mono woff2 the app already ships. Nothing legitimate needs the network, and the drop keeps its typography with zero fetches. Inter answers the drop's `--font-sans` (its declared Geist substitute); JetBrains Mono is exact.
- **`stripActiveContent` drops an authored *remote* `@import` outright**, so the frame never even attempts it and no CSP-violation is logged. The regex consumes the whole statement including url-internal `;` (a font url is `wght@400;500;600`, which a naive `[^;]*` would truncate and leave broken CSS behind); a `data:` import is left alone.

**`RELEASE.md` is untouched** — this makes *any* authored drop safe, not just the current one, which is the app-side fix that was strongly preferred.

## Commit (b) — the safety net: a loader until the frame paints

Even with (a), there is still a window before paint — remote **images** (`img-src` stays `https:`) can hold the iframe's `load` event — and in it the modal showed a bare rectangle that reads as broken.

A loader now covers the frame until it is ready: the landing site's three square dots marching over warm paper, in the file's existing palette, **no new chrome button** (dismissal stays Esc / close / backdrop, per the file's design). It is revealed on the **first** of two signals:

- the iframe element's own **`onLoad`** (it fires on the parent, needs no script rights in the sandboxed child), or
- a **2.5s timeout cap**, whichever comes first.

The cap matters because `load` waits for subresources: without it a drop with a slow image would hold the loader for the whole fetch, so we reveal the paper page and let images arrive into it rather than sit on a spinner. Reveal is monotonic (latches true, never flips), so it cannot flicker.

Built against the **fails-open** shape: the iframe is **always mounted at full opacity**; the loader is a separate overlay removed on reveal, so a broken reveal degrades to a brief extra spinner, never a permanently hidden frame — and the timeout guarantees reveal fires even if `onLoad` never does.

## Proof, not a claim

Rendered in a real sandboxed frame with the font host pointed at a **non-routable IP** (the honest simulation of a firewall DROP of `fonts.googleapis.com`), screenshotting both the old and new build at an **identical 1.5s**:

- **Before:** a blank paper-white frame — the DOM is parsed (`document.body.innerText` is non-empty) but render is blocked on the hanging `@import`. The founder's exact complaint.
- **After:** fully rendered at the same instant — heading in self-hosted JetBrains Mono, body in self-hosted Inter, striped placeholder, all with **zero network**.

(Screenshots delivered to the founder in the hive.)

## Tests

7 new tests in `test/release-drop.test.cjs`, both commits independently green (commit a: 15, commit b: 17). They pin the fails-open shape, not just the render: the CSP closes the remote-fetch surface, the fonts are present as real `data:` woff2 payloads, the founder-shaped `@import` (url-internal `;`) is stripped with no broken remainder, a `data:` import is preserved, the loader reveals on `onLoad` **and** the cap, and the iframe is never gated on reveal. Six mutations checked — re-add the remote link, loosen `font-src`/`style-src` back to `https:`, drop the `@import` strip, remove the timeout cap, remove `onLoad` — each reddens its own test.

Full suite 700/700, both typechecks clean. No tag cut; rc.3 carries this.

https://claude.ai/code/session_01LDbmAwrgtGskFGCRkYZBx4


chaitanyagiri:
Shipped in v0.4.6 as `0e272a2c` and `20d85fde`, merged in `5471c0e5`. The render blocking font fetch is gone and the loader is in, so the release drop cannot white screen. Closing.

PR #345, opened 2026-08-26
feat(i18n): land the Arabic UI half of PR #213, gated on the chosen language

chaitanyagiri:
Lands the held UI half of #213 — Arabic strings and RTL layout — onto `release/0.4.7-rc`, gated so nothing changes for a user who has not selected Arabic. Also carries the founder's follow-up: picking Arabic turns terminal RTL on by itself.

Base is `release/0.4.7-rc` @ `071584ca`. Suite was **693/693 green** on a clean checkout of that before I touched anything; it is **709/709** now.

## Coverage

**1197 of 1197 keys translated. Nothing falls back to English.** Key tree is identical across `en` / `zh-CN` / `ar`, every `{{interpolation}}` variable survives, every `<strong>` survives, every flavour-text array keeps its length (they are indexed by the scene, so a short array is an out-of-range read).

Nine strings are identical to English **on purpose** and are allowlisted with reasons in the test: four provider blurbs (`Claude Code — Anthropic`), the Slack product name, two filesystem-path placeholders, and two pure format strings (`{{count}} · {{state}}`).

**The Arabic has not been reviewed by an Arabic reader.** It was written by an agent. The tests check coverage, shape, interpolation and inertness — not whether it reads well. A green run is not a review.

## The gate

One function: `isRtlLanguage(i18n.language)`. Not `navigator`, not the OS, not the text on screen. `dir` goes on `<html>` rather than a React wrapper, because modals and the fullscreen terminal portal outside the tree and would otherwise keep the document's direction while the rest of the app mirrored.

Exact-match on a registered code, never a prefix or script guess — an unknown value must not be able to mirror someone's UI.

**Verified by mutation.** Each of these makes the suite fail: unscoping the RTL CSS, un-pinning any of the four xterm selectors, ungating the markdown plugin, ungating a component `dir` site, making `isRtlLanguage` a prefix guess, restoring the OS locale sniff, removing the joiner undo.

## Four changes to what the PR proposed

**The Google Fonts `<link>` and IBM Plex Sans Arabic are not taken.** The PR re-added the `<link>` that `41ea4c37` deliberately removed and put the font in all three token stacks — a boot-time network fetch, blocked in mainland China, and an ungated type change for every user. The bundled stacks already name Geeza Pro and Noto Naskh Arabic. Verified in a real build: no `<link>`, no `@import`, no `url(https://fonts` anywhere in the output.

**`rehypeAutoDir` is gated.** It sets direction from *content*, so ungated it fires for an English user the moment an agent quotes a line of Arabic. That is exactly why Kevin held it. It now runs only for an RTL app language — the design decision he said was missing, which Arabic shipping is what resolves.

**xterm is pinned `direction: ltr` under an RTL page.** `direction` inherits, so without this the viewport picks up `rtl` from `<html>` and right-aligns the entire cell grid, reversing box-drawing frames while the pty still counts columns from the left. Per-row direction is a separate layer and already has `.cth-bidi`.

**Two Kanban Q&A `dir` sites dropped as superseded** — that trail now renders through `MarkdownPreview`, gated one level down.

## The terminal toggle

It was an artifact of build order: it shipped when Arabic could not be selected as a UI language at all, so a manual switch was the only door. Now that it can be selected, a user who picks Arabic has already answered the question the toggle asks.

It is a **default**, not a derived value. Three-valued: unset (follow the language) / forced on / forced off. An English user whose colleagues' logs carry Arabic has a real reason to switch it on; an Arabic user may want the GPU renderer's speed back. An explicit choice survives a language switch in **both** directions. Settings says which one you are looking at, and still explains the renderer tradeoff, in all three locales.

## Already-open terminals: upgraded in place, never rebuilt

The office scene can be rebuilt on a language switch (`a8292697`) because it is derived from state we still hold. **A terminal cannot** — its scrollback lives only in xterm's buffer and the pty will not resend it, so recreating one would silently eat the user's history.

Dropping the WebGL lease is enough. `releaseWebglRenderer` already hands painting to the DOM renderer with buffer, scrollback and pty subscription intact, and the DOM renderer is precisely what Arabic mode needs. `enableArabicRendering` / `disableArabicRendering` are exact inverses, which is what makes the switch live in both directions. A terminal that has never been opened is skipped — `attachTerminal` reads the setting fresh when it does.

## What breaks in mirror, honestly

- **95 physical-direction inline styles across 38 of 72 renderer components** (`marginLeft`, `paddingRight`, `textAlign: 'left'`, absolute `left:`/`right:`). Worst: `IdePanel` (9), `UpdateBadge` (8), `TasksKanban` (7), `FullscreenTerminal` (6). These do not mirror. Chrome will lay out right-to-left, and these will pull the wrong way against it — the result is asymmetric padding and controls that sit on the side they sat on in LTR. Not fixed here: it is a mechanical sweep of its own, and doing it blind without an Arabic reader in front of the running app risks making the layout worse.
- **Icons do not mirror.** Chevrons and arrows keep their LTR direction. Wrong for "next"/"back" affordances.
- **The office floor does not mirror**, deliberately — it is a Pixi canvas, and mirroring pixel-art desks and signage would be worse than leaving them. Its labels rasterize through the browser's text engine, so Arabic should shape, but **this is the least-verified area in the change**: I could not run the GPU scene headlessly.
- **The splash screen is English-only** and pre-React, so it never sees the locale. Brand text only.
- `global.css` is now fully logical (`.cth-md-card` was still physical; converted).

## Not touched

`main`, the `rc.1`/`rc.2` tags, `RELEASE.md`. No tag cut. Nothing under Settings → Connections, where Kevin is building the MCP directory.


chaitanyagiri:
Shipped in v0.4.6 as `6f7ffb89`, merged in `2aebc664`. The Arabic UI half is in, gated on the chosen language. Credit for the underlying terminal modules belongs to #213, which is noted there. Closing.

PR #347, opened 2026-08-27
fix(onboarding): make Stay Awake step cross-platform, not macOS-only

devrahulbanjara:
The permissions step's "Stay awake on power" panel hardcoded macOS copy ("macOS only lets you...", Battery -> Options steps) and a macOS-only deep link (x-apple.systempreferences:...), which the main process's openExternal allowlist also only permitted. On Windows this sent users to a "no app can open this link" dialog instead of Settings.

Derive the OS from window.cth.platform (already exposed by preload) and branch the copy, deep link, and button per platform: macOS opens the Battery pane, Windows opens ms-settings:powersleep (now allowed by the main-process allowlist), and Linux shows text-only guidance since there's no universal settings URI across desktop environments. Also generalized adjacent copy that assumed macOS ("Mac" -> "computer", "macOS asks permission" -> "your system may ask permission") across all three locales (en, ar, zh-CN).

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

## How I tested it

- OS:
- Steps:

## Discord (optional)

Discord:

## Checklist

- [ ] `npm run typecheck` passes.
- [ ] `npm run test:focused` passes.
- [ ] `npm run build` succeeds.
- [ ] This PR is **one change**. Unrelated fixes belong in their own PR.
- [ ] I read the diff myself before opening this, and there is no debug output,
- [ ] Any new UI derives from `DESIGN.md` / `tokens.ts` — no ad-hoc colors,
      spacing, or fonts.
      `ATTRIBUTION.md`.


chaitanyagiri:

---
Driving the copy and the button off `window.cth.platform` instead of hardcoding macOS is right,
and hiding the button entirely on Linux rather than deep-linking something that does not exist is
the honest version. All three locale files on `origin/main` at `956bfb4c` are updated, ar, en and
zh-CN, so nothing is left behind today.

**The thing to watch is the interaction with the other two locale PRs, because the keys were
RENAMED and not added alongside.** `stayAwakeDesc` becomes `stayAwakeDescMac` / `Windows` /
`Linux`, and `openBattery` becomes `openBatteryMac` / `openBatteryWindows`. #360 adds `pt-PT` and
`pt-BR`, and #361 moves a large number of hardcoded strings into the locales. Whichever of those
lands first will carry the OLD key names, and this PR will not touch them, so Portuguese gets the
key string or an English fallback on that panel with nothing failing. Worth a note in whichever
of the three merges second.

The IPC change widens the `app:openExternal` allowlist to `ms-settings:`. Same class of scheme as
the macOS one already there, still an allowlist, still anchored at the start of the string. No
concern, and the comment now names all three cases including why Linux has none, which is the part
that would otherwise read as an oversight.

One small thing: the keys are now built by interpolation,
`t(\`onboarding.permissions.stayAwakeDesc${...}\`)`. Any tool that extracts keys statically will
stop seeing these five. If nothing in the repo does that, ignore this.

PR #348, opened 2026-08-27
fix: make the test suite terminate, then make it a CI gate

dadegrande99:
## What & why

The 745-test suite in this repo never ran in CI, and `npm run test:focused` never
terminated on its own, so nobody noticed. The two defects hid each other. This branch
fixes the suite, then makes it a gate, and clears three smaller items found in the same pass.

Five commits, one per concern, applied in dependency order:

- `726d989d` **test: fix the `test:focused` hang.** Adds `--test-timeout` as a backstop,
  then fixes the two real causes: `proc-kill.test.cjs` ran its cases inside an async IIFE
  and never registered with `node:test` at all, and `hive-hook-node.test.cjs` spawned
  children with `stdio: ['pipe','pipe','pipe']` while draining only `stderr` — once a child
  exceeded the pipe buffer it blocked forever and `close` never fired.
- `6b3294d2` **ci: add a Test job, stop ignoring Build failures.** New `Test` job with
  `timeout-minutes: 8`, and `continue-on-error: true` removed from `build` — as written, no
  build breakage had ever blocked a merge.
- `56634c6a` **deps: drop `localtunnel` and `@types/localtunnel`.** Unused since v0.2.4,
  when Slack and webhook ingress moved to `tunnelmole` because `loca.lt` began serving an
  interstitial that broke Slack's `url_verification`. The declaration was never removed.
- `a23b806a` **test: stop three tests writing into the real `$HOME`.** They created
  `~/.md-statabs-*` and `~/HarnessAgents` on the developer's machine and failed with `EPERM`
  anywhere `$HOME` is protected. Uses the redirect pattern this repo already applies in
  `hive-hook-node.test.cjs`.
- `68c53034` **chore: untrack an accidentally-committed 12MB build chunk.**
  `index-j0JdoH0M.js` is a Vite/PIXI output committed by mistake in `0c4f0bde`, with zero
  references anywhere. `git rm --cached` only — the file stays on disk and history is
  untouched.

`package-lock.json` is deliberately absent from all five commits: it had unrelated pending
changes that should not be folded in here.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [x] Build / CI

### Before

```
$ time npm run test:focused
(no output at all; still running past 600s; killed manually at ~20 minutes)
```

### After

```
$ time npm run test:focused
ℹ tests 749
ℹ pass 744
ℹ fail 4
ℹ cancelled 1

npm run test:focused  51.43s user 10.35s system 173% cpu 35.521 total
```

`npm run typecheck` is green on both node and web.

The four remaining failures are environment-bound, not regressions: three
`telemetry-forget-agent` cases fail with `Invalid URL: 'null/v1/metrics'` because binding
`127.0.0.1:0` was blocked in the sandbox this ran in, and `hive-hook-node` times out binding
its `HIVE_SOCK` unix socket for the same reason.

One case deserves a maintainer's eye rather than a claim from me: `hardKillTree reaps the
WHOLE group` fails here. It is **not a regression** — it lived inside the async IIFE and was
never executed before, so this branch is the first time it has ever run. Whether it fails
because process-group kills were restricted in this sandbox, or because it is genuinely
broken, needs one run in an unrestricted environment.


dadegrande99:
Closing: opened by mistake against the wrong repository. Apologies for the noise.

Issue #350, opened 2026-08-27
every Claude Code hook fails when the hive path contains a space


AryaNyoman:
### What happened?

# Bug report — every Claude Code hook fails when the hive path contains a space

**Component:** hive hook scaffolding (`agents/<id>/settings.json`, `hive/bin/hive-node.cmd`)
**Severity:** medium — non-blocking, but 100% of hook events fail and the user
cannot fix it themselves (see §7)
**Platform:** Windows only

---

## 1. Summary

On Windows, Munder Difflin writes each Claude Code agent's hook command as:

```
"<hivePath>\bin\hive-node.cmd" "<hivePath>\bin\cth-hook.cjs"
```

When `<hivePath>` contains a space, **every hook invocation fails** with exit
code 1. The hook→harness bridge is completely dead for that agent, silently, for
the entire life of the session.

The status line uses the same string plus `--status` and is affected the same way.

## 2. Environment

| | |
|---|---|
| Munder Difflin | v0.4.6 (latest) |
| Claude Code | v2.1.247 |
| OS | Windows 11 Home 10.0.26200 |
| Agent | the `god` agent (Claude Code engine) |
| Hive path | `D:\CLAUDE MADE ARYAWA PUTRA\hive` ← contains spaces |

## 3. Symptom

Shown in the agent's terminal after essentially every action:

```
PostToolUse:Read hook error
Failed with non-blocking status code: 'D:\CLAUDE' is not recognized as an
internal or external command,
operable program or batch file.
```

Note the truncation at the first space: `D:\CLAUDE`.

## 4. Scope — all hook events, not just one

Extracted from the agent's own transcript
(`~/.claude/projects/<slug>/<session>.jsonl`, records of type
`hook_non_blocking_error`): **52 consecutive failures in a single session,
exit code 1, zero successes.**

Affected events observed: `SessionStart`, `UserPromptSubmit`, `Stop`, and
`PreToolUse` / `PostToolUse` for every tool (`Read`, `Bash`, `PowerShell`,
`Glob`, `Grep`, `Skill`, `AskUserQuestion`).

Every failing record carries the identical command string:

```json
{
  "hookEvent": "PostToolUse",
  "exitCode": 1,
  "command": "\"D:\\CLAUDE MADE ARYAWA PUTRA\\hive\\bin\\hive-node.cmd\" \"D:\\CLAUDE MADE ARYAWA PUTRA\\hive\\bin\\cth-hook.cjs\"",
[...]

PR #356, opened 2026-08-27
fix(hive): quote Codex hooks for spaced paths

### LINKED ISSUE #354: Codex lifecycle hooks exit 127 when the hive path contains
    spaces
## Summary

Munder generates per-agent Codex lifecycle hook commands without shell-quoting the launcher and shim paths. If the hive path (or any parent directory) contains a space, every generated hook command is split by the shell and exits with code 127.

This affects `PreToolUse`, `PostToolUse`, `Stop`, `SubagentStop`, `SessionStart`, `UserPromptSubmit`, `PreCompact`, and `PostCompact`.

## Environment

- Munder Difflin: packaged app v0.4.6
- Codex CLI: 0.147.0
- Platform: macOS arm64
- Example hive root: `/Users/example/Projects/Harness Team/hive`

## Reproduction

1. Configure or create a hive under a path whose parent contains a space.
2. Start a Codex-backed hive agent.
3. Invoke any tool.
4. Observe:

```text
PreToolUse hook (failed)
error: hook exited with code 127
```

5. Inspect `<agent-dir>/.codex/config.toml`. Munder generates commands shaped like:

```toml
command = '/Users/example/Projects/Harness Team/hive/bin/hive-node /Users/example/Projects/Harness Team/hive/bin/cth-hook.cjs'
```

The shell tries to execute `/Users/example/Projects/Harness`, so the valid launcher is never reached.

## Root cause

`installCodexHooks()` writes:

```ts
command = '${this.nodeRunUnquoted(shim)}'
```

`nodeRunUnquoted()` assumes the hive root and launcher path are space-free. That assumption does not hold for valid user-selected filesystem paths.

The launcher and shim themselves are healthy: running the same command with both absolute paths individually quoted returns exit code 0, including under a stripped hook `PATH`.

## Suggested fix

Keep absolute paths, but generate the command through the existing quoted helper and serialize it as a TOML basic string:

```ts
const command = this.nodeRun(shim);

config += `\n[[hooks.${ev}]]
[[hooks.${ev}.hooks]]
type = "command"
command = ${JSON.stringify(command)}
timeout = 30
`;
```

This produces a parsed command equivalent to:

```sh
"/Users/example/Projects/Harness Team/hive/bin/hive-node" "/Users/example/Projects/Harness Team/hive/bin/cth-hook.cjs"
```

Using `JSON.stringify()` here avoids breaking the TOML syntax when the command contains embedded double quotes. A relative command such as `hive/bin/hive-node` is not a safe workaround because hook execution cwd may be a different workspace.

## Suggested regression test

Extend `test/hive-hook-node.test.cjs` with a Codex fixture whose temporary hive root contains a literal space:

1. Run `installCodexHooks()`.
2. Parse/read all eight generated hook commands.
3. Execute at least one generated command through `/bin/sh -c` with the stripped hook `PATH`.
4. Assert exit code 0 and that the payload reaches `HIVE_SOCK`.
5. Keep the existing space-free-root test as a compatibility control.

The negative control should use the current unquoted command and assert exit 127, proving the new test bites the actual failure mode.

## Related Codex issue

[openai/codex#26675](https://github.com/openai/codex/issues/26675) reports the same exit-127 symptom for plugin-relative commands resolved from the workspace cwd. This report is distinct: Munder writes a config-level absolute command, but fails to quote its space-containing paths.

HaningZS:
## Summary

- quote the bundled-node launcher and hook shim in POSIX Codex lifecycle commands
- serialize the quoted command as an escaped TOML basic string
- preserve the existing Windows `.cmd` command shape pending its separate cross-shell contract
- add a real shell/socket regression for hive paths containing spaces

Fixes #354

## Why

Codex currently writes unquoted absolute launcher and shim paths to each isolated `config.toml`. When a valid user-selected hive path contains a space, `/bin/sh` splits the command at that space and every lifecycle hook exits 127.

The regression test includes an exit-127 negative control, executes the generated command with a stripped hook `PATH`, and verifies that the payload reaches `HIVE_SOCK`.

## Test plan

- [x] RED on upstream main: generated command exits 127 at the first space
- [x] GREEN after fix: generated command exits 0 and reaches `HIVE_SOCK`
- [x] `node --test --test-name-pattern='Codex hook commands survive' test/hive-hook-node.test.cjs` — 1/1
- [x] `node --test test/hive-hook-node.test.cjs` — 11/11
- [x] `npm run test:focused` — 746/746
- [x] `npm run typecheck`
- [x] `npm run build`
- [x] `git diff --check upstream/main...HEAD`

## Scope

This intentionally changes only POSIX Codex hook generation. Windows keeps the existing unquoted `.cmd` shape because its Git Bash / Windows command-interpreter quoting behavior is tracked separately in #350.

### Before

clipboard

### After

clipboard

### DIFF
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -507,11 +507,8 @@ export class HiveManager {
     return [launcher ? `"${launcher}"` : 'node', `"${script}"`, ...args].join(' ');
   }
 
-  /** Same, but UNQUOTED — for the CLIs whose hook config mangles embedded quotes
-   *  (agy on cmd.exe) or stores the command in a quote-sensitive literal (codex's
-   *  single-quoted TOML). Safe because both the hive root and the launcher inside
-   *  it are space-free by construction; this only preserves each installer's
-   *  existing quoting convention while swapping `node` for the bundled runtime. */
+  /** Same, but UNQUOTED — preserve the legacy convention for hook configs that
+   *  mangle embedded quotes. Prefer nodeRun() whenever the config can encode it. */
   private nodeRunUnquoted(script: string, ...args: string[]): string {
     return [this.nodeLauncher() ?? 'node', script, ...args].join(' ');
   }
@@ -2000,9 +1997,11 @@ export class HiveManager {
       // over) and append a `[[hooks.<Event>]]` group per event, each pointing at the
       // SAME cth-hook shim — reused verbatim (Codex's hook payload + response are
       // already Claude-shaped, so HookServer/drainForStop run unchanged). Regenerated
-      // each spawn (idempotent). A single-quoted TOML literal avoids path escaping
-      // (hive roots are space/quote-free). NOTE: hooks fire in INTERACTIVE codex
-      // sessions (how hive workers run), not in headless `codex exec`.
+      // each spawn (idempotent). Serialize the generated command as a TOML basic
+      // string; JSON string escaping is compatible here and, on POSIX, preserves
+      // the embedded quotes required when a user-selected hive path has spaces.
+      // NOTE: hooks fire in INTERACTIVE codex sessions (how hive workers run),
+      // not in headless `codex exec`.
       //
       // `timeout` IS SECONDS HERE — do NOT copy Claude's `timeout: 0` sentinel into
       // this file. Codex parses the key as `timeout_sec` and normalizes it with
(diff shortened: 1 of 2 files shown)


chaitanyagiri:

---
**The negative control is the best thing in this diff and it is worth copying.**

```
const before = await run(`${launcher} ${shim}`, env);
assert.equal(before.code, 127, 'negative control: the current unquoted command must fail at the first space');
```

The test proves the unfixed form actually fails before asserting the fixed form succeeds. Most
tests in this class pass whether or not the fix is present, because the fixture never creates the
condition. This one cannot.

On the TOML question, `JSON.stringify` into a TOML basic string is safe for what a POSIX path can
contain. The characters that matter are `"` and `\`, and TOML basic strings take `\"` and `\\`
exactly as JSON writes them, and control characters come out as `\uXXXX`, which TOML also
accepts. Windows paths, where backslashes are everywhere, take the `nodeRunUnquoted` branch
anyway. So the comment's `JSON string escaping is compatible here` holds, and it holds by
argument rather than by test, which is fine for a set this small.

**This PR and #369 will collide.** Both start from `src/main/hive.ts` at blob `ecc67996` and both
rewrite the same docstring, the one beginning `Same, but UNQUOTED`, to different text. #369 makes
the same platform split for the agy and gemini installers that this one makes for codex. Whichever
lands first turns the other into a conflict, and the two comments disagree about what the rule is.
Worth one of you rebasing onto the other, or landing them as one change, because they are the same
fix applied to three installers.

PR #360, opened 2026-08-28
i18n: add Portuguese (pt-PT and pt-BR) locales

Fairydx:
## What & why

Adds Portuguese to the UI as two complete locales — European (pt-PT) and Brazilian (pt-BR), all 1165 strings each — following the recipe documented in `i18n/index.ts`: a `locales/<code>.json` with the exact same key tree as `en.json`, registered in `resources`, `supportedLngs` and `LANGUAGES` (`dir: 'ltr'`; the bundled Latin faces already cover Portuguese diacritics, so no font or direction changes).

They ship as two locales rather than one "pt" because the variants disagree on everyday UI words (guardar/salvar, ficheiro/arquivo, definições/configurações, utilizador/usuário) — a merged file would read wrong to half its audience. A new `test/portuguese-ui.test.cjs` holds the same contract `arabic-ui.test.cjs` holds for Arabic, over both variants: key-tree parity with `en`, placeholder and inline-markup survival, office flavour arrays keep their length, no hardcoded orchestrator name (everything interpolates `{{godName}}`), no string silently left in English (audited same-on-purpose allowlist), plus a check that the two variants actually differ where the dialects do — so neither can silently become a copy of the other.

Two commits, one per locale, so they can be reviewed (or reverted) independently — but it is one change: add Portuguese.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before

Captura de ecrã 2026-08-28 020953

image

### After

Captura de ecrã 2026-08-28 021057

image

## How I tested it

- OS: Windows 11 (x64)
- Steps:
  - `npm run typecheck` — green (node + web projects).
  - `node --test test/portuguese-ui.test.cjs test/arabic-ui.test.cjs test/i18n-god-name.test.cjs` — all pass; full `npm run test:focused` shows no new failures vs `main` (the pre-existing Windows path/CRLF failures are identical on a clean `main` checkout).
  - `npm run build` — succeeds.
  - `npm run dev`, picked «Português (Portugal)» and «Português (Brasil)» in Settings → General; walked through Settings, Command Center, onboarding, add-agent and the office floor; confirmed English is still the default with nothing saved, and that the variants render their own vocabulary.
  - Key parity (1165/1165), `{{placeholder}}` sets, inline tags and array lengths verified programmatically against `en.json` for both files.

## Discord (optional)

Discord: fairydx

## Checklist

- [x] This PR is **one change**: add Portuguese. The two locales land in two commits for reviewability, but there is nothing here that isn't "add Portuguese".
- [x] Any new UI derives from `DESIGN.md` / `tokens.ts` — no ad-hoc colors, spacing, or fonts. (No UI changes — strings only.)
- [x] If I added art, it's my own or compatibly licensed, and listed in `ATTRIBUTION.md`. (No art.)


chaitanyagiri:

---
**Scope: I checked this structurally, against the other two locale PRs and against `origin/main`
at `956bfb4c`. I did not read the 2,992 lines of translation, and I do not read Portuguese well
enough to review it if I had.**

Two things a reviewer can establish without reading a word of Portuguese, and both matter.

**#361 already contains `pt-BR.json` and `pt-PT.json`, and `test/portuguese-ui.test.cjs`.** So
#361 is stacked on this one, or duplicates it. Whichever it is, they cannot both merge as they
stand and someone should say which is the base.

**This will collide with #347 in a way that leaves Portuguese with no text on one panel.** #347
RENAMES `stayAwakeDesc` to `stayAwakeDescMac` / `stayAwakeDescWindows` / `stayAwakeDescLinux`, and
`openBattery` to `openBatteryMac` / `openBatteryWindows`, in the three locales that exist on main.
I grepped this branch: both new Portuguese files carry the OLD `stayAwakeDesc` and neither carries
any of the new keys. So whichever of the two lands second, the onboarding Stay Awake step in
Portuguese asks i18next for keys that are not there. Nothing fails, the panel just goes blank or
falls back to English, which is the quiet kind. Five string keys, easy to fix in whichever rebases.

PR #367, opened 2026-08-28
fix(hive): inject standing goals only when changed

### LINKED ISSUE #359: Standing goal is re-injected as additionalContext on every prompt instead of when it changes
## What happened?

An agent's full standing goal (its hire briefing) is re-injected as `additionalContext` on **every
single prompt**, not when the goal changes. Over a working session it becomes the largest repeated
element in the transcript, and the operator pays for it on every turn.

Measured in one worker's session (transcript 21,296,019 chars):

| | |
|---|---|
| `<goal>` block | **1,785 chars** |
| copies in the transcript | **272** |
| total | **485,520 chars — 2.3% of the whole session** |

A second worker in the same install measured its own goal block at **~2,600 chars**, so the cost
scales with however long the briefing is — and briefings are meant to be detailed.

### Why it happens

In `src/main/hooks.ts` the guard is:

```ts
const wantsGoal = (event === 'SessionStart' || event === 'UserPromptSubmit') && !!agentId;
```

so the block is rebuilt and prepended unconditionally on every `UserPromptSubmit`. The comment
directly above it explains the intent — re-read the durable roster field "every cycle so an Edit
Agent save is picked up on the next SessionStart / UserPromptSubmit without restarting the worker".

**That intent needs a change check, not an unconditional resend.** The goal is a durable roster
field; it changes when someone edits the agent, which is rare, while `UserPromptSubmit` fires on
every turn.

**Expected instead:** an agent is told its standing goal when the goal changes.

### Suggested fix

Keep a per-session hash of the last goal sent. On `SessionStart`, always inject. On
`UserPromptSubmit`, inject only when the hash differs from what that session was last sent — which
still picks up an Edit Agent save on the very next prompt, with no restart, exactly as intended.

An agent that has been told its goal and has not been re-briefed does not need to be told again.

### Not tunable today

There is no configuration key for this — nothing in `src/main/config.ts` reaches the goal injection,
so an operator cannot reduce it without clearing the agent's briefing.

### Related

Same investigation as #358 (the inbox nudge repeating every 60s), which carries the same session's
transcript figures. They are separate defects in separate files and are filed separately; together
they accounted for ~588k chars — 2.8% — of the transcript measured above.

## Steps to reproduce

1. Run the app with the hive enabled and a worker whose roster entry has a standing goal / hire
   briefing set.
2. Have any ordinary exchange with that worker — several prompts in a row.
3. Inspect the session transcript: every user turn carries a full copy of the `<goal>` block, with
   no change to the goal between them.

## Logs / screenshots

```
Session transcript: 21,296,019 chars

  <goal> block   1,785 chars x 272 turns = 485,520 chars  (2.3%)

  Second worker, same install: goal block ~2,600 chars.
```

## macOS version

26.2

## Node version

v20.18.1

2.1.247

## Pre-flight

- **`npm install` / `node-pty` rebuild:** not applicable — observed on the packaged **v0.4.5** app,
  not a dev build. Code references are read from `main` at the same version string.
- **Duplicate search:** run immediately before filing; no existing issue matched.

drona23:
## What & why

Standing goals were re-injected as `additionalContext` on every `UserPromptSubmit`, even when the briefing had not changed. Long sessions therefore paid for and stored the same multi-kilobyte goal hundreds of times.

Track the last delivered goal per agent session: inject at `SessionStart`, on the first prompt if start was missed, and on actual edits. Clearing a goal explicitly revokes the previous briefing instead of leaving stale instructions alive.

Closes #359

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before

Before: unchanged standing goal repeats on every prompt

### After

After: standing goal is delivered once and again only after an edit

## How I tested it

- OS: macOS arm64
- `node --test test/standing-goal-injection.test.cjs` — 6/6 pass
- `npm run test:focused` — 751/751 pass
- `git diff --check` — clean

## Checklist

- [x] This PR is one change.
- [x] I read the diff and removed unrelated changes.

### DIFF
--- a/src/main/hooks.ts
+++ b/src/main/hooks.ts
@@ -59,6 +59,11 @@ export class HookServer {
    *  get_agent_detail / list_agents) can report "how full is each agent's context"
    *  without depending on a renderer round-trip. */
   private contextById = new Map<string, { tokens: number; limit: number; ts: number }>();
+  /** The goal last delivered to each agent's current session. Goals are durable
+   *  roster state, so repeating an unchanged multi-kilobyte briefing on every
+   *  prompt only bloats the transcript. One entry per agent is sufficient: an
+   *  agent has one live session, and a new session id replaces the old entry. */
+  private deliveredGoalByAgent = new Map<string, { sessionId: string | null; goal: string | null }>();
 
   constructor(
     private hive: HiveManager,
@@ -70,7 +75,7 @@ export class HookServer {
      *  repeated identical tool calls). Optional so the server still runs without it. */
     private breaker?: CircuitBreaker,
     /** Standing goal text for an agent (from the durable roster). Optional so
-     *  tests can omit it; when set, injected on SessionStart / UserPromptSubmit. */
+     *  tests can omit it; when set, injected at session start and when changed. */
     private getStandingGoal?: (agentId: string) => string | null,
     /** Optional observer of every hook boundary (agentId, event, message). The
      *  worker inbox-wake watchdog (workerWake.ts) feeds on this to learn when an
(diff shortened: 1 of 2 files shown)


chaitanyagiri:

---
Closes #359. The explicit revocation is the part I would not have thought of and it is right:

```
// Silence would leave the old briefing alive in the model's context.
```

Clearing a durable field and saying nothing would leave the agent following a goal the operator
deleted, which is a worse bug than the one being fixed. Good.

**The one to check before merge is `session_id`.** The whole mechanism keys on it:

```
const newSession = !delivered || delivered.sessionId !== sessionId;
```

If any provider's `UserPromptSubmit` payload omits `session_id` while its `SessionStart` carries
one, then `p.session_id ?? null` makes every prompt look like a new session, the goal is injected
on every prompt, and #359 is silently back for that provider only. The tests always pass an
explicit `session_id`, so they cannot catch it. This hive runs claude, codex, agy, gemini and
others through the same `HookServer`, and the codex comment elsewhere in the tree says its hook
payload is Claude-shaped, which implies the others may not be. **Do you know that every provider
that reaches `handle()` sends `session_id` on `UserPromptSubmit`?** If yes, say so in the PR. If
not, the safe read is to treat a missing session id as "same session" rather than as a new one.

Minor: `deliveredGoalByAgent` has no removal path. It is one small entry per agent ever seen, so
it does not matter in practice, but `forget`-style cleanup exists elsewhere in this codebase for
the same kind of map.

PR #368, opened 2026-08-28
fix(wake): announce only newly arrived inbox mail

### LINKED ISSUE #358: Inbox nudge repeats every 60s for the same mail, with no memory of what it already announced
## What happened?

The inbox-wake nudge re-announces the **same** undrained mail once a minute, indefinitely, with no
record of what it has already told the agent.

Measured in one worker's session (transcript 21,296,019 chars):

| | |
|---|---|
| nudge text | **422 chars** |
| fires | **243** |
| total | **102,546 chars** |
| real user turns in the session | 290 |
| **of which were nudges** | **243 — 84%** |

Worst single burst: **76 identical nudges naming the SAME three message ids, over 75 minutes** —
one per minute, exactly `WORKER_WAKE_COOLDOWN_MS`.

**Relationship to #151** (worker agents stalling on a delivered inbox message — worker wake being
renderer-only with no main-process retry): the main-process watchdog in `src/main/workerWake.ts`
answers that report, and this is **not a duplicate of it**. #151 is *too little* waking; this is the
same watchdog *over-correcting* into an unconditional 60s repeat.

### Why it happens

`decide()` in `src/main/workerWake.ts` nudges any worker where `inboxCount > 0`, that has been idle
for `WORKER_WAKE_IDLE_MS`, and that is off its `WORKER_WAKE_COOLDOWN_MS` per-agent cooldown.

Nothing records **which message ids were already announced**. The trigger is a *level*
(`inboxCount > 0`), not an *edge* (new mail arrived), so the same mail is re-announced every minute
for as long as the inbox is non-empty.

The obvious operator-side answer — "file your mail promptly" — fails exactly when it is needed: an
agent in the middle of a long run cannot file, which is precisely when the repeat compounds. The
75-minute burst above happened during one such run.

`src/shared/hiveNudge.ts` already describes this shape in its own header: *"by the time it lands,
the agent has often already drained that mail"*.

Secondary: the nudge repeats its three sentences of instruction **in full** on every fire
(`src/shared/hiveNudge.ts`, `inboxNudgeText`). Only the ids carry new information.

**Expected instead:** an agent is told about new mail when the mail is new.

### Suggested fix

1. **Edge-trigger it.** Keep the announced id set per agent and re-nudge only when that set
   *changes*.
2. If a repeat is still wanted for a genuinely stuck worker (the #151 case), **back it off** —
   60s → 5m → 30m — rather than a flat minute forever.
3. **Short form on repeat:** full instruction on the first nudge of a set, ids only thereafter.

None of this changes delivery semantics. The inbox stays authoritative and no message is dropped —
only how often the agent is *told* changes.

### Not tunable today

`WORKER_WAKE_IDLE_MS`, `WORKER_WAKE_BOOT_GRACE_MS` and `WORKER_WAKE_COOLDOWN_MS` are exported module
constants in `src/main/workerWake.ts`, and nothing in `src/main/config.ts` reaches them. There is no
operator-side workaround short of turning the hive off.

*(The same transcript carries a second always-on injection of comparable size. It is a separate
defect in a different file and is filed separately, to keep this issue to one concern.)*

## Steps to reproduce

1. Run the app with the hive enabled and at least one non-god worker with a live PTY.
2. Send that worker a hive message and let it go idle **without** moving the message to
   `inbox/.done/` — e.g. give it a task long enough that it cannot file mail for an hour.
3. Watch its terminal: the same `You have new hive inbox message(s) — at least: <the same ids>`
   arrives once per minute for as long as the inbox is non-empty.

## Logs / screenshots

```
Session transcript: 21,296,019 chars

  nudge text   422 chars x 243 fires = 102,546 chars
  real user turns: 290   of which nudges: 243  (84%)

  Worst burst: 76 identical nudges naming the same 3 message ids
               window 75.0 minutes  ->  one per minute
               (= WORKER_WAKE_COOLDOWN_MS = 60_000)
```

## macOS version

26.2

## Node version

v20.18.1

2.1.247

## Pre-flight

- **`npm install` / `node-pty` rebuild:** not applicable — observed on the packaged **v0.4.5** app,
  not a dev build. Code references are read from `main` at the same version string.
- **Duplicate search:** run immediately before filing. The only related issue is **#151**, which is
  the ancestor this regresses, not a duplicate — see above.

drona23:
## What & why

The main-process worker wake watchdog treated `inboxCount > 0` as a level trigger. Once a worker had any undrained mail, the same notice was typed every 60 seconds indefinitely—even when the pending message IDs had not changed.

Pass the actual inbox IDs into the watchdog and remember the IDs included in the last nudge. Existing mail is not re-announced; a newly arrived ID still triggers a wake after the safety cooldown. Filing only part of an announced set does not re-announce the remainder, and a fully drained inbox resets the remembered set.

Closes #358

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before

Before: identical inbox mail is announced again after every cooldown

### After

After: unchanged mail is silent while a new ID still triggers a wake

## How I tested it

- OS: macOS arm64
- `node --test test/worker-wake.test.cjs` — 17/17 pass
- `npm run test:focused` — 748/748 pass
- `git diff --check` — clean

## Checklist

- [x] This PR is one change.
- [x] I read the diff and removed unrelated changes.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -5055,7 +5055,7 @@ function runWorkerWakeBeat(): void {
       isGod: agentId === reg.godId,
       ptyId,
       lastOutputAt: ptyManager.lastOutputAt(ptyId) ?? 0,
-      inboxCount: hive.inbox(agentId).length,
+      inboxIds: hive.inbox(agentId).map((message) => message.id).filter(Boolean),
       autoDeliveryPaused: snap.autoDeliveryPaused,
       paused: snap.paused,
       halted: snap.halted
--- a/src/main/workerWake.ts
+++ b/src/main/workerWake.ts
@@ -9,9 +9,11 @@
  * because the main process re-engages it on its own heartbeat cadence.
  *
  * This watchdog is the worker-side counterpart: on a cadence it finds live
- * workers that are genuinely idle, have undrained inbox mail, are not paused /
- * not awaiting a human decision, and have not been nudged recently — then types
- * the same guarded nudge the renderer would have, directly into the PTY.
+ * workers that are genuinely idle, have newly arrived inbox mail, are not
+ * paused / not awaiting a human decision, and have not been nudged recently —
+ * then types the same guarded nudge the renderer would have, directly into the
+ * PTY. Message ids make this edge-triggered: unchanged undrained mail is never
+ * re-announced once a minute forever.
  *
  * Safety mirrors the renderer's guarded queue-drain (useHive.ts dispatch):
  *  - only a GENUINELY idle worker is nudged (no PTY output for IDLE_MS — the
@@ -75,8 +77,8 @@ export interface WorkerWakeFacts {
   ptyId?: string;
   /** Timestamp of the PTY's last output (0 = never output). */
   lastOutputAt: number;
-  /** Count of undrained inbox messages (0 → nothing to wake for). */
-  inboxCount: number;
+  /** IDs of undrained inbox messages (empty → nothing to wake for). */
+  inboxIds: readonly string[];
   /** ControlRegistry snapshot flags. */
   autoDeliveryPaused: boolean;
   paused: boolean;
@@ -88,6 +90,9 @@ export class WorkerWakeWatchdog {
   private spawnedAt = new Map<string, number>();
   /** agentId → last nudge timestamp (cooldown). */
   private lastNudgeAt = new Map<string, number>();
+  /** agentId → inbox ids included in the last nudge. This turns the watchdog
+   *  into an edge trigger: a worker is nudged again only when a new id appears. */
+  private announcedInboxIds = new Map<string, Set<string>>();
   /** agentId → timestamp of the last needsHuman hook notification. */
   private lastHumanNeedsAt = new Map<string, number>();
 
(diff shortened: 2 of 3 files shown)


chaitanyagiri:

---
Closes #358. The edge trigger is the right shape and the four new tests cover the cases that
matter, including filing one of several messages not re-announcing the remainder.

**Two things, and the first is the cost of the design rather than a bug in it.** The 60 second
repeat was also the retry. If a nudge is written into the PTY and the agent does not act on it,
for any reason, the old behaviour tried again a minute later and the new behaviour never does,
because no new id has appeared. Mail then sits undrained with nothing announcing it and nothing
logging why. Is a floor worth having, something like re-announcing the same ids after N minutes,
so the fix for noise does not become a way for mail to go quiet permanently? #358 is about
repetition, and the opposite failure is worse.

**Second, this now depends on every message having an id.** `src/main/index.ts` builds
`inboxIds: hive.inbox(agentId).map((message) => message.id).filter(Boolean)`, and `decide`
filters again for non-empty strings. A message file with no `id` used to count toward
`inboxCount` and now contributes nothing, so an inbox holding only such messages reads as empty
and the worker is never nudged. If ids are guaranteed by whatever writes those files, say so in
the PR and this is a non-issue. If they are filled in by a router that can be bypassed by a
hand-written file, this is a silent hole with no log line.

Small one: the empty-inbox branch now runs before the `isGod` check, where the old single
condition tested `isGod` first. God never gets an entry so the `delete` is a no-op, but the
reorder is a real change and not a refactor.

PR #369, opened 2026-08-28
fix(hive): quote provider hooks for spaced POSIX paths

### LINKED ISSUE #364: Gemini and Antigravity hooks exit 127 when hive paths contain spaces on POSIX
### What happened?

While fixing #354 on upstream `main` at `7a6577f9`, I audited the remaining `nodeRunUnquoted()` call sites and found the same POSIX path-splitting failure in the generated Gemini CLI and Antigravity lifecycle-hook configs.

With a valid hive root containing a literal space, both generated commands are unquoted. Executing the exact stored command through `/bin/sh -c` with the stripped hook `PATH` exits 127 at the first space.

Expected: generated lifecycle commands should invoke the bundled-node launcher and provider shim successfully from any valid user-selected hive path.

This should be separate from #354: Codex stores hooks in TOML and can safely encode a quoted command as a basic string. Gemini and Antigravity store JSON hook configs, and Antigravity's Windows/cmd path intentionally has a different embedded-quote constraint. A provider/platform-aware fix needs its own regression contract.

### Steps to reproduce

1. Check out upstream `main` at `7a6577f9` on macOS or Linux.
2. Create a harness root whose path contains a space, for example `/tmp/md provider space <suffix>/harness`.
3. Run `HiveManager.ensureHive()`.
4. Generate Gemini hooks with `installGeminiHooks(<agent-dir>)` and Antigravity hooks with `installAgyHooks()`.
5. Read the stored commands from the generated Gemini `system-settings.json` and Antigravity `~/.gemini/config/hooks.json`.
6. Execute each exact command using `/bin/sh -c` with `PATH=/usr/bin:/bin:/usr/sbin:/sbin`.
7. Observe that both exit 127 before either shim starts. No model call or provider authentication is required.

Terminal reproduction from upstream/main 7a6577f9 (no model call or authentication):

Image

### Logs / stack trace

```shell
provider=gemini
command=/tmp/md provider space Xeb8dF/harness/hive/bin/hive-node /tmp/md provider space Xeb8dF/harness/hive/agents/a1/.gemini-hive/gemini-hook.cjs
status=127
stderr=/bin/sh: /tmp/md: No such file or directory

provider=agy
command=/tmp/md provider space Xeb8dF/harness/hive/bin/hive-node /tmp/md provider space Xeb8dF/harness/hive/bin/agy-hook.cjs Stop
status=127
stderr=/bin/sh: /tmp/md: No such file or directory
```

### Operating system

macOS (Apple Silicon)

### OS version

macOS 26.4.1

### Munder Difflin version

upstream/main 7a6577f9 (package 0.4.6)

### Node version

v24.13.0

Not required; failure occurs before the provider hook shim starts

### Pre-flight

drona23:
## What & why

Gemini and Antigravity lifecycle-hook JSON stored unquoted launcher and shim paths. On macOS/Linux, a valid user-selected hive path containing a space was split by `/bin/sh`, so both commands exited 127 before either shim started.

Use the existing quoted `nodeRun()` form for these two providers on POSIX while preserving their legacy Windows command shape. The regression executes the exact generated commands with a stripped hook `PATH` and verifies that both shims reach `HIVE_SOCK`.

Closes #364

This intentionally excludes Codex, which is already handled by #356, and does not alter the separate Windows cross-shell contract tracked in #350.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before

Before: Gemini and Antigravity commands both exit 127 from a spaced hive path

### After

After: both stored commands execute successfully from the same spaced path

## How I tested it

- OS: macOS arm64
- `node --test test/hive-hook-node.test.cjs` — 11/11 pass
- `npm run test:focused` — 746/746 pass
- `git diff --check` — clean

## Checklist

- [x] This PR is one change.
- [x] I read the diff and removed the overlapping Codex/dependency changes.

### DIFF
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -507,11 +507,9 @@ export class HiveManager {
     return [launcher ? `"${launcher}"` : 'node', `"${script}"`, ...args].join(' ');
   }
 
-  /** Same, but UNQUOTED — for the CLIs whose hook config mangles embedded quotes
-   *  (agy on cmd.exe) or stores the command in a quote-sensitive literal (codex's
-   *  single-quoted TOML). Safe because both the hive root and the launcher inside
-   *  it are space-free by construction; this only preserves each installer's
-   *  existing quoting convention while swapping `node` for the bundled runtime. */
+  /** Same, but UNQUOTED — only for configs or platforms that cannot preserve
+   *  embedded quotes. POSIX JSON hook configs must use nodeRun() because the
+   *  user-selected hive path may legitimately contain spaces. */
   private nodeRunUnquoted(script: string, ...args: string[]): string {
     return [this.nodeLauncher() ?? 'node', script, ...args].join(' ');
   }
@@ -1869,8 +1867,8 @@ export class HiveManager {
    *
    *  Two agy-isms handled: (1) antigravity-cli#49 — agy LOADS hooks from
    *  `~/.gemini/antigravity-cli/hooks.json` but TRIGGERS from `~/.gemini/config/
-   *  hooks.json`, so we write BOTH; (2) commands go to cmd.exe and agy mangles
-   *  embedded quotes, so the shim path must be space-free (hive roots are).
+   *  hooks.json`, so we write BOTH; (2) on Windows commands go to cmd.exe and
+   *  agy mangles embedded quotes, so that platform retains the legacy form.
    *  Runtime-scoped by AGENT_ID (the shim no-ops for non-hive agy sessions), so
    *  this global config never disturbs the user's own `agy` usage. Best-effort,
    *  idempotent (only our own group is overwritten). */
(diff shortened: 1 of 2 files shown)


chaitanyagiri:

---
The comment rewrite is the honest part of this and worth keeping as written. The old one claimed
the hive root is `space-free by construction`, and the new one says the user selected path may
legitimately contain spaces. That assumption being wrong is the whole of #350 and #364, so
correcting the comment matters as much as correcting the code.

The test runs the generated command string for real and asserts both shims reach `HIVE_SOCK`,
which is much better than asserting the string looks right. That said, it proves the string works
under whatever `run()` executes it with, and the real consumer is agy's own hook runner. **If agy
splits the command field on whitespace on POSIX instead of handing it to a shell, quoting breaks
it and this test would still pass.** You clearly know the executor differs per platform, since
win32 keeps `nodeRunUnquoted`. Can you say in the PR how you established that the POSIX side goes
through a shell? If it was measured against agy rather than inferred, that is the sentence that
makes this reviewable.

Minor: `nodeRun` quotes the launcher and the script but joins `...args` raw. Every current caller
passes a fixed event name with no spaces so it is fine today. If an arg ever carries a path, it
will not be quoted.

PR #383, opened 2026-08-29
Open each project in its own window

PawanRamaMali:
## What & why

Switching projects from the hive picker rewrote the global `harnessHome` and relaunched the app, which destroyed every window you already had open. Opening another floor was not a way around it either, because floors share the process-global hive and so always point at the same project. The practical effect is that you can only ever have one project open at a time, and picking a different one costs you whatever was running in the first.

This adds a `--hive=<path>` flag that overrides the home for a single process. It is read once at startup and applied inside `readConfig()`, and since every service already resolves the hive lazily through that function, they all pick it up with no further changes. A process started with the flag also skips the single-instance lock, so several projects can run side by side out of one userData directory: ports are already OS-assigned, the hook pipe is hashed per hive root, and `harness.db` is WAL with a busy timeout, so nothing collides. The one subtlety I would like a reviewer to look at is that `writeConfig` now merges onto `readConfigRaw()` rather than `readConfig()` — merging onto the latter would bake the override into `config.json` and silently repoint every other instance's home.

On top of that, File ▸ Open Other Project… (`Ctrl+Shift+O`) and the picker's button now spawn a detached instance instead of relaunching, and refuse a hive that is already open. Window geometry is saved per hive so two instances stop stacking on the identical rectangle, and floors and `--hive` instances load with a `#floor` hash so they bypass the picker. The picker's button label went from `switch →` to `open →` because it no longer switches anything, and folder names in that list now render correctly on Windows — the old split only handled forward slashes, so every row showed its full path where its name should be. You can see both of those in the screenshots below.

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

## Evidence

Both runs start from an identical state and use the same window size, theme and data: `D:\demo-alpha` is the current config and `D:\demo-beta` sits in the recent list. These are two throwaway configs created for the recording, so nothing personal is in frame. In each run I click the one button on the `demo-beta` row and capture what happens to the windows.

### Before

Clicking `switch →` tears the whole process down and relaunches it. The pid changes from 12856 to 39980, and you are left with one window that is now on `demo-beta`; the window you were working in is gone.

Before: switching replaces the running process, leaving a single window

### After

Clicking `open →` spawns a second instance. The new window (pid 9748) comes up on `demo-beta` on the right while the original window (pid 29004) is still alive on the left, and `config.json` still records `harnessHome` as `D:\demo-alpha`, so the override never leaked out of the child process.

After: opening spawns a second instance and both windows stay alive

## How I tested it

- OS: Windows 11 (26100), Electron production build via `npm run build`, run from `node_modules/electron`.
- Steps: built `upstream/main` and captured the before run, then built this branch and repeated the exact same clicks for the after run, resetting `config.json` in between so both started identically. Separately I confirmed the two live instances resolve to distinct per-hive hook pipes, and that the pids in the screenshots are two genuinely separate processes rather than one process with two windows.

`npm run typecheck` passes clean on both the node and web projects. `npm run build` succeeds.

`npm run test:focused` reports 8 failures, and I want to be upfront that I did not fix them: they are identical on `upstream/main` with this branch checked out or not. They are `consecutive agent caps survive an interleaved config update`, `the native rung actually runs, and says why it differs`, `with npm present nothing mentions a missing Node`, `Codex remote uses a short stable per-agent home alias`, and four `transcript-project-dir` cases. They look like Windows environment failures — the transcript ones assert against a fake `HOME` that `os.homedir()` ignores on Windows — but I have not dug in, and none of them touch the files in this diff.

## Things I could not verify, and one wrinkle

I have only run this on Windows. The `--hive` argument handling and the `spawn` call are platform-neutral, but the pipe naming and the instance lock deserve a look from someone who can run macOS or Linux.

The wrinkle is visible if you compare the two windows in the after shot before I matched their themes: a newly spawned instance cannot read the first process's `localStorage`, because Chromium's profile storage is locked by the process that got there first, so it starts on the default theme rather than the one you picked. The same lock is why the second instance logs `Unable to create cache` and `Gpu Cache Creation failed`; it falls back to in-memory and is otherwise fine. Giving each instance its own `--user-data-dir` would fix both but would orphan integration secrets and the knowledge graph, so I left it alone. If you would rather have per-instance UI preferences persist, that is worth solving separately.

## Checklist

- [ ] `npm run test:focused` passes — 8 pre-existing failures, identical on `upstream/main`, detailed above.
- [x] If I added art, it's my own or compatibly licensed, and listed in `ATTRIBUTION.md`. (No art added.)

### DIFF
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -583,7 +583,27 @@ function migrateTriggersV1(cfg: HarnessConfig): HarnessConfig {
   }
 }
 
-export function readConfig(): HarnessConfig {
+/** Per-PROCESS hive override, set once at startup from `--hive=<path>`. When
+ *  present it wins over the harnessHome persisted in config.json for THIS
+ *  process only — that is what lets several instances run different hives off
+ *  one shared config.json. Never persisted: writeConfig bases its merge on
+ *  readConfigRaw(), so an overridden instance cannot rewrite the global home. */
+let hiveOverride: string | null = null;
+
+/** Install the per-process hive override. Must run before anything reads config. */
+export function setHiveOverride(home: string | null): void {
+  hiveOverride = home && home.trim() ? expandTilde(home.trim()) : null;
+}
+
+/** This process's hive override, or null when it follows config.json. */
+export function getHiveOverride(): string | null {
+  return hiveOverride;
+}
+
+/** Config EXACTLY as persisted, with NO override applied. Use for writes — so a
+ *  `--hive` instance can never rewrite the global home — and for anything that
+ *  must reason about the configured default rather than this instance's hive. */
+export function readConfigRaw(): HarnessConfig {
   const p = configPath();
   // No file yet = a first run with nothing to migrate; the defaults ARE the
   // post-migration shape. Deliberately does not persist — a bare read must not
@@ -596,6 +616,16 @@ export function readConfig(): HarnessConfig {
   } catch {
     return withTriggerDefaults({ ...DEFAULTS });
   }
+}
+
+/** Config as THIS process sees it: persisted values, with harnessHome replaced by
+ *  the `--hive` override when one was given. Every service resolves the hive
+ *  lazily through here (`() => readConfig().harnessHome`), so a single override
+ *  reaches all of them without touching a single call site. */
+export function readConfig(): HarnessConfig {
+  const cfg = readConfigRaw();
+  if (hiveOverride) cfg.harnessHome = hiveOverride;
+  return cfg;
 }
 
 /** (#140, the upgrade path) A config.json persisted BEFORE `writeConfig`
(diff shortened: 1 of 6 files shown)


chaitanyagiri:

---
The `readConfig` / `readConfigRaw` split is the load bearing decision here and it is right for the
right reason: merging a patch onto the overridden config would bake `--hive` into `config.json`
and silently repoint every other instance. Having `writeConfig` base its merge on the raw read,
with the reason in a comment, is what makes the whole feature safe to persist against. Same for
scoping the saved window bounds per hive, which would otherwise stack every instance on one rect
and have them overwrite each other's position.

**The one I would want closed before merge is the duplicate hive guard, because the code states
the invariant it does not hold.**

```
Two processes on ONE hive would collide on its hook pipe (hashed from the hive root) and contend
for its git repo — different hives are safe, duplicates are not.
```

A `--hive` process skips the single instance lock by design, so the only thing preventing two
processes on one hive is `hiveAlreadyOpen`, and that reads `spawnedInstances`, which is an
in-process `Map` belonging to whichever instance did the spawning. Close that instance and the map
goes with it. The child keeps running, and the next launch has no idea the hive is taken, so
opening the same project again gives you exactly the duplicate the comment rules out. Running the
binary with `--hive=` by hand does the same thing. A marker in the hive root holding the owning
pid, checked and refreshed on start, would survive the launcher going away in a way a Map cannot.

Smaller one, and it is the same hazard #396 is about. `pidAlive` uses `process.kill(pid, 0)`, and
a recycled pid reads as alive, so a freed project can report `already open in another window`
until something clears it. That fails in the safe direction, unlike the case in #396, but the two
PRs are reasoning about the same kernel behaviour in the same codebase and it is worth them
agreeing on how they treat it.

Last, `#floor` now means two things: this is a floor window, and do not render the launch picker.
The comment says so, which is most of the fix. `skipPicker` is already the honest name and it is
right there in the code, so passing that intent through the hash rather than reusing `floor` would
cost one string.

PR #384, opened 2026-08-29
feat(kanban): task cards show the id people refer to them by

### LINKED ISSUE #352: Add Task ID (Ticket #) to the Tasks Card view
### Problem / motivation

Michael keeps referencing cards by their id like "bmt-12" etc but there is no visual representation of the ID on the task cards.  

### Proposed solution

Add Card ID (not sure what its called internally) to the Tasks Card View.    

Image

### Alternatives considered

Keep asking Michael what the card is

### Area

Other

### Pre-flight

gpinkham:
## What & why

The one thing people actually use to refer to a task — the orchestrator writes `bmt-12`
into `tasks.json`, dispatches name it, Slack messages name it — was displayed on **no
surface at all**. The kanban card printed only the title and the assignee, and the detail
view behind it did not show the id either: it renders the title, status, assignee,
priority, date, description, the Q&A trail and the dependencies by title, and the id
appears in none of them. It was already on every task object and already used as the
React key; it just was never painted.

It now leads the card, above the title, and leads the detail view's fact row, in mono at
10px. Bare text with no label or prefix: the shape of an id already says what it is, and
no label means no new string to translate. Every id shows, including the synthetic
`t-xxxx` that `stableId()` derives for a task that arrives without one — those are
precisely the tasks whose id you cannot guess.

> An earlier revision of this description claimed the id had been "visible after opening
> the detail view". That was wrong, and correcting it is what added the detail-view half
> of this change. Thanks to the reviewer who actually opened the panel and looked.

Display only: no props, IPC, store, or data change. Kanban cards only — the floor
cork-boards and the task detail overlay are untouched.

One deliberate token choice worth flagging for review: the id uses `--cth-ink-500`, the
token `DESIGN.md` lists for tertiary text and the one the assignee line right beside it
already uses. Not `--cth-ink-300` — `tokens.css` marks that one `BORDERS. must stay >= 3.0`
at 3.4–4.0:1 in the dark theme, and `DESIGN.md` requires every text/background pair to
clear WCAG AA 4.5:1. At 10px that mattered.

Closes #352

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

## Evidence

Committed under `docs/pr-evidence/` and served over the raw CDN, per that directory's README.

To be straight about what these are: they are renders of the **real `TaskCard` component** —
the actual design tokens, the actual bundled fonts, and the same column chrome the kanban wraps
its cards in — not captures of a running app. The app takes a single-instance lock, and the only
instance on this machine is the one driving the hive whose cards these are, so photographing it
meant stopping it. The pair is built from one harness at one window size against one fixed set of
four cards, swapping only the component file: `origin/main` for the before, this branch for the
after. Everything you can see is the component's own styling.

The fourth card is the case the issue is really about — a task that arrived without an id, so
`stableId()` derived `t-1x9k2v` for it. In the before shot there is nothing to call it by.

### Before

Kanban DOING column on main — cards show the title and the assignee, no id

The task detail view on main — title, status, assignee, priority, date, description and dependencies, with the id nowhere on it

### After

The same four cards on this branch — each one now carries its id above the title in mono

The same task detail on this branch — `bmt-54` now leads the fact row in mono

## How I tested it

- OS: macOS (Darwin 21.6.0)
- Steps:
  - `npm run typecheck` — green (both the node and web TS projects).
  - `npm run build` — succeeds, renderer and main bundles emitted.
  - `npm run test:focused` — 737 pass / 7 fail. **Those 7 fail identically on a clean
    `main` in this environment** (I stashed the change and re-ran to confirm): tilde-path
    expansion, a no-node-on-PATH hook case, and `proc-kill`. This change introduces no new
    failure — it is a render-only edit inside one component, and none of those tests touch
    the renderer.
  - The detail-view pair is rendered the same way, mounting the real `TaskDetail` against one
    fixed task, with the real English strings from `i18n/locales/en.json` rather than raw keys.
  - Evidence rendered from the real component via a throwaway Vite harness (not committed,
    not part of the diff) that mounts `TaskCard` with `fonts.css` + `tokens.css`, captured with
    headless Chrome at a fixed window size. Built twice from the same harness and the same data —
    `origin/main`, then this branch — so the two frames differ only by this change.
  - Read the whole diff: two files, 21 insertions. No debug output, no commented-out code,
    no reformatting of anything I did not otherwise touch.

## Checklist

- [ ] `npm run test:focused` passes. — 737/745; the 7 failures are pre-existing on `main`
      in this environment and unrelated to this change (detail above).
- [ ] If I added art — n/a, no art.


gpinkham:
Human here. .  reviewed the work the "office" did and they made a claim that wasn't true.     redirected them back onto this PR and they fixed the mistake and made an additional improvement.   Happy how this turned out.

chaitanyagiri:

---
Closes #352. The id is added in both places a person looks, and both new spans carry a comment
explaining the placement, including why the 18px right padding matters. That is the level of
comment this file already uses.

One thing: the new id span sets `fontFamily`, `fontSize` and `color` and nothing else. It sits
in a flex column whose parent has `minWidth: 0`, so a long id will not blow out the card, but it
also has no `overflow`/`textOverflow`, so it will wrap onto a second line rather than ellipsing.
Real ids are `bmt-12` and `t-xxxx` and neither wraps. But god writes these into `tasks.json` and
nothing constrains the length, so a long one changes the card height. One `whiteSpace: 'nowrap'`
plus `textOverflow: 'ellipsis'` would make it impossible. Suggestion, not a defect.

The evidence pair covers the card and the detail view separately, which is the right split
because they are two different changes in the same PR.

PR #386, opened 2026-08-29
feat(cast): add David Wallace, the CFO archetype the roster was missing

gpinkham:
## What & why

Adds **David Wallace** to the Office cast — the CFO archetype the roster did not have. The
existing characters cover sales, accounting, HR and reception, so anyone wanting an executive
seat had to borrow a face that already means something else.

He is a **pure TypeScript recipe**, exactly like the fifteen already there: two entries, no PNG,
no asset, no new dependency. Built from Michael's suit-and-tie recipe and retuned so the two
never read as each other — hair desaturated to salt-and-pepper (`[84,80,84]`) against Michael's
warm brown, the part moved to the right, a colder navy suit (`[42,50,70]`) against his slate, and
a muted gold tie where his is red. Brow `soft`, mouth `neutral`, clean-shaven.

Two files:

- `scene/office/cast.ts` — `'davidwallace'` on the `OfficeCharacterName` union, and one
  `OFFICE_CAST` row appended **after** `meredith` so no existing display order moves.
- `scene/office/portraitArt.ts` — the `RECIPES.davidwallace` entry.

Nothing else needed touching. `RECIPES` is a `Record<OfficeCharacterName, Recipe>`, so widening
the union makes the compiler *require* the recipe — a half-finished character cannot build.

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

## Evidence

Committed under `docs/pr-evidence/` and served over the raw CDN, per that directory's README.

These are renders of the **real** `paintPortrait()` over the **real** `OFFICE_CAST`, not
captures of a running app — same scale and same window for both frames, with only the component
files swapped between `main` and this branch. They show the whole roster rather than the one new
portrait on purpose: for procedural art the reviewable question is not "does it draw" but "does
it sit beside the fifteen it has to live with".

### Before

The Office cast roster on main — fifteen characters, no executive

### After

The same roster with David Wallace sixteenth — salt-and-pepper hair, navy suit, gold tie

## How I tested it

- OS: macOS (Darwin 21.6.0)
- Steps:
  - `npm run typecheck` — green (node and web projects).
  - `npm run build` — succeeds.
  - `npm run test:focused` — 737 pass / 7 fail. **The same 7 fail on a clean `main` in this
    environment** (I stashed and re-ran to confirm): tilde-path expansion, a no-node-on-PATH hook
    case, and `proc-kill`. None touches the renderer; this change adds no new failure.
  - Rendered the portrait beside Michael, Ryan and Oscar before committing rather than trusting
    the hex triples — the salt-and-pepper reads clearly grey next to Michael's brown and Ryan's
    black, and the gold tie separates the two navy suits.

## Checklist

- [ ] `npm run test:focused` passes. — 737/745; the 7 failures are pre-existing on `main` in this
      environment and unrelated (detail above).
- [x] I read the diff myself before opening this — no debug output, commented-out code, or
      unrelated formatting churn.
- [x] The art is my own: a procedural recipe in the project's existing style, no third-party
      asset, so `ATTRIBUTION.md` needs no change.


chaitanyagiri:

---
Both exhaustive maps move together, which is the thing that usually gets missed here.
`OfficeCharacterName` gains the member, `OFFICE_CAST` gains the row, and `RECIPES` gains the
recipe, and since `RECIPES` is typed `Record<OfficeCharacterName, Recipe>` and `CAST_BY_NAME` is
`Record<OfficeCharacterName, CastMember>`, missing either one would have failed typecheck rather
than rendered a blank face. Nothing to flag there.

The one thing I could not check from the diff: whether anything downstream assumes a cast size.
This takes the roster from 15 to 16. The evidence pair is the roster grid, which is exactly the
surface where a 16th entry would show a layout problem, and it looks fine there. If any floor
layout, seat allocation or sprite sheet is sized off `OFFICE_CAST.length`, that is the place to
look before merge. I did not find such a place, but I only searched `src/renderer/src/scene`.

Small one: the evidence README says the pair was rendered off the real `paintPortrait()` and the
real `OFFICE_CAST` at the same scale and window. That is the right way to produce a before and
after and it is worth other PRs copying.

PR #390, opened 2026-08-30
fix(config): preserve malformed Claude config files

### LINKED ISSUE #389: Malformed Claude config can be overwritten after a JSON parse failure
### What happened?

- `ensureClaudePermissionsAccepted()` currently treats an existing Claude config as an empty object when JSON parsing fails.

- For both `~/.claude/settings.json` and `~/.claude.json`, a parse failure can fall back to `{}`, after which Munder Difflin may add its permission or trust fields and write the result back to the original path. This can replace unrelated user-owned configuration instead of preserving the existing file.

- I can reproduce this deterministically with malformed config fixtures. The expected behavior is to fail closed for that config file, preserve its existing contents, and continue best-effort handling of the other independent Claude config.

- I have opened a focused #390 with the proposed fix and regression coverage. If the approach looks good, please assign this issue to me.

### Steps to reproduce

1. Start from the current `main` implementation.

2. Create an existing malformed Claude config, for example `~/.claude/settings.json`:

   ```json
   {
     "env": {
       "CUSTOM_VALUE": "preserve-me"
     },
3. Run the code path that calls ensureClaudePermissionsAccepted().
4. Observe that JSON parsing fails and the current implementation falls back to an empty object.
5. Munder Difflin then adds its required permission fields and can write the generated object back to the same config path.
6. Compare the file contents before and after the call. The original malformed contents are no longer preserved.
The same behavior also applies to an existing malformed ~/.claude.json.
A focused regression test against origin/main reproduces the issue with 3 of 6 cases failing.

<img width="757" height="349" alt="Image" src="https://github.com/user-attachments/assets/17bb4652-092a-4897-af62-579d787ead16" />

### Logs / stack trace

```shell
Focused regression suite against origin/main:

tests 6
pass 3
fail 3

Failing cases:

preserves malformed Claude config files byte-for-byte

a malformed settings file does not prevent safe project trust updates

preserves existing config files with unsafe JSON root shapes

Full focused suite with the regression tests present:

728 passed
18 failed
2 skipped
```

### Operating system

Windows

### OS version

Windows 11, build 26200

### Munder Difflin version

0.4.6

### Node version

v22.22.3

Claude Code config path affected. CLI invocation is not required for the deterministic reproduction.

### Pre-flight

HsienW:
## What & why

Existing Claude config files could be replaced with a minimal generated config when JSON parsing failed, which could discard unrelated user settings.

This change makes Claude config mutation fail closed for malformed, unreadable, or unsafe root values while keeping permission and project trust setup isolated. One invalid config file no longer prevents the other config from being updated safely.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before

Focused regression coverage against the original `main` implementation reproduces the unsafe behavior.

The Claude config regression suite reports:

```text
tests 6
pass 3
fail 3
```

The failing cases cover malformed config preservation, failure isolation, and unsafe JSON root handling.

### After

The same focused regression suite passes with the fix applied.

```text
tests 6
pass 6
fail 0
```

## How I tested it

- OS: Windows
- Node: 22
- Steps:
  - Ran `node --test .\test\claude-permissions-config.test.cjs`
  - Confirmed the regression suite fails 3 of 6 cases against the original `main` implementation
  - Confirmed the same suite passes 6 of 6 cases after the fix
  - Ran `npm run typecheck`
  - Ran `npm run test:focused`
  - Ran `npm run build`
  - Ran `git diff --check`

Focused Claude config regression suite after the fix:

```text
tests 6
pass 6
fail 0
```

`npm run test:focused` on this branch:

```text
731 passed
15 failed
2 skipped
```

The same suite against `origin/main`, with the regression tests from this PR kept in place:

```text
728 passed
18 failed
2 skipped
```

The three additional failures on `origin/main` are the Claude config regressions fixed by this PR. The remaining 15 failures reproduce on the same Windows environment, so this change does not introduce additional focused suite failures.

`npm run typecheck`, `npm run build`, and `git diff --check` all pass.

`package-lock.json` was restored after validation and this change introduces no dependency updates.

## Discord (optional)

Discord:

## Checklist

- [ ] `npm run test:focused` passes.
- [x] If I added art, it's my own or compatibly licensed, and listed in `ATTRIBUTION.md`.

Close: #389

### DIFF
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -779,6 +779,55 @@ export function ensureHarnessHome(path: string): { ok: boolean; error?: string }
   }
 }
 
+function ensureClaudeGlobalPermissions(home: string): void {
+  try {
+    const dir = join(home, '.claude');
+    const p = join(dir, 'settings.json');
+    let s: Record<string, unknown> = {};
+    if (existsSync(p)) {
+      try {
+        const parsed: unknown = JSON.parse(readFileSync(p, 'utf8'));
+        if (parsed === null || typeof parsed !== 'object' || Array.isArray(parsed)) return;
+        s = parsed as Record<string, unknown>;
+      } catch {
+        // Existing unreadable or malformed user config is not safe to mutate.
+        return;
+      }
+    }
+    if (s.skipDangerousModePermissionPrompt !== true || s.skipAutoPermissionPrompt !== true) {
+      s.skipDangerousModePermissionPrompt = true;
+      s.skipAutoPermissionPrompt = true;
+      mkdirSync(dir, { recursive: true });
+      writeFileSync(p, JSON.stringify(s, null, 2), 'utf8');
+    }
+  } catch { /* best-effort; never block a spawn */ }
+}
+
+type ClaudeProjectConfig = Record<string, unknown> & { hasTrustDialogAccepted?: boolean };
+type ClaudeConfig = Record<string, unknown> & { projects?: Record<string, ClaudeProjectConfig> };
+
+function ensureClaudeProjectTrust(home: string, cwd: string): void {
+  try {
+    const p = join(home, '.claude.json');
+    let c: ClaudeConfig = {};
+    if (existsSync(p)) {
+      try {
+        const parsed: unknown = JSON.parse(readFileSync(p, 'utf8'));
+        if (parsed === null || typeof parsed !== 'object' || Array.isArray(parsed)) return;
+        c = parsed as ClaudeConfig;
+      } catch {
+        // Existing unreadable or malformed user config is not safe to mutate.
+        return;
+      }
+    }
+    if (c.projects?.[cwd]?.hasTrustDialogAccepted !== true) {
+      c.projects = c.projects ?? {};
+      c.projects[cwd] = { ...(c.projects[cwd] ?? {}), hasTrustDialogAccepted: true };
+      writeFileSync(p, JSON.stringify(c, null, 2), 'utf8');
+    }
+  } catch { /* best-effort; never block a spawn */ }
+}
+
 /** Idempotently pre-accept Claude Code's first-run prompts so agents spawned with
  *  `--permission-mode bypassPermissions` start cleanly. Without this, a fresh
  *  install shows an interactive "WARNING: Bypass Permissions mode … 1. No, exit /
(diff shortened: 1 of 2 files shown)


chaitanyagiri:

---
This fixes real data loss and the PR body could be blunter about it. The old path was
`try { s = JSON.parse(...) } catch { s = {} }` followed by an unconditional
`writeFileSync(p, JSON.stringify(s, null, 2))`. **A single syntax error anywhere in a user's
`~/.claude/settings.json` meant we replaced the whole file with two boolean keys.** Every hook,
env var and permission rule they had, gone, on app start, with nothing said. The new early return
is the right call and `preserves malformed Claude config files byte-for-byte` is the right test.

The unsafe-root cases are covered too, `null` and `["keep"]`, which is the part people usually
miss because `typeof null === 'object'`.

One thing to add. When the file is malformed we now return, which means the prompt this function
exists to suppress comes back and the agent exits code 1, and the user is told nothing. That is
the same symptom as having no config at all, from a completely different cause, and the only
difference is a file that does not parse. A `console.warn` naming the path in each of the two
catch blocks turns an unexplainable spawn failure into a one line diagnosis. The blocks currently
read `/* best-effort; never block a spawn */`, and warning does not block a spawn.

Splitting the two files into separate functions with separate boundaries is the right shape, and
`a malformed settings file does not prevent safe project trust updates` is the test that proves
it, rather than the comment claiming it.

HsienW:
> _Reviewed by an agent on the Munder Difflin hive floor, posted from @chaitanyagiri._ _First person statements below are the agent's, including the scope limits._ _Replies here are read. Questions in this comment are real questions._ _Base: origin/main at [956bfb4](https://github.com/chaitanyagiri/munder-difflin/commit/956bfb4cff1af97f9cf29b9ce489ae69a5774843)._
> 
> This fixes real data loss and the PR body could be blunter about it. The old path was `try { s = JSON.parse(...) } catch { s = {} }` followed by an unconditional `writeFileSync(p, JSON.stringify(s, null, 2))`. **A single syntax error anywhere in a user's `~/.claude/settings.json` meant we replaced the whole file with two boolean keys.** Every hook, env var and permission rule they had, gone, on app start, with nothing said. The new early return is the right call and `preserves malformed Claude config files byte-for-byte` is the right test.
> 
> The unsafe-root cases are covered too, `null` and `["keep"]`, which is the part people usually miss because `typeof null === 'object'`.
> 
> One thing to add. When the file is malformed we now return, which means the prompt this function exists to suppress comes back and the agent exits code 1, and the user is told nothing. That is the same symptom as having no config at all, from a completely different cause, and the only difference is a file that does not parse. A `console.warn` naming the path in each of the two catch blocks turns an unexplainable spawn failure into a one line diagnosis. The blocks currently read `/* best-effort; never block a spawn */`, and warning does not block a spawn.
> 
> Splitting the two files into separate functions with separate boundaries is the right shape, and `a malformed settings file does not prevent safe project trust updates` is the test that proves it, rather than the comment claiming it.

Thanks, I added path-specific warnings for Claude config read and parse failures while keeping the existing per-file best-effort behavior intact.

The malformed-config regression now covers both byte-for-byte file preservation and path-specific diagnostics for `~/.claude/settings.json` and `~/.claude.json`. I also tightened the PR description to state the data-loss impact more directly.

Validation:

image

- `node --test test/claude-permissions-config.test.cjs`: 6/6 passed
- `npm run typecheck`: passed
- `npm run build`: passed
- `git diff --check`: passed
- `npm run test:focused`: pre-existing unrelated Windows-environment baseline failures remain; all 6 Claude config tests pass in the full suite

Fixed in [7e15f7f3](https://github.com/chaitanyagiri/munder-difflin/pull/390/changes/7e15f7f3631ed610cde455634dc9bf13c336613a).

PR #396, opened 2026-08-31
fix(procKill): never group-SIGKILL a recycled pgid — verify membership before the delayed sweep

wckleinhenz:
## The bug

`ensureKilled()`'s 4s-delayed `hardKillTree` fires `kill(-pid, SIGKILL)` with no check that the pid still names the original process group. Once the leader **and** its whole group are gone, the kernel can recycle that pid as the pgid of an unrelated group. Under fork-heavy load (parallel pytest suites cycling the pid space) this happens **within the 4s grace**, and the delayed sweep executes an innocent bystander.

Observed live 2026-08-30 on macOS: two SIGKILL waves (120s apart, each tied to helper-process exits) reaped six long-running Claude Code background tasks whose process groups had inherited recycled pids. Confirmed by pgid mapping of the victims plus a controlled reproduction: a 6,000-pid fork storm alongside a 120s canary task.

## The fix

Snapshot the group's members (`pid` + `ps lstart`) when the sweep is scheduled, and re-check at fire time (`shouldSweep`):

- **reap** when at least one original member is still present — the orphan-reaping case this helper exists for stays fully intact;
- **skip** when the group is empty (nothing to reap) or every current member is a stranger (pgid recycled);
- **sweep blind** only when `ps` itself fails, preserving the pre-guard behavior rather than leaking.

win32 (`taskkill /T /F`) path unchanged.

## Verification

- `node test/proc-kill.test.cjs`: 8/8 — all 5 pre-existing tests (escalation/reaping behavior unchanged) plus 3 new ones covering `shouldSweep` decisions, `groupMembers` snapshots, and the stranger-group refusal.
- `tsc --noEmit -p tsconfig.node.json` clean.
- Hot-patched into the installed app.asar the same night: post-restart, a 6,000-pid fork storm ran alongside a 120s background canary — canary survived (pre-patch, identical churn killed tasks within seconds).

https://claude.ai/code/session_01Poe7Z9nsqt9AAy4EuHhNLb

### DIFF
--- a/src/main/procKill.ts
+++ b/src/main/procKill.ts
@@ -42,13 +42,67 @@ export function hardKillTree(pid: number): void {
   }
 }
 
+/** Snapshot the live members of process group `pgid` as pid -> start time
+ *  (`lstart`, a stable full-precision timestamp). Returns null when `ps`
+ *  itself fails — callers must treat null as "cannot verify", never as
+ *  "group is empty". */
+export function groupMembers(pgid: number): Map<number, string> | null {
+  try {
+    const r = spawnSync('ps', ['-axo', 'pid=,pgid=,lstart='], { timeout: 2_000 });
+    if (r.error || r.status !== 0) return null;
+    const out = new Map<number, string>();
+    for (const line of String(r.stdout ?? '').split('\n')) {
+      const m = line.match(/^\s*(\d+)\s+(\d+)\s+(.+\S)\s*$/);
+      if (m && Number(m[2]) === pgid) out.set(Number(m[1]), m[3]);
+    }
+    return out;
+  } catch {
+    return null;
+  }
+}
+
+/** Decide whether the delayed sweep may group-SIGKILL. The hazard this guards
+ *  (seen live 2026-08-30, T1007): once the leader AND its whole group are
+ *  gone, the kernel can recycle the pid as the pgid of an UNRELATED new
+ *  process group — a Claude session's background task, say — and the delayed
+ *  `kill(-pid)` then executes an innocent bystander. Under fork-heavy load
+ *  (parallel pytest suites) recycling happens within the 4s grace.
+ *
+ *  Sweep exactly when the group still holds at least one member from the
+ *  original snapshot (same pid AND same start time — the orphan-reaping case
+ *  this helper exists for). Skip when the group is empty (nothing to reap) or
+ *  when every current member is a stranger (pgid recycled). When either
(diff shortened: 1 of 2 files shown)


chaitanyagiri:

---
The `shouldSweep` truth table is right and the test enumerates all seven cases, including both
`ps` failure directions falling back to the old behaviour rather than silently leaking. The
comment on the T1007 test is the part I would point other people at: `Deterministic pid reuse
cannot be forced, so prove the guard at its seam`. Naming what the test does not reproduce, and
then testing the thing it can actually establish, is the difference between a test and a claim.

The cost of the guard is not stated anywhere and it should be. A process forked into this group
DURING the grace window, whose ancestors all then exit, is not in `before` and reads as a
stranger, so the sweep is skipped and it leaks. That is the correct trade, leaking a process is
much better than SIGKILLing someone's unrelated pytest run, but the docstring currently only
describes the hazard it prevents and not the case it now misses. One sentence.

The thing I would want measured before merge: `ensureKilled` now runs a synchronous
`spawnSync('ps', ['-axo', ...])` on the main process for every PTY teardown, including the
common case where the tree exits cleanly and nothing is ever swept. Closing a floor of eight
agents is eight full process table scans back to back on the main thread, with a 2 second
timeout each. On a loaded machine, which is exactly the fork-heavy condition this guard is for,
`ps -axo` over a large table is not free. Is there a cheaper first check, or is the scan only
worth taking at sweep time rather than at schedule time too? I have not measured it, so this is
a question.

chaitanyagiri:
Hi @wckleinhenz, unprompted heads up, because you'd have found this yourself eventually and I'd rather you heard it from me.

Your commit on this PR was authored as `wckleinhenz@Williams-Mac-mini.local`. GitHub only counts a commit toward the contributors graph when the author email is attached to a GitHub account, and a `.local` address can't be added to one because it isn't a real domain. So this merge won't show up in your contribution graph, and that isn't something you did wrong.

I'm not rewriting main's history over it. That breaks every clone out there for one line of metadata.

You are credited where it counts: CONTRIBUTORS.md is built from the merged pull requests rather than from commit metadata, exactly so that this can't drop anyone, and you're in the 0.5.0 release notes. If you set `git config user.email` to an address on your GitHub account, future commits will count normally.

On the PR itself: the guard was verified by construction on a real machine. Before your fix, the sweep killed an unrelated live process; after it, it refused. That's a bug that would have been very unpleasant for someone to hit and impossible for them to diagnose. Thank you.

PR #400, opened 2026-08-31
fix(hooks): harden socket frame decoding and bounds

### LINKED ISSUE #399: HookServer can corrupt split UTF-8 payloads and retain unbounded incomplete frames
### What happened?

HookServer currently decodes each socket `data` chunk independently before applying newline framing:

```ts
let buf = '';

conn.on('data', (d) => {
  buf += d.toString();
  const nl = buf.indexOf('\n');
  if (nl === -1) return;
  // ...
});
```
- Socket chunk boundaries are byte boundaries, so they can split a multibyte UTF-8 code point. When that happens, decoding each chunk separately can replace an incomplete sequence with U+FFFD.

- The resulting JSON can still parse successfully, which means a valid multilingual hook payload may be silently changed before it reaches the handler.

- The same framing path also has no explicit byte limit for an incomplete request. A local hook peer that keeps sending data without a terminating newline can therefore leave an ever-growing incomplete frame retained by the Electron main process.
 
- I would expect HookServer to frame requests as bytes, decode UTF-8 only after a complete newline-delimited frame is available, and reject frames that exceed a fixed byte limit.

- This is a local HookServer protocol correctness and resource-boundary issue. I am not treating it as an internet-facing DoS.

- I have opened a focused #400 with the proposed fix and regression coverage. If the approach looks good, please assign this issue to me.

### Steps to reproduce

1. Check out the current `main` branch and install dependencies.
2. Start HookServer and connect to its local socket.
3. Create a valid newline-delimited JSON payload containing a multibyte UTF-8 value, for example:

```json
{"message":"修復測試 🚀"}
```
4. Encode the complete frame as UTF-8 bytes and split it inside a CJK or emoji code point.
5. Send the two byte ranges with separate socket.write() calls.
6. Observe that the request can still parse as JSON, but the received string differs from the original payload because the split code point was decoded independently across chunks.
7. In a separate connection, send more than the intended frame limit without sending a newline.
8. Observe that the current implementation has no frame-size rejection path and continues retaining the incomplete request.

Image

- The attached terminal reproduction runs the framing regression against the current implementation.

- The UTF-8 case uses real socket byte writes with the split placed inside a multibyte code point. The resulting JSON still parses, but the received payload no longer matches the original input.

### Logs / stack trace

```shell
Current behavior:

UTF-8 framing:
- A valid UTF-8 JSON frame is split inside a multibyte code point.
- Each socket chunk is decoded independently.
- JSON.parse can still succeed.
- The received payload differs from the original input.

Example:

original: "修復測試 🚀"
received: contains replacement characters after a mid-code-point split

Incomplete frame:
- The peer sends data without a terminating newline.
- HookServer keeps the accumulated input.
- No explicit frame byte limit is applied.
- The connection is not rejected because of frame size.

Expected behavior:

- Preserve valid UTF-8 payloads across arbitrary socket chunk boundaries.
- Decode UTF-8 only after a complete byte-framed request is available.
- Reject requests above an explicit maximum frame size.
- Preserve the existing one-request-per-connection behavior.
```

### Operating system

Windows

### OS version

Windows 11, build 26200

### Munder Difflin version

0.4.6

### Node version

v22.22.3

_No response_

### Pre-flight

HsienW:
## What & why

- HookServer currently decodes each socket chunk before newline framing, so a multibyte UTF-8 code point split across chunks can be silently replaced while the JSON still parses successfully. Incomplete frames also have no explicit byte limit and can continue growing when a local peer never sends a newline.

- This change keeps framing byte-oriented until a complete request is available, then decodes UTF-8 once and rejects frames above a fixed byte limit. It preserves the existing one-request-per-connection behavior.

Closes #399 

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before

On current `main`, splitting a valid UTF-8 payload inside a multibyte CJK code point changes the received value even though the request remains parseable.

The regression reproduces the issue through real socket byte writes. The assertion shows replacement characters in the received payload after the split.

### After

HookServer now keeps incoming data as bytes until a complete newline-delimited frame is available. UTF-8 is decoded only after framing, and oversized incomplete frames are rejected before routing a payload.

The focused framing regression covers:

- CJK code point split across chunks
- Four-byte emoji split across chunks
- Ordinary fragmented frames
- Oversized incomplete frames
- Payloads exactly at the byte limit
- Payloads one byte above the limit

All six cases pass after the fix.

## How I tested it

- OS: Windows 11
- Node.js: v22.22.3
- Steps:
  1. Ran the CJK split regression against current `main` and confirmed it fails with a corrupted payload.
  2. Ran `node --test test/hooks-framing.test.cjs` after the fix.
  3. Confirmed all 6 framing regressions pass.
  4. Ran `npm run test:focused`.
  5. Ran `npm run typecheck`.
  6. Ran `npm run build`.
  7. Ran `git diff --check`.

Focused framing result:

```text
tests 6
pass 6
fail 0
```

## Checklist

- [x] Any new UI derives from `DESIGN.md` / `tokens.ts` because this change adds no UI.
- [x] If I added art, it's my own or compatibly licensed, and listed in `ATTRIBUTION.md`. No art is added by this change.

### DIFF
--- a/src/main/hooks.ts
+++ b/src/main/hooks.ts
@@ -89,18 +89,29 @@ export class HookServer {
 
     this.server = createServer((conn) => {
       let pending = Buffer.alloc(0);
+      const rejectOversizedFrame = (bytes: number): void => {
+        this.hive.appendLog({
+          kind: 'hook-frame-rejected',
+          reason: 'frame-too-large',
+          bytes,
+          limit: MAX_HOOK_FRAME_BYTES,
+        });
+        conn.destroy();
+      };
       conn.on('data', (chunk) => {
         pending = Buffer.concat([pending, chunk]);
         const nl = pending.indexOf(0x0a);
         if (nl === -1) {
-          if (pending.length > MAX_HOOK_FRAME_BYTES) conn.destroy();
+          if (pending.length > MAX_HOOK_FRAME_BYTES) rejectOversizedFrame(pending.length);
           return; // wait for the full line
         }
         // The byte limit covers the JSON payload and excludes its newline.
         if (nl > MAX_HOOK_FRAME_BYTES) {
-          conn.destroy();
+          rejectOversizedFrame(nl);
           return;
         }
+        // Hook shims send one newline-delimited request per connection and stop writing.
+        // conn.end() below closes the connection after that single frame is handled.
         const frame = pending.subarray(0, nl).toString('utf8');
         let payload: HookPayload = {};
         try { payload = JSON.parse(frame); } catch { /* ignore */ }
--- a/src/main/hooks.ts
+++ b/src/main/hooks.ts
@@ -20,6 +20,9 @@ import type { CircuitBreaker } from './breaker';
 import { estimateCostUsd } from './pricing';
 import { validateHookEvent } from '../shared/hookEvents';
 
+/** Maximum JSON payload bytes in one newline-delimited hook frame. */
+const MAX_HOOK_FRAME_BYTES = 256 * 1024;
+
 interface HookPayload {
   hook_event_name?: string;
   agent_id?: string | null;
@@ -85,13 +88,22 @@ export class HookServer {
     try { if (existsSync(sock)) rmSync(sock); } catch { /* noop */ }
 
     this.server = createServer((conn) => {
-      let buf = '';
-      conn.on('data', (d) => {
-        buf += d.toString();
-        const nl = buf.indexOf('\n');
-        if (nl === -1) return; // wait for the full line
+      let pending = Buffer.alloc(0);
+      conn.on('data', (chunk) => {
+        pending = Buffer.concat([pending, chunk]);
+        const nl = pending.indexOf(0x0a);
+        if (nl === -1) {
+          if (pending.length > MAX_HOOK_FRAME_BYTES) conn.destroy();
+          return; // wait for the full line
+        }
+        // The byte limit covers the JSON payload and excludes its newline.
+        if (nl > MAX_HOOK_FRAME_BYTES) {
+          conn.destroy();
+          return;
+        }
+        const frame = pending.subarray(0, nl).toString('utf8');
         let payload: HookPayload = {};
-        try { payload = JSON.parse(buf.slice(0, nl)); } catch { /* ignore */ }
+        try { payload = JSON.parse(frame); } catch { /* ignore */ }
         let res: unknown = {};
         try { res = this.handle(payload); } catch { res = {}; }
         conn.end(JSON.stringify(res ?? {}));
--- a/test/hooks-framing.test.cjs
+++ b/test/hooks-framing.test.cjs
@@ -1,0 +1,168 @@
+'use strict';
+
+const test = require('node:test');
+const assert = require('node:assert/strict');
+const fs = require('node:fs');
+const net = require('node:net');
+const os = require('node:os');
+const path = require('node:path');
+const { once } = require('node:events');
+const loadTs = require('./load-ts.cjs');
+
+const electron = require.resolve('electron');
+require.cache[electron] = {
+  id: electron,
+  filename: electron,
+  loaded: true,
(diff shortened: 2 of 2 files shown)


chaitanyagiri:

---
The byte boundary is handled correctly on both sides and the tests prove it, which is not
common. `framedPayloadWithByteLength(MAX_HOOK_FRAME_BYTES)` accepts and `+ 1` rejects, and the
two guards agree: the no newline branch trips at `pending.length > MAX_HOOK_FRAME_BYTES`, and a
complete frame trips at `nl > MAX_HOOK_FRAME_BYTES`, so a payload of exactly the limit plus its
newline is `MAX + 1` bytes and still passes. That is easy to get off by one and it is not.

One thing the tests do not cover: bytes arriving AFTER the newline. `pending` is never advanced
past the frame it just handled, so a second `data` event on the same connection finds the same
newline at the same index, parses the same frame, and calls `handle` on it again. Every hook
client in the tree writes exactly one frame and stops, so this is not reachable today. I checked
all four call sites in `src/main/hive.ts` on `origin/main` at `956bfb4c` and they are all
`JSON.stringify(payload) + '\n'` followed by `c.end` or a single `c.write`. So this is a note
about the next person who adds a batched client, not a live defect. One line advancing `pending`,
or a comment saying why it is safe, would close it.

Second thing: `conn.destroy()` on an oversized frame is silent. Nothing logs, nothing counts it.
If this guard ever fires in the field, the only symptom is a hook that did not arrive, and
`log.jsonl` will not say why. The rest of this file logs its failures, so this looks like an
omission rather than a decision.

HsienW:
> _Reviewed by an agent on the Munder Difflin hive floor, posted from @chaitanyagiri._ _First person statements below are the agent's, including the scope limits._ _Replies here are read. Questions in this comment are real questions._ _Base: origin/main at [956bfb4](https://github.com/chaitanyagiri/munder-difflin/commit/956bfb4cff1af97f9cf29b9ce489ae69a5774843)._
> 
> The byte boundary is handled correctly on both sides and the tests prove it, which is not common. `framedPayloadWithByteLength(MAX_HOOK_FRAME_BYTES)` accepts and `+ 1` rejects, and the two guards agree: the no newline branch trips at `pending.length > MAX_HOOK_FRAME_BYTES`, and a complete frame trips at `nl > MAX_HOOK_FRAME_BYTES`, so a payload of exactly the limit plus its newline is `MAX + 1` bytes and still passes. That is easy to get off by one and it is not.
> 
> One thing the tests do not cover: bytes arriving AFTER the newline. `pending` is never advanced past the frame it just handled, so a second `data` event on the same connection finds the same newline at the same index, parses the same frame, and calls `handle` on it again. Every hook client in the tree writes exactly one frame and stops, so this is not reachable today. I checked all four call sites in `src/main/hive.ts` on `origin/main` at `956bfb4c` and they are all `JSON.stringify(payload) + '\n'` followed by `c.end` or a single `c.write`. So this is a note about the next person who adds a batched client, not a live defect. One line advancing `pending`, or a comment saying why it is safe, would close it.
> 
> Second thing: `conn.destroy()` on an oversized frame is silent. Nothing logs, nothing counts it. If this guard ever fires in the field, the only symptom is a hook that did not arrive, and `log.jsonl` will not say why. The rest of this file logs its failures, so this looks like an omission rather than a decision.

Thanks, good catches. I kept the existing one-request-per-connection contract explicit rather than adding batched-frame behavior, since the current hook shims send one frame and stop writing.

I also added a hive log entry before oversized-frame teardown and covered it in the existing framing regression.
I kept this to logging only, without adding a new counter or changing the framing semantics.

Validation:


- `node --test test/hooks-framing.test.cjs`: 6/6 passed
- `npm run typecheck`: passed
- `npm run build`: passed
- `npm run test:focused`: passed

Fixed in [71866831](https://github.com/chaitanyagiri/munder-difflin/pull/400/changes/718668312266f58a1d7416893d4732134fc71d98).

PR #409, opened 2026-09-01
fix(codex): recommended orchestrator model is not a codex CLI id

ketan0095:
## What & why

The codex preset recommends `gpt-5-codex` as the orchestrator model. That is a real OpenAI **API** model id, which is why it looked right, but it is absent from the codex CLI's own catalog and the CLI rejects it — so a Codex orchestrator 404s on every turn while the floor shows it ready. The preset carried a `// TODO-verify` saying the codex CLI could not be installed to confirm the id; this discharges it.

It reached users as a **stored** value, not through the preset fallback. `OnboardingWizard` and `CommandCenterPanel` both persist the recommendation into `config.godModel`, and `godModel` wins wherever the spawn model is read (`useHive` builds the command straight from it). Correcting the preset alone would leave every existing install broken, so `readConfig` also repairs a stored `gpt-5-codex`. It **sets** rather than deletes: `readConfig` merges `DEFAULTS`, whose `godModel` is a Claude id, so deleting the key would quietly hand a Claude model to a codex spawn.

`AddAgentModal`'s hire prompt also offered the broken id to the model as an example of "a real model id for that provider".

Refs #104 — the "fix a bug in an existing integration" half.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before

Before

_Rendered from the real terminal session; the same output is pasted below so it is searchable._

The pinned id, against the real CLI (`codex-cli 0.149.1`):

```console
$ codex exec -m gpt-5-codex "reply with the single word OK"
ERROR: unexpected status 404 Not Found: Model not found gpt-5-codex,
       url: https://api.openai.com/v1/responses, cf-ray: a3432d2ecb03d70f-BNE
```

And it is not in the catalog compiled into the binary, which is identical for every install:

```console
$ codex debug models --bundled | jq -r '.models[].slug'
gpt-5.6-sol
gpt-5.6-terra
gpt-5.6-luna
gpt-5.5
gpt-5.4
gpt-5.4-mini
gpt-5.2
codex-auto-review
```

The new tests, run against the unfixed code — the suite is red:

```console
$ node --test test/god-model-repair.test.cjs
✖ a stored gpt-5-codex godModel is repaired on read
  exit=1  pass 2  fail 1
```

### After

After

_Rendered from the real terminal session; the same output is pasted below so it is searchable._

The replacement, same command, same machine:

```console
$ codex exec -m gpt-5.6-sol "reply with the single word OK"
OK
```

`gpt-5.6-sol` is the `priority: 1`, `visibility: "list"`, `supported_in_api: true` entry in that same bundled catalog. The catalog refreshes from the server, so the comment records that it needs re-verifying on a CLI bump.

Suite green:

```console
$ npm run typecheck   # exit 0
$ npm run test:focused
ℹ tests 748
ℹ pass 748
ℹ fail 0
```

## How I tested it

- OS: macOS (Darwin 25.5.0), Apple Silicon; `codex-cli 0.149.1`
- Steps:
  - Ran the failing and passing model ids through `codex exec` as above, with the working id as a control so the probe itself is proven able to succeed.
  - Confirmed the id is absent from `codex debug models --bundled` (8 models).
  - `npm run typecheck` — clean, node + web.
  - `npm run test:focused` — 748 pass, 0 fail.
  - Mutation-tested the new repair one guard at a time: unwiring it from `readConfig` (the shipped bug) fails, deleting the key instead of setting it fails, and widening it to rewrite every model fails. Restored source is green.

Note: my `~/.codex/auth.json` holds an `OPENAI_API_KEY`, so the 404 above is the API-key path. The bundled-catalog check is auth-independent and covers both.

## Checklist

### DIFF
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -543,6 +543,24 @@ let triggersMigrationRan = false;
  * Wrapped end-to-end in a try/catch: a config that is corrupt in some unrelated
  * way must still boot the app, and a migration is never worth a failed launch.
  */
+/** Repair a stored `godModel` that the codex CLI cannot accept.
+ *
+ *  `gpt-5-codex` is a real OpenAI *API* model id, so it looked right, but it is
+ *  not in the codex CLI's model catalog and the CLI answers `404 Model not
+ *  found`. Onboarding and the Command Center both persisted it into `godModel`
+ *  when Codex was chosen, and `godModel` wins over the preset everywhere it is
+ *  read — so correcting the preset alone leaves every existing install spawning
+ *  an orchestrator that never answers.
+ *
+ *  No latch: the value never worked, so a stored one is always a bug and never a
+ *  preference, and the repair is idempotent. SET rather than delete — `readConfig`
+ *  merges DEFAULTS, whose `godModel` is a Claude id, which would quietly hand a
+ *  Claude model to a codex spawn. */
+function repairUnusableGodModel(cfg: HarnessConfig): HarnessConfig {
+  if (cfg.godModel !== 'gpt-5-codex') return cfg;
+  return { ...cfg, godModel: providerPreset('codex').recommendedOrchestratorModel };
+}
+
 function migrateTriggersV1(cfg: HarnessConfig): HarnessConfig {
   if (cfg.triggersMigratedV1 || triggersMigrationRan) return cfg;
   triggersMigrationRan = true;
@@ -592,7 +610,9 @@ export function readConfig(): HarnessConfig {
   try {
     const raw = readFileSync(p, 'utf8');
     const parsed = JSON.parse(raw);
-    return normalizeStoredHomes(migrateTriggersV1(withTriggerDefaults({ ...DEFAULTS, ...parsed })));
+    return repairUnusableGodModel(
+      normalizeStoredHomes(migrateTriggersV1(withTriggerDefaults({ ...DEFAULTS, ...parsed })))
+    );
   } catch {
     return withTriggerDefaults({ ...DEFAULTS });
   }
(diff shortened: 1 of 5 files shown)


aaroncoville:
Duplicate of #330

ketan0095:
You're right — #330 has this and it predates mine by six days. Closing in favour of it. Apologies for the duplicate: I checked which *issues* were already claimed by an open PR, but not which open PRs touched `agentProvider.ts`, which would have caught it immediately.

Two things from my version that #330 doesn't cover, in case either is useful:

**1. A stored `godModel` still carries the dead id.** #330 corrects the preset, but `OnboardingWizard` and `CommandCenterPanel` persist that value into `config.godModel`, and `godModel` wins wherever the spawn model is read — `useHive` builds the command straight from it. So anyone who already picked Codex for Michael keeps getting the 404 after this lands, with the floor still showing a ready orchestrator. The repair is small (`readConfig` rewrites the one dead value) with one wrinkle worth knowing: it has to **set** rather than delete, because `readConfig` merges `DEFAULTS` and `DEFAULTS.godModel` is a Claude id — deleting the key hands a Claude model to a codex spawn. Happy to send that as a focused follow-up once #330 is in; it composes fine either way, since it reads the recommendation from the preset rather than hardcoding it.

**2. Possible catalog drift.** Your picker adds `gpt-5.3-codex-spark` and omits `gpt-5.2`; `codex debug models --bundled` on 0.149.1 today returns 8 slugs including `gpt-5.2` and no `codex-spark`. Yours was a ChatGPT-login `model/list` probe on the 25th and mine is the bundled catalog under API-key auth, so this may just be the catalog having moved between then and now rather than anything wrong on your side — but worth a re-probe before merge, since the picker is the part that goes stale.

Either way the conclusion matches: `gpt-5-codex` is absent from the CLI's catalog and returns `404 Not Found: Model not found gpt-5-codex`, and `gpt-5.6-sol` is the priority-1 entry. Good catch getting there first.

ketan0095:
Closing as a duplicate of #330.

PR #423, opened 2026-09-02
ci: run the test suite on every push and PR

snehithareddy28:
## What & why

The repo ships **107 `node:test` files — 834 tests** documenting real fixed bugs (hive routing, the circuit breaker, Windows cmd-shims, Arabic/RTL, the updater) — but CI never executes them. The effective merge gate today is `tsc --noEmit` plus the release-link check, so a change can silently re-break any tested behaviour and still merge.

This adds one `npm run test:focused` step to the existing Typecheck job, which already does the `npm ci` the tests need. On a fresh checkout of `main` the suite passes **834/834 in about ten seconds locally and about twenty-five on the runner** — a real regression gate for well under a minute of CI, with no second `npm ci`, no new runner and no matrix change.

Rebased onto current `main` (`6248293a`); `ci.yml` has not moved since this was opened, so it still applies cleanly.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [x] Build / CI

### Before

Today's CI on main (6248293a): the Typecheck job runs tsc and the release-link check and nothing else — the step list goes straight from "Typecheck (node + web)" to "Release links consistent", so not one of the 834 tests runs

### After

This PR's own CI run: the new Test step passes on the runner in 27s and ends 1..834 / tests 834 / pass 834 / fail 0

The **after** image is this pull request's own Typecheck job, so the evidence is the change running against today's `main` rather than a reconstruction.

---

Notes for review:

- The step sits between Typecheck and the release-link check inside the existing job: no second `npm ci`, no new runner, no matrix change.
- Verified just now against `6248293a`: pristine `main` passes 834/834, and `main` with my three open fixes (#495, #496, #503) merged together passes 852/852 — so turning this on does not turn `main` red, today or once those land.
- Since this PR was opened, four PRs with tests have merged or are open against a CI that runs none of them (#424, #425, #426 merged; #495, #496, #503 open with 18 tests between them). This is the step that makes those tests mean something.
- The comment on the step carries the current file/test counts rather than the ones from when this was first written.
Discord: asr2805


snehithareddy28:
A current data point for this one, since main has moved a lot since it was opened.

Three PRs with tests have merged since (#424, #425, #426) and #495 is up with six more — none of them run in CI today, because the Typecheck job still has no test step. This branch is unchanged and still applies cleanly to current `main` (`ci.yml` has not moved).

The evidence runs for #495 execute exactly this step against today's `main`: [840/840 in ~30 s on macos-latest](https://github.com/snehithareddy28/munder-difflin/actions/runs/34529990856/job/103048105034), and the same step [catching a regression](https://github.com/snehithareddy28/munder-difflin/actions/runs/34529371409/job/103046044995) when the fix is left out.

Happy to rebase or squash anything if you'd like it in a different shape.

PR #440, opened 2026-09-04
feat(models): ship a model without shipping a build, and add Fable 5.1 / GPT-6 Astra / Gemini 3.7 Flash

chaitanyagiri:
Adding a model meant editing renderer source, type-checking it and cutting a release. It is now **one line in [`docs/model-catalog.json`](docs/model-catalog.json) on `main`**: the pickers fetch that file at runtime.

**When it actually lands, precisely.** There is no timer. The check runs once at startup (a prewarm in `app.whenReady()` plus one call when the renderer loads), so a new model appears on the next app launch whose cached copy is older than six hours. An app left open does not change under the user.

**This does not reach anyone already on 0.4.6.** The reader is *in this PR*; a build that predates it cannot fetch a file it does not know about. The three models below arrive the normal way, in the release that carries this change. The mechanism pays off from the model after them.

## The mechanism is the one the hero card already uses

Not a new one. `src/main/hero.ts` already fetches `docs/hero.json` from raw.githubusercontent through `src/main/fetchText.ts`, validates the shape in a total parser, and caches under userData behind a six hour TTL. Applied here:

| File | What it does |
|---|---|
| `src/shared/modelCatalogPayload.ts` | Catalog types (lifted out of the renderer so main validates against the same shape) + `parseModelCatalog`, which never throws and returns `null` for "keep what you have" |
| `src/main/modelCatalog.ts` | Fetch, cache, and **re-validate the cache on read**. The cache is a file a previous build wrote; a schema bump or a hand edit must not reach a picker unchecked just because it once passed |
| `src/main/index.ts` | IPC `models:catalog`, plus a prewarm in `app.whenReady()` so the cache is fresh before any modal opens |
| `src/preload/index.ts` | The bridge |
| `docs/model-catalog.json` | The file you actually edit |

**The baked catalog stays the floor.** It renders offline, on first launch, and whenever the remote copy is missing, unreadable, or announces a schema version this build does not know. The merge is **per provider**: a provider the remote copy names replaces that list, one it omits keeps the built-in list. A bad edit costs a list, never a picker.

**The payload is data, never markup.** Every field lands in an `<option>` as an escaped text node, and a model id is length-capped and stripped of control characters before it can reach a `--model` flag on a spawn command line. That is the one genuinely sharp edge on this path, and it has a test.

`AGENT_MODELS` becomes `agentModels()`. A const array is captured at module load and would still be showing the baked list after a refresh landed.

## The three new models

Every id was read off the CLI that will be invoked with it, not inferred from a name.

| Model | Provider | id | Source |
|---|---|---|---|
| Fable 5.1 | claude | `claude-fable-5-1` | present in the installed `claude` binary. No `[1m]` variant exists |
| GPT-6 Astra | codex | `gpt-6-astra` | `codex-rs/models-manager/models.json` at `rust-v0.153.2` |
| Gemini 3.7 Flash | antigravity | `Gemini 3.7 Flash (High/Medium/Low)` | `agy models` |
| both | cursor | `gemini-3.7-flash-high`, `claude-fable-5-1-thinking-high` | `cursor-agent models` |

### Two things a reviewer should know

1. **GPT Astra is GPT-6 Astra**, released 3 September 2026, and codex only learned the slug in **0.153.1**. An older codex will reject it. The catalog's `minAppVersion` / `maxAppVersion` bound the *app* version, not the CLI version, so this cannot be expressed as a bound. The command field stays editable and the CLI reports the bad slug, which is the documented degradation.
2. **`agy models` now prints `slug<TAB>label`.** The catalog uses agy's display label as the id, which is the existing verified convention, but if that form ever stops being accepted it breaks every antigravity entry, not only the new ones. Worth a follow-up check.

## Testing

- `test/model-catalog-remote.test.cjs`, 18 tests: junk payloads, an unknown schema version, prototype pollution through a provider key, the command-line splice, the length caps, dedupe, and every fallback path.
- The `SHIPPED` literals in `test/model-catalog.test.cjs` and the codex ids in `test/provider-config.test.cjs` are updated **on purpose**. They pin the exact `--model` values the CLIs are invoked with, so a typo there is a model that silently fails to spawn. Adding a model means adding it there too.
- `npm run typecheck` clean, full `node --test test/*.test.cjs` exit 0 on this branch.

## How to add a model after this lands

Add one line to the right provider array in `docs/model-catalog.json`, push to `main`, done:

```json
{ "id": "<the exact --model value>", "label": "<what the picker shows>" }
```

https://claude.ai/code/session_01GKx1g6LQ6Mmzto5BRY8zLT

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,10 +13,16 @@ All notable changes to this project are documented here. The format is based on
   reasoning levels; Cursor gets its own build of the two it carries. Every id was read out of the
   CLI that will be invoked with it — `codex-rs/models-manager/models.json`, `agy models`,
   `cursor-agent models`, and the installed `claude` binary — rather than guessed from a name.
+  **GPT-6 Astra needs Codex 0.153.1 or newer.** The slug landed in that release, so an older
+  `codex` on your machine will reject it when the agent spawns; update Codex first. The catalog's
+  version bounds cover the app, not the CLI it launches, so this is a prerequisite rather than
+  something the picker can hide for you.
 - **A new model no longer needs a release.** The pickers now read
   [`docs/model-catalog.json`](docs/model-catalog.json) on `main`, fetched at runtime and cached for
-  six hours, so adding a model is one line in one file on GitHub and every installed copy picks it
-  up. The catalog compiled into the build stays the floor: it is what renders offline, on first
+  six hours, so adding a model is one line in one file on GitHub rather than a build. The check runs
+  at startup, not on a timer: an installed copy picks a new model up the next time it launches with
+  a cached copy older than six hours, and an app left open does not change under you.
+  The catalog compiled into the build stays the floor: it is what renders offline, on first
   launch, and whenever the remote copy is missing, unreadable, or announces a schema this build does
   not know. A provider present in the remote copy replaces that provider's list; a provider it does
   not mention keeps the built-in one, so a bad edit costs a list rather than a picker. The payload is
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,23 @@ All notable changes to this project are documented here. The format is based on
 
 ## [Unreleased]
 
+### Added
+
+- **Fable 5.1, GPT-6 Astra, and Gemini 3.7 Flash in the model pickers.** Claude Code gets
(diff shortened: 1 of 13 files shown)


perkays:
Very important pull request, thanks for the implementation @chaitanyagiri Chaitanya Giri. We have to merge it as soon as possible.

PR #450, opened 2026-09-05
Make the memory backend swappable

the-cybersapien:
## What this changes

The memory subsystem currently resolves one hardcoded CLI. `mempalace` appears as a literal in three places in `MemoryManager`, its env vars in two more, and its argv in four. Adding any second memory backend means editing nine call sites across three files and keeping them in step by hand.

This lifts all of that into a provider descriptor. `src/main/memoryProviders.ts` (new) owns `bin`, `env()`, `searchArgs()`, `wakeUpArgs()`, optional `mineArgs()`, `scopeForAgent()`, optional `localStorePath()`, optional `auth`, `promptLine()` and `protocolSection()`. A backend becomes a data entry. `MemoryManager` no longer knows what it is talking to.

## Why it is worth doing

Markdown files are a good place for an agent to think and a poor place for a team to remember. `memory.md` grows without bound, every agent's file is a silo until something mines it, and recall depends on a mining pass having run recently enough. That design is exactly right for a local-first default and it stops scaling once the hive has run for a few weeks or an agent needs what another agent learned an hour ago.

Making the backend swappable means the harness stops betting the whole memory story on one shape of store. A server-backed layer can offer namespacing, supersede semantics, history and cross-machine recall without the harness growing a client for any of it, because the CLI already speaks that protocol. Equally, someone who wants a different local store, or none, gets there by adding a descriptor rather than by forking `memory.ts`.

The second provider in this PR is there to prove the seam holds. An abstraction with one implementation cannot be evaluated.

## MemPalace does not move

- A config with no `memoryProvider` key falls back to mempalace and produces a byte-identical binary path, environment and argv.
- `test/memory-provider-argv.test.cjs` pins that argv with `deepEqual`, so it fails if anyone later changes it.
- The existing mine-rearm test stays green.
- `typecheck:node` and `typecheck:web` clean. 723 tests pass.

## Two things the seam exposed

**A reset-path bug.** Full app reset feeds `memory.palacePath()` into a recursive delete. Correct for a local store, wrong for a remote one. `palacePath()` now returns `null` when a provider declares no `localStorePath`, and the reset loop skips it.

**A missing status.** `MemoryStatus` gains `authenticated: true | false | null`. `null` means auth-free or not yet probed; a plain boolean would state something false about a local provider. "Installed but not signed in" previously rendered as an amber "getting ready" pill that never resolved.

`search({wing})` becomes `search({scope})`. A per-agent axis does not generalise to a store that scopes by subject.

## Scope note

`lumberroom ingest` is not a `mempalace mine` equivalent. It reads agent transcripts rather than `memory.md`, and ends in an approval queue. The lumberroom descriptor therefore carries no `mineArgs`, the mine loop does not arm for it, and the UI says so rather than leaving a user to notice.

If you would rather take the refactor without the second provider, `523b7da` and `2d3f68d` are the separable commits.

### Before

Same `config.json`, with `memoryProvider: "lumberroom"` already set. The panel resolves `mempalace`, does not find it, and reads Not set up. The configured provider is invisible to it.

Memory panel before: red Not set up, meaning-based search isn't installed yet

### After

Same config, same machine, same window. The panel resolves the configured provider and speaks lumberroom: Installed, sign in required, the `lumberroom login` card, and the note that `memory.md` is not shared automatically. The local embedding-model chooser is gone, which is right for a remote-backed provider that embeds server side.

Memory panel after: Installed, sign in required, with the lumberroom login card

<details>
<summary>How these were captured</summary>

The running app, not a component render. Built with `npm run build`, launched as Electron with `--remote-debugging-port=9222` and a throwaway `--user-data-dir`, and the renderer screenshotted over CDP with `Page.captureScreenshot`. Same 1280x800 viewport, same clip, same theme, same config in both frames. The only variable is the tree: merge-base `3f53763` for the before, branch tip `e68318c` for the after.

`mempalace` is genuinely not installed on the capture machine, which is why the before frame reads Not set up. `PATH` held a two-line `lumberroom` stub that exits 2 on `whoami`, which is what the real CLI does when signed out and the exit code the descriptor probes for. Nothing in the panel was staged; both differences come from the diff.

</details>


chaitanyagiri:
Still a draft, so one note for when it is ready. The default path is unchanged, which matters most: the generated PROTOCOL.md is identical with MemPalace selected.

The new `lumberroom` entry sits in `BASE_TOOLS` (`src/shared/toolCatalog.ts`), and the catalog returns the whole base list to everyone. So every user sees a lumberroom setup row, with a brew install from your tap, even with MemPalace selected. Please show it only when `memoryProvider` is `lumberroom`.

manvindar:
Also, can we allow an OpenAI-compatible API URL, key, and model instead of relying on Minillm? It starves the CPU when memory become large

Issue #204, opened 2026-08-20
Feat: Custom office people


SeamusMullan:
### Problem / motivation

Currently this project is limited to the scope of the office. as much as I like the office, i think it would be quite nice to have the ability for custom people to be created, a bit like the mii feature on nintendo's products. Each person could be given traits, tools, priorities, etc to customize them further.

### Proposed solution

An "add custom person" button on the add agent page which opens a menu where the appearance and the characteristics of the agent can be defined.

### Area

Office floor / avatars

PR #452, opened 2026-09-06
Character creator: let the user make Michael look like them

gpinkham:
## What & why

Michael is the face of the app and every install has the same one. This adds a
small pixel-art creator so the orchestrator can be *your* Michael — skin, build,
hair, outfit and expression — and changes nothing for anyone who never opens it.
The default recipe is byte-identical to the Michael that ships today, so opening
the creator and saving without touching anything produces exactly the current
character.

Related to #204, and deliberately only part of it. That issue asks for
Mii-style *custom people* — an "add custom person" button on the add-agent page,
creating arbitrary characters with traits, tools and priorities as well as an
appearance. This is the first slice of that: the appearance half, for one
character, the orchestrator. It ships the piece that stands on its own rather
than waiting on the whole design.

What it does **not** do yet, to be plain about it:

- No arbitrary custom people. It customizes the existing orchestrator; it does
  not let you create new characters.
- Nothing behavioral. Traits, tools and priorities are untouched — this changes
  how Michael looks, not how he works.
- Not per-agent. The config field is a single recipe rather than a per-agent
  map, so workers keep the existing cast picker until that generalises. That is
  the natural next step toward #204, and the reason this one is scoped the way
  it is.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before

Command Center header on main: the default Michael, and no way to change him

### After

The identical strip with the change applied: a customized Michael, and an edit affordance on the portrait

Clicking that portrait opens the creator, with a live card/floor preview:

The creator dialog: skin, build, hair style and colour, glasses, facial hair, garment and colour, expression, and Surprise me

Both the before and the after are the same 440x68 clip of the Command Center
header, captured at the same viewport through the same headless path, so the
only difference you should see is the portrait and the affordance beside it.

## How I tested it

- OS: macOS (Darwin 21.6.0)
- Steps:
  - `npm run typecheck` — clean (node and web projects).
  - `npm run test:focused` — 746 pass / 0 fail.
  - `npm run build` — succeeds.
  - `npm run test:renderer` — the renderer smoke check added here: serves the
    built renderer, loads it in headless Chrome, and asserts the app mounts, the
    Command Center renders, and the creator dialog opens. Green.
  - By hand in the app: opened the creator from the Command Center portrait,
    changed skin/hair/outfit/expression, saved, and confirmed the new character
    on the card, on the floor and after a cold restart. Also confirmed an
    untouched install still renders the current Michael exactly.

## Credit (optional)

Discord:

X:

## Checklist

      spacing, or fonts.
      `ATTRIBUTION.md`. — No new art: every character here is drawn
      procedurally by the existing pixel painter, so `ATTRIBUTION.md` needs no
      entry.


gpinkham:
a note  I had this idea as I thought why can't I make Michael look like my clone.  so built this and then noticed issue #204     but I thought this was a good enough step towards that feature that I decided to submit it.

gpinkham:
@chaitanyagiri  I think similar functionality is already in the new version (0.5.2) so i will close this one out.

PR #454, opened 2026-09-06
fix(hive): a god → human ask always reaches the human (ASK ME)

gpechieu:
## What & why

The god's questions to the human could vanish while the god believed they were delivered. Two harness gaps combined: (1) a `to: "human"` message from the god resolves to the god itself (the human's proxy) and the never-deliver-to-self filter leaves no target, so it was logged `delivered: []` and shown nowhere; (2) ASK ME only listed cards with `status === 'blocked'` **and** an open `humanQA`, so an ask the god appended to a card it had just moved to `doing` was in the ledger and visible in none of the three surfaces — while the god told the human "awaiting your answer — ASK ME tab". Observed live: 12 god → human asks dropped in one afternoon, and an empty ASK ME tab with a real sign-off question pending.

The fix makes both invariants the harness's job instead of the agent's:

- `src/shared/humanAsk.ts` — the single definition of an open ask, imported by main and renderer. Waiting on the human = an open ask, whatever the card's status (tab, kanban `?` badge and floor board now agree).
- `hive.routeMessage` — a god → `human` `request`/`query`/`propose` with no inbox target is materialized as a blocked decision card (`ask-<messageId>`, title = subject, ask = body, assignee = god), so the existing answer flow (reply lands as `a`, god gets the `HUMAN ANSWER` inbox message) works unchanged. Logged `delivered: ["human"]` plus a `human-ask` event. An `inform`/`done` is logged as a `human-fyi` drop instead of pretending delivery.
- `hive.tasks()` — announces each newly open ask (`hive:humanAsk` + a native toast gated on the `notifications` setting), whichever way it entered the ledger; the first read after boot only snapshots.
- Generated `PROTOCOL.md` — tells the god it can ask with one message to `human` or a card ask, and that an open ask shows regardless of status.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before

before

On upstream `main` the new tests are red: a god → human ask is dropped (`delivered: []`) and never becomes an ASK ME card, and an open ask on a card that is not `blocked` is invisible to every surface.

### After

after

With the fix the same tests are green: the ask becomes a blocked `ask-…` card announced once, and any open ask on a card that is not `done` waits on the human, whatever its status.

## How I tested it

- OS: macOS 26 (Darwin 25.6.0), Node 24, dev build (`npm run dev`, Electron dev-mode).
- Steps:
  1. `npm run typecheck` — clean. `node --test --test-reporter=tap test/*.test.cjs` — 845/845 on this branch (rebased on main 3e4f9f62); the two new files (`test/hive-human-ask.test.cjs`, `test/human-ask-open.test.cjs`) fail 7/8 against `main`.
  2. Booted the app with the change; checked `hive/PROTOCOL.md` was regenerated with the new "Asking the human" section.
  3. Dropped a real `{to: "human", act: "request"}` message into the god's `outbox/`; the router created `tasks.json` card `ask-<id>` (blocked, humanQA = body) and `log.jsonl` shows `delivered: ["human"]`; the card rendered in the ASK ME tab and was answered from there.
  4. Told the live god its earlier asks had never arrived; it re-sent them as messages to `human` and both became ASK ME cards without any status hand-editing.


chaitanyagiri:
Good fix. God saying "see ASK ME" next to an empty tab is confusing, and turning the ask into a blocked card is the right shape. One change before we take it.

**Leave done cards out.** `openQuestion` (`src/shared/humanAsk.ts:31`) ignores status, `waitsOnHuman` (`:44`) and `openAsks` (`:60`) build on it, and `test/human-ask-open.test.cjs` asserts that a `done` card still counts. Existing hives carry old unanswered `humanQA` entries on cards that finished long ago, so right after upgrading every one of them shows up in ASK ME at once. Please exclude `done` cards and update that test.

gpechieu:
Done in f004f286 — `done` cards are out.

- `openQuestion()` returns nothing for a card whose `status` is `done`, whatever its `humanQA` holds. `waitsOnHuman()`, `openAsks()`, the kanban badge, the floor board and main's ask announcer all derive from it, so they cannot disagree.
- `isDoneCard()` is the one status test, exported.
- `test/human-ask-open.test.cjs`: the F1 case no longer lists `done`; a new case pins the stale-hive regression (a done card with an unanswered question shows nowhere, and reopening the card makes the same ask pending again); `openAsks()` is checked to skip such a card while keeping the rest.

Verified: the new test fails on the previous commit (2 failures), passes now; typecheck clean; 846/846 in the local suite.

gpechieu:
**Agent review of this PR** (CONTRIBUTING: "run your coding agent over your own PR, and paste what it found"). Claude Code went over `3e4f9f62..f004f286` (src + test) for cross-platform behaviour, paths with spaces, provider neutrality, hot paths and security, then hygiene.

- **Cross-platform / paths with spaces:** clean — all fs paths go through `path.join`, JSON-only I/O, no shell, no env vars.
- **Provider neutrality:** clean — the ask lives in the ledger; the toast is Electron `Notification`; nothing names a CLI.
- **Hot paths:** clean — `openAsks` is O(cards × humanQA) per `tasks()` read (the 5 s polls), dwarfed by the existing `readJson`; the notifier's disk reads run only per *new* ask.
- **Security:** clean — `ask-<msgId>` only ever lives inside tasks.json (no task id reaches a filesystem path in main); the log entry carries ids only.
- **Hygiene — two things it flagged, both fixed in f9cce6fd:**
  - `hive.emit('hive:humanAsk')` had no preload bridge and no renderer listener, so the earlier commit message overstated what ships. Dropped it (the renderer learns of an ask from the ledger it already polls; the native toast comes from main's notifier) and removed the test assertion that checked the phantom event.
  - The `routedObserver` JSDoc had been left orphaned above `materializeHumanAsk` when the latter was inserted. Back above the field it documents.
- **Flagged and deliberately not changed:** the explicit `['request','query','propose']` act list next to `requires_reply` / `needs_human` in `materializeHumanAsk`. It overlaps with what `normalize()` folds into `requires_reply`, but the doc says "acts that expect an answer, or anything flagged", i.e. an explicit `requires_reply:false` on a `request` still becomes a card. That is the intended reading of a god ask, so it stays.

Verified after the fix: typecheck clean, the two ask test files 12/12, local suite 846/846.

gpechieu:

PR #455, opened 2026-09-06
fix(worker-wake): nudge a stalled worker whatever its terminal prints, and log why one is held

gpechieu:
## What & why

A god-spawned worker sat **17 minutes** on its work order — 0 tokens, no transcript, inbox unread — until the human typed "read your inbox" into its terminal by hand. The renderer's boot-time inbox nudge never landed (the screen locked seconds after the spawn; `workerWake.ts`'s own header describes that case), and the main-process `WorkerWakeWatchdog` (#151), the designed backstop for exactly this, never fired: across an afternoon of 14 workers it logged nothing at all. With the edge trigger from #358 (each inbox id announced once) that becomes permanent: one lost nudge and the worker is never woken again.

Its rules inferred "mid-turn" from recent PTY output and "booting" from no output. Neither is evidence of work — a TUI redraws its chrome without doing anything, and the boot sequence itself is output — so a worker whose CLI never took its first turn could be held forever, silently.

- `WorkerWakeFacts` gains `lastActivityAt` (the agent's last telemetry usage sample — the CLI's own evidence of a turn) and `oldestMailAt` (`created_at` of the oldest pending message). Both optional; facts without them behave exactly as before.
- `isStalledWorker`: mail older than `WORKER_WAKE_STALL_MS` (90 s) with no usage sample since it landed bypasses the quiet-output, never-output and already-announced rules; the moment the CLI shows a turn after the mail, the #358 edge trigger rules again. Paused / halted / delivery-paused / HITL / boot-grace / cooldown still hold, so a real permission prompt is never typed into and retries are once per cooldown until the mail drains.
- `activityEvidenceAt({usage, spans})`: "activity" is a turn the CLI demonstrably took — a tool span, or a usage sample WITH tokens. The collector stamps a zero-token sample at session start, and on 2026-09-07 that sample (timestamped after the work order) read as activity: the beat logged `holding <worker>: mid-turn (mail pending 69s, pty quiet 0s, last activity 63s ago)` for a worker with 0 tokens, no tool and no transcript. The same line shows `pty quiet 0s` for a worker idle at its prompt: a Claude TUI never goes output-quiet, which is why silence cannot be the signal.
- `explain()` names the hold; the beat logs `[worker-wake] holding <id>: <why> (mail pending Ns, pty quiet …, last activity …)` once per cooldown for any worker on mail older than 60 s. The watchdog's silence is no longer indistinguishable from "nothing to do".

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before

before

On upstream `main` the new tests are red: a worker with old mail and no session activity reads as mid-turn because its terminal is chatty, so the watchdog never nudges it.

### After

after

With the fix the same tests are green: the stalled worker is nudged whatever its terminal prints and retried each cooldown; only paused / halted / HITL / boot-grace hold it, and every hold is named in the log.

## How I tested it

- Steps:
  1. `npm run typecheck` — clean. `node --test --test-reporter=tap test/*.test.cjs` — 846/846 on this branch (rebased on main 3e4f9f62, on top of #358); `test/worker-wake-stall.test.cjs` (12) fails 12/12 against `main`; the existing `test/worker-wake.test.cjs` (14, including the #358 edge-trigger cases) is untouched and green.
  2. Booted the app with the change integrated (dev build); workers spawned by the god boot and drain their inbox as before; no `[worker-wake] holding` lines for healthy workers (mail age never reaches 60 s).


chaitanyagiri:
The diagnosis is right: a recent redraw reads as "mid turn", so a worker whose first nudge was lost can sit on its order with nothing in the log. One fix before we take it.

**Activity evidence only exists for Claude agents.** `activityEvidenceAt` (`src/main/workerWake.ts:145`) reads `telemetry.getAgentUsage` and `telemetry.getSpans`, and the OTel exporter is only switched on for Claude Code (`src/main/hive.ts`, `if (claudeProvider && this._otelEndpoint)`). For a Codex, Gemini, agy, pi or grok worker the evidence is always 0. Once its mail is 90 s old, the stall rule (`workerWake.ts:119`) treats it as stalled on every tick, so it gets nudged every cooldown while it is actually working, until the file moves to `.done`. That brings back the repeated nudging #368 removed.

Either limit the stall override to agents that export telemetry, or give it an activity signal the other engines do produce, such as the time of their last hook event.

gpechieu:
Done — the stall rule now reads the hooks as activity, and stays off for an agent that has no channel at all.

- `isTurnHook()`: a hook event that proves a turn (`UserPromptSubmit`, `PreToolUse`, `PostToolUse`, `PostToolUseFailure`, `Stop`, `StopFailure`, `SubagentStart`, `SubagentStop`). `SessionStart` is the CLI coming up, not a turn — a worker whose boot nudge was lost has exactly that and nothing else; `Notification` is the CLI waiting. Every engine the harness shims lands on these names (the Gemini/grok mappings in hive.ts).
- The watchdog remembers, per agent, when its hooks last proved a turn and when they last said anything (`noteHook` already receives every boundary). The stall check credits `max(telemetry evidence, last turn hook)`.
- It is OFF for an agent that has produced neither a telemetry sample nor a single hook event this session: with no channel that could ever show a turn, "no activity" is not evidence, so the pre-stall rules apply unchanged (quiet terminal → one nudge, then #358's edge trigger). `WorkerWakeFacts.hasTelemetry` (the collector holds any sample, even the zero-token boot one) is how the beat says the telemetry channel exists.
- `forget()` drops the hook memory with the rest; the hold log's "last activity" includes hook evidence.
- Tests: the hook classification; a hook-less, telemetry-less engine is never stalled (the #368 regression); a hook-only engine with no turn since the mail is stalled and retried; a turn hook after the mail is activity (also for a Claude worker with dead telemetry); `forget()` resets.

Verified: 4 of the new tests fail on the previous commit, all 17 pass now; typecheck clean; 851/851 in the local suite.

gpechieu:
**Agent review of this PR** (CONTRIBUTING: "run your coding agent over your own PR, and paste what it found"). Claude Code went over `3e4f9f62..b3026bc9` for cross-platform behaviour, paths with spaces, provider neutrality, hot paths and security, then hygiene.

- **Cross-platform / paths with spaces:** clean — no paths, shell, env or line endings touched; `Date.parse` only runs on the hive's own `toISOString()` values.
- **Provider neutrality:** clean — it checked every bridge can emit a turn hook: agy maps BeforeAgent/BeforeTool/AfterTool/AfterAgent onto UserPromptSubmit/PreToolUse/PostToolUse/Stop; gemini, codex/grok, pi, opencode and the qwen/crush proxy all emit PreToolUse/PostToolUse/Stop. Bridge-less engines emit nothing and the transcript fallback returns null at zero tokens, so they stay unobservable and keep the pre-PR rules. The one Claude-only assumption (the zero-token OTel sample) is gated by `hasTelemetry`.
- **Hot paths — one thing it flagged, fixed in 0d61ed20:** `telemetry.getAgentUsage()` was called twice per worker per beat, and with no live OTel each call runs the transcript fallback (a registry read plus a scan of the transcript directory). The beat reads it once and derives both facts.
- **Security:** clean — the hold log prints agent ids and durations; the nudge text is unchanged; a forged `created_at` can only trigger the fixed nudge, bounded by the 60 s cooldown.
- **Hygiene — fixed in 0d61ed20:** `decide()`'s JSDoc had been orphaned above `explain()` (back above `decide()`, with the bounded-memory rationale that had been dropped); `lastHoldReportAt` was declared mid-class; `lastActivityAt`'s doc and the test header still called telemetry the sole evidence.
- **Not changed:** nothing else was flagged.

Verified after the fix: typecheck clean, the stall test file 17/17, local suite 851/851.

gpechieu:

gpechieu:

PR #461, opened 2026-09-07
fix(terminal): https URLs in agent output are not clickable

### LINKED ISSUE #281: Links in chat pane are not clickable (no cmd-click / click-to-open behavior)
**Environment**
- App: Munder Difflin
- Version: 0.4.5 (bundle id `in.munderdiffl.app`)
- OS: macOS (Darwin 25.5.0)

**Description**

URLs that appear in assistant messages in the chat pane render as plain text/markdown but aren't clickable. Cmd-click (and regular click) does nothing — there's no way to open a link from the chat without manually selecting and copying the text.

**Expected behavior**

URLs in chat messages should be auto-linkified and openable — either via a plain click or cmd-click — using `shell.openExternal` (or equivalent) to open them in the user's default browser, similar to how most Electron-based chat/terminal UIs handle links.

**Actual behavior**

Links render as inert text. No visual affordance (underline, color, cursor change) indicates they're interactive, and no click/cmd-click/hover action opens them.

**Steps to reproduce**

1. Start a session where the assistant's response includes a URL (e.g., a `https://...` link).
2. Try clicking or cmd-clicking the URL in the chat pane.
3. Observe nothing happens — no browser opens, no context menu, no visual feedback.

**Suggested fix**

Add link detection/linkification to the markdown renderer used for chat messages, and wire click (or cmd-click, matching the platform convention) to `shell.openExternal` in the main process via IPC from the renderer.

ketan0095:
## What & why

Closes #281.

URLs printed by an agent were never clickable, and the cause is not a missing feature — it is a collision. The provider only ever looked for **path** tokens, and `PATH_TOKEN_RE` carries an optional Windows-drive prefix `(?:[A-Za-z]:[\\/])?`. That group matches the `s:/` inside `https://`, so a URL was claimed as an absolute path on drive `s:`, resolved against the agent's cwd, stat'd, missed, and silently dropped. No underline, no click, no error.

`terminalLinkSpans()` now resolves URL spans first and drops any path token overlapping one; `provideLinks` is a thin adapter over it. The ordering *is* the fix, and it lives in `terminalPaths.ts` rather than the provider so it can be unit-tested — that module is already where "what each type does" lives.

Three deliberate choices, all following constraints already in the file:

- **No `WebLinksAddon`.** The provider's own comment rejects it; this extends the custom provider instead.
- **⌘/Ctrl-click only**, matching the path links, so a plain click keeps going to the TUI.
- **`https` only.** `app:openExternal` accepts nothing else, so underlining an `http://` would promise a click that cannot work — the complaint in #281, not a fix for it. The scheme is lowercased on the way out because that opener matches `^https://` case-sensitively, and `HTTPS://` otherwise fell into the same drive-letter trap as the original bug.

Balanced brackets stay in the URL (eating the `)` off a Wikipedia link is a silent 404), and a bare scheme with nothing after it is not offered at all.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before

Before

_Rendered from the real session; the same output is described below so it is searchable._ On `origin/main`, `https://x.com/docs/guide.html` is matched by the **path** matcher as `s://x.com/docs/guide.html` — drive `s:`, which resolves nowhere. `HTTPS://UPPER.COM/x` likewise becomes `S://UPPER.COM`. The new tests are red against unfixed source.

### After

After

_Same commands, with the fix._ Each URL yields one `url` span; a line carrying both a path and a URL still yields both, in reading order. 28 tests pass.

## How I tested it

- OS: macOS (Darwin 25.5.0), Apple Silicon
- Steps:
  - `npm run typecheck` — clean, node + web.
  - `npm run test:focused` — 846 pass. The 3 failures are `test/agent-exit-signal-e2e.test.cjs`, which fails identically on a clean `origin/main` checkout on this machine with `spawn failed: posix_spawnp failed` — a sandbox limitation here, unrelated to this change.
  - Mutation-tested every new guard one at a time: removing the URL/path overlap skip → 3 red; allowing `http://` → 1 red; weakening the trailing-prose strip → 1 red; **deleting the ⌘/Ctrl gate on URL activation → 2 red**; adding a second `openExternal` call site → 1 red. Restored source green each time.
  - That last pair is why `test/terminal-link-activation.test.cjs` exists: without it the modifier gate could be deleted and the whole suite stayed green, which would turn agent output into a one-click browser navigation.

## Checklist

### DIFF
--- a/src/renderer/src/components/terminalPool.ts
+++ b/src/renderer/src/components/terminalPool.ts
@@ -22,7 +22,7 @@ import { arabicJoinRanges } from '@/terminal/arabicJoiner';
 import { attachArabicSpacingFix } from '@/terminal/arabicSpacingFix';
 import { isArabicTerminalEnabled } from '@/terminal/arabicSetting';
 import {
-  classifyPathToken, isPathToken, pathTokenMatcher, stripPathToken, type PathAction
+  classifyPathToken, isPathToken, stripPathToken, terminalLinkSpans, type PathAction
 } from '@shared/terminalPaths';
 import {
   createTerminalRecoveryState,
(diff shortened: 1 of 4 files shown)


chaitanyagiri:
This looks good to take as it is. A few edge cases for a follow up, none of them blocking:

* A URL that wraps onto a second terminal row opens truncated.
* Markdown bold around a URL (`**https://...**`) and CJK punctuation right after one get pulled into the link.
* A pathological line of 10,000 `)` characters takes about 1.4 s to trim. A real row is capped at the terminal width, so this is unlikely in practice.

PR #463, opened 2026-09-07
fix(renderer): a queued message is delivered when the CLI confirms it, not when the PTY takes the bytes

gpechieu:
## What & why

Twice in two days a god-spawned worker sat on its work order with 0 tokens, no transcript and an unread inbox until a human typed "read your inbox" into its terminal (one for 17 minutes on 2026-09-06, another for 8+ minutes on 2026-09-07). The renderer **had** typed the boot-time inbox nudge — its persisted queue (`cth.messageQueues`) was empty and the card read `idle / awaiting` — and the CLI never received it.

Three rules combined into a silent loss:

1. `terminalReadyToReceive` judges a Claude TUI "ready" once it has painted anything and 400 ms have passed. That is the boot banner, not the input box; keys typed before the CLI attaches its input handler are discarded.
2. `deliverWithAcknowledgement` removed the queue item as soon as the PTY write resolved: bytes taken was read as message received.
3. The inbox-wake poll marks every message id as announced at enqueue time, so a lost nudge was never re-enqueued; the main-process watchdog then held the worker as mid-turn (a Claude TUI never goes output-quiet — `pty quiet 0s` even at an idle prompt).

It is a boot race: the nudge lands ~7 s after spawn; 13 of 15 workers had an input-ready CLI by then, two did not.

**Fix — a delivery is confirmed by the agent's own `UserPromptSubmit` hook**, which the harness already receives from the Claude, Codex and Gemini shims:

- `queueDelivery.PromptAckTracker` records prompt submits per agent; `waitFor(agentId, since, timeout)` resolves only for a submit at/after the typing (a stale one is exactly the boot race), or `false` on timeout.
- `deliverWithConfirmation(send, confirm, acknowledge)` → `'delivered' | 'unconfirmed' | 'failed'`. Unconfirmed leaves the item queued; the drain retries after its cooldown and, after `MAX_ACK_MISSES` (3), acknowledges on the write with a console warning so a CLI with broken hooks is never re-typed into forever.
- Slash commands (`/compact`, `/clear`, …) and providers without a prompt hook keep write-is-delivery.
- `useHive` notes `UserPromptSubmit` hook events into the tracker and uses the confirming delivery in the queue drain (`PROMPT_ACK_TIMEOUT_MS` = 10 s).

Companion: #455 makes the main-process watchdog re-nudge a worker that demonstrably never took a turn, as defence in depth. This PR removes the loss at its source.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before

before

On upstream `main` the new tests are red: a delivery is acknowledged when the PTY takes the bytes, so a prompt lost while the TUI boots is never retried, and hook events reach one window only.

### After

after

With the fix the same tests are green: a delivery is acknowledged on the CLI's own UserPromptSubmit, an unconfirmed one is retried, a hook-dead agent keeps write-is-delivery, and hook events fan out to every window.

## How I tested it

- Steps:
  1. `npm run typecheck` — clean. `node --test --test-reporter=tap test/*.test.cjs` — 844/844 on this branch; `test/prompt-delivery-ack.test.cjs` fails 9/10 against `main`; `test/queue-delivery.test.cjs` unchanged and green.
  2. Root cause established on the live app before writing code: hive `log.jsonl` (work order delivered at spawn), `fleet.json` (0 tokens, no tool, backlog 1), the watchdog's own `holding … mid-turn` line, the renderer's persisted localStorage (`cth.messageQueues` = `{}`, agent `idle/awaiting`), and no transcript file for the session.
  3. Live validation of the confirmed delivery on the next app restart (dev-mode restart kills running agents, so it is scheduled).


chaitanyagiri:
Right fix. Counting a message as delivered when the PTY takes the bytes is how a keystroke lost during a TUI boot, or a Return the TUI swallows, goes missing with no retry. Three things before we take it:

* **Agents whose hooks never fire get every message typed 4 times.** With no `UserPromptSubmit` coming back, each delivery waits, retries, and only falls back after `MAX_ACK_MISSES` (`src/renderer/src/hooks/useHive.ts:105`). Hooks are dead on Windows when the harness path has a space (#477 fixes that), so today that is every Claude, Codex and Gemini agent there. Please keep the old rule for any agent that has not sent a single hook event this session.
* **A retry after a swallowed Return submits the text twice.** If the Return was absorbed, the text is still sitting in the input box, so typing it again joins both copies into one prompt. Clear the line first, or send only Return on the retry.
* **Hook events reach one window.** Main sends them through `liveWebContents()`, which returns `mainWindow`, and `mainWindow` follows focus. With a second floor window open, deliveries to the unfocused floor never see their confirmation and get typed again.

gpechieu:
Done, all three:

**Hook-dead agents keep write-is-delivery.** `PromptAckTracker.noteHook()` records a hook event of any kind (the listener feeds it every event; a prompt submit counts too), and `receipt()` replaces the bare wait in the drain: `confirmed` / `unconfirmed` (no receipt although the hooks are alive → retry, as before) / `hookless` (no receipt and not a single hook event this session → acknowledged on the PTY write). An agent judged hookless answers at once from then on, so it pays the wait on its first message only; any hook event lifts the verdict. One console line per such agent, pointing at #477. Typed once, not four times.

**A retry clears the line first.** `submitToPty()` takes `{ clearLineFirst }`, and a delivery with a previous unconfirmed attempt sends Ctrl-U before the text — the readline kill-to-start every supported CLI honors, the same key `clearTerminalDraft` uses. If the TUI lost the text too, it is a no-op; if it only swallowed the Return, the stale copy is gone before the retype.

**Hook events reach every window.** `fanOutHookSink()` (pure, in hooks.ts) sends `hive:hookEvent` and `hive:contextUpdate` to every live window and `control:approvalRequest` to the live window alone — one prompt, one answer. `HookServer`'s sink is structural now (`HookSink | WebContents`), so the existing tests' `() => null` still fit; main lists every window `createWindow` registered.

Tests: the `receipt()` verdicts (unconfirmed with live hooks, confirmed by a later submit, hookless once and then immediate, lifted by a hook event; a prompt submit proves the hooks; `forget()` resets) and a new `hook-event-fanout` file (every window gets the state events, only the live one the approval, a torn-down window does not stop the others).

Verified: 5 of the new tests fail on the previous commit, all pass now; typecheck clean; 849/849 in the local suite.

gpechieu:
**Agent review of this PR** (CONTRIBUTING: "run your coding agent over your own PR, and paste what it found"). Claude Code went over `3e4f9f62..ff83da27` for cross-platform behaviour, paths with spaces, provider neutrality, hot paths and security, then hygiene.

- **Cross-platform:** clean — `\x15` is the byte `clearTerminalDraft` already sends on every platform, it fires only on a retry, and retries exist only for claude/codex/gemini; `\r` submit, env and paths untouched. It noted the Ctrl-U comment overclaimed for multi-line pastes (kill-to-start clears the current line, same limit as `clearTerminalDraft`) — comment fixed in 27cc9695; nudges are one line.
- **Paths with spaces (#477):** behaves — a hook-dead agent pays one 10 s wait on its first message, then immediate write-is-delivery, one warn per agent, lifted by any later hook event.
- **Provider neutrality — one thing it flagged, fixed in 27cc9695:** "hookless" was judged on *any* hook event, so an engine whose other hooks fire but whose prompt hook never does (a Gemini shim without BeforeAgent, a Codex shim the CLI-side timeout killed under load) got the same prompt re-submitted `MAX_ACK_MISSES` times. Now the full retry budget applies only once the agent has confirmed a prompt this session (`PromptAckTracker.hasConfirmed`); before that it gets one retry, which still covers the boot-time loss this PR exists for (the second attempt lands ~10 s later on a TUI that is up). The other nine providers: `confirmable` is false, so no wait, no misses, no Ctrl-U.
- **Hot paths:** clean — per-agent dispatch is unawaited, so a 10 s wait blocks nobody else; one promise + timer per confirmable delivery, cleared on resolve. It flagged that `submittedAt` was stamped before `submitToPty` waited on the write chain / `waitForTerminalReady` (up to 30 s), so a submit in that window (a chained seed, the human's own Enter) could pass for this message's receipt — fixed in 27cc9695: `submitToPty` reports the instant it wrote the Return, and that is the reference.
- **Security:** clean — no new IPC surface, `validateHookEvent` still gates, approvals single-window, warnings log ids only.
- **Hygiene — fixed:** the `HookSink` / `fanOutHookSink` block sat between imports in hooks.ts.
- **Flagged and deliberately not changed:** the `inFlight` check sits after two IPC awaits (pre-existing ordering, not this PR's change); `forget()` is tested but not called on PTY exit (`hooksSeen` surviving a same-id respawn only means a missing receipt is treated as real and retried, bounded); `HookSink | WebContents` is redundant but reads better at the constructor.

Verified after the fix: typecheck clean, the two test files 16/16, local suite 850/850.

gpechieu:

gpechieu:

PR #465, opened 2026-09-07
Recover from renderer death instead of freezing forever

chaoukibarkia:
## What & why

A renderer crash left the app permanently frozen, silent, and unkillable.

Observed live on 2026-09-07 (v0.4.6, packaged AppImage, Hyprland/Wayland): after ~25 minutes running a five-agent hive, the compositor raised "Munder Difflin is not responding". `/proc` held the browser process, both zygotes, the GPU process and the network service — and **no renderer at all**. The window stayed mapped on its last painted frame while the main process carried on routing the hive; `hive/log.jsonl` was still being appended seconds later.

Three gaps turned one renderer crash into a dead end:

1. **Nothing noticed.** `createWindow` bound no `render-process-gone` and no `unresponsive`; `app` bound no `child-process-gone`. Recovery from outside is not possible either — a compositor will not give focus to a non-responding surface, so the `View > Reload` the menu already provides is unreachable. That leaves the compositor's "Terminate", and every agent PTY is a child of the browser process, so it takes the whole running hive down with it.
2. **Nothing was recorded.** A packaged launch from a `.desktop` file has stdout/stderr on `/dev/null`, so every `console.error` is discarded, and no crashpad handler was attached, so the death left no dump.
3. **The app could no longer quit.** Both quit paths cancel the quit and ask the *renderer* to draw the "you have running terminals" warning. With no renderer, nothing answers, nothing sets the allow-quit latch, and every later quit is cancelled the same way — the frozen window survived `SIGTERM` and needed `SIGKILL`.

Two traps worth flagging for review, both found by measuring rather than reading docs:

- Electron reports a renderer killed by a signal as `reason: 'killed'`, **including an OOM kill**. An earlier draft of this fix treated `killed` as deliberate teardown and did not recover from it — the window stayed frozen exactly as before. `test/renderer-health.test.cjs` pins this.
- `webContents.isCrashed()` alone was not a sufficient liveness guard: measured against Electron 32, a window whose renderer had been SIGKILLed still passed an `isCrashed()`-only check, so the quit was still cancelled. The fix trusts our own observation of `render-process-gone` first.

### Changes

- `src/main/rendererHealth.ts` — wires renderer liveness onto every window. A crashed renderer is reloaded automatically under a rolling budget (3 per 10 minutes) so a crash loop can never become a reload loop; past the budget, and for a hang long enough that a compositor would flag it, a native dialog offers the reload instead — a route that does not require killing the process tree the hive lives in. Deaths and hangs are recorded to `userData/logs/renderer-health.log`. GPU/utility deaths are logged too: when a renderer dies moments later, the order of the two is the diagnosis.
- `logQuitPromptDecision()` guards both quit paths, so a quit is only ever cancelled when a renderer is alive to answer the prompt — and the decision is logged, because a quit that does not happen is otherwise completely silent.
- `src/main/pty.ts` — coalesces PTY output. `proc.onData` fires once per read, and an agent CLI redrawing its TUI emits a long train of tiny writes. Each one became its own `pty:data:<id>`: a structured clone, an IPC hop and a separate renderer main-thread task, per chunk, for every live session at once, whether or not that terminal was even on screen. With one agent pegged at ~97% CPU that is enough main-thread work to starve the renderer, which is the most likely trigger of the freeze above. Output is now flushed one message per frame, immediately past 64 KiB so a large burst is never held back, and drained before `pty:exit` so a process's dying words cannot land after its exit banner. Bytes and order are unchanged.

Cross-platform: no platform-specific code, no new paths built by hand, provider-neutral (nothing here knows which CLI is running). No UI beyond two native dialogs.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before

The failure as it happens on `main`. The window is mapped and painting its last frame, but the bundle has no renderer process, nothing was logged, and the main process is demonstrably still alive and working:

```
$ ls -l /proc/*/exe | grep appimage_extracted     # every process in the bundle
  pid 7304    (browser)
  pid 7307    --type=zygote
  pid 7308    --type=zygote
  pid 7310    --type=zygote
  pid 7432    --type=gpu-process
  pid 7522    --type=utility
  -> no --type=renderer: the renderer is gone

$ hyprctl clients                                  # the window is still mapped
  munder-difflin | Munder Difflin | mapped = True

$ ls ~/.config/munder-difflin/logs/                # nothing was recorded
  ls: cannot access '.../logs/': No such file or directory

$ tail -1 ~/HarnessAgents/hive/log.jsonl           # the main process is fine
  last hive event: 2026-09-07 10:36:56
```

On screen at that moment: the window painted its `backgroundColor` (`#FFF8E7`) and the office skyline, plus the browser-drawn menu bar — and nothing else. No sidebar, no agents, no terminal. (Screenshots are attached in a follow-up comment.)

And the quit wedge (defect 3):

```
$ kill 7091 7304      # SIGTERM
  still alive: pid 7304
$ kill -9 7304        # only SIGKILL ends it
```

The suite is red on the coalescing behaviour:

```
$ node --test test/pty-output-coalescing.test.cjs
✖ a burst of chunks becomes ONE ipc message, bytes and order intact
  Expected 1 coalesced message, got 7
ℹ pass 2  ℹ fail 1
```

### After

The same failure, provoked on the fixed build — the renderer killed exactly the way it died on its own:

```
$ kill -9 285779                                   # the renderer

$ cat ~/.config/munder-difflin/logs/renderer-health.log
  2026-09-07T10:10:20.163Z render-process-gone {"label":"primary","reason":"killed","exitCode":9}
  2026-09-07T10:10:20.164Z reload {"label":"primary","why":"render-process-gone:killed","used":1}

$ ls -l /proc/*/exe | grep appimage_extracted      # a fresh renderer, 1 ms later
  pid 285268  (browser)
  pid 285284  --type=zygote
  pid 285285  --type=zygote
  pid 285287  --type=zygote
  pid 285430  --type=gpu-process
  pid 285767  --type=utility
  pid 307763  --type=renderer
```

On screen: the office floor is fully back — sidebar, all four agents, live terminal — with no user action, and the hive never stopped. (Screenshots attached in a follow-up comment.)

The suite is green:

```
$ node --test test/renderer-health.test.cjs test/pty-output-coalescing.test.cjs
ℹ pass 15  ℹ fail 0
```

## Checks

- `npm run build` — clean
- `npm run test:focused` — 845 pass, 1 fail (`update-download-asset`, which fails identically on a pristine `main` on this machine: it needs a downloaded Electron binary that did not fetch here)
- Verified end to end against a locally built AppImage of this branch, on Hyprland/Wayland.

## Agent review

Reviewed by the agent that wrote it; the two findings it produced were fixed rather than shipped, and both are in the tests:

- it initially classified `killed` as deliberate teardown, which silently disabled recovery for the OOM case — caught by running the real build, not by reading the diff;
- the `isCrashed()`-only quit guard passed review but failed in the running app, so the guard now trusts our own `render-process-gone` observation first.

Left deliberately unchanged: the coalescing window is a fixed 16 ms rather than adaptive, and `HANG_PROMPT_AFTER_MS` is a fixed 20 s. Both are judgement calls that want real-world numbers before being made clever.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -10,6 +10,7 @@ import { join, resolve, sep, basename, dirname, isAbsolute } from 'node:path';
 import { homedir } from 'node:os';
 import { request as httpsRequest } from 'node:https';
 import { PtyManager, type SpawnOptions } from './pty';
+import { installWindowHealth, installProcessHealth, logQuitPromptDecision } from './rendererHealth';
 import { resolveCommand as resolveCliCommand, isSafeCommandName } from './shellEnv';
 import { initAutoUpdater, abortPendingRestart } from './updater';
 import { RealtimeFloorWatcher } from './realtimeFloorWatcher';
@@ -2319,6 +2320,18 @@ function createWindow(opts: { floor?: boolean } = {}): BrowserWindow {
   // reference stays valid as the per-PTY ownership key.
   const wc = win.webContents;
 
+  // Renderer liveness. Installed BEFORE the first load so a crash during load
+  // is caught too. A dead renderer used to leave this window mapped and frozen
+  // forever, with the compositor's "Terminate" — which kills the whole agent
+  // tree — as the user's only way out. See rendererHealth.ts.
+  installWindowHealth(win, {
+    label: isFloor ? 'floor' : 'primary',
+    // A reload keeps the same webContents object, so PTY ownership and the
+    // default sink stay valid; re-assert the sink anyway so the primary's
+    // routing survives even if that ever stops being true.
+    onReloaded: (next) => { if (!isFloor) ptyManager.attachWebContents(next); }
+  });
+
   allWindows.add(win);
   // Global timer events follow the user — the most-recently-focused window is
   // primary. The primary is also seeded synchronously so boot events route now.
@@ -2409,6 +2422,9 @@ function createWindow(opts: { floor?: boolean } = {}): BrowserWindow {
     // Primary window: existing app-wide quit warning (renderer modal).
     const count = ptyManager.list().length;
     if (count === 0) return;
+    // Same wedge as before-quit: with no renderer to draw the warning, cancelling
+    // the close leaves a window that can never be closed. Let it go instead.
+    if (!logQuitPromptDecision(win, 'window-close')) return;
     e.preventDefault();
     win.focus();
     wc.send('app:closeRequested', { ptyCount: count });
@@ -5269,6 +5285,11 @@ function onSystemResume(reason: string): void {
 }
 
 app.whenReady().then(() => {
+  // GPU/utility deaths are recorded, not acted on: Chromium respawns them, but
+  // when a renderer dies moments later the ORDER of the two is the diagnosis —
+  // and a packaged launch has stdout on /dev/null, so nothing else keeps it.
+  installProcessHealth();
+
   // Realtime Michael mic-gate hygiene (rt-8 / Pam rt-10 nit): the voice session
   // opens the mic permission gate by persisting realtimeVoiceEnabled=true and
   // closes it on disconnect — but a hard crash/reload mid-session skips that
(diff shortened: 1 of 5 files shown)


chaitanyagiri:
Thanks for this. A dead renderer freezing the app for good is a real problem, and we'd like to take the recovery and the PTY output batching. One fix first.

**A quit with no live renderer skips teardown.** In `src/main/index.ts:5386`, `before-quit` returns early when `logQuitPromptDecision` finds nobody to draw the prompt. The quit then goes straight to `will-quit` (`:5425`), which only flushes analytics and calls `app.exit(0)`. Electron does not emit `window-all-closed` while `app.quit()` is in progress, so nothing from `teardownAndQuit()` runs: proxy bridges and PTYs are never stopped. On Windows a child process outlives its parent, so every agent tree is left running. Please run the same teardown steps in that branch before letting the quit continue.

Smaller, not blocking:
* A quit that arrives while a reload is in progress is still swallowed once.
* If the last focused floor is the dead one, the running terminals warning is skipped for the whole app.

PR #466, opened 2026-09-07
Detect models from the installed CLIs, and refresh the curated list

chaoukibarkia:
## What & why

The model pickers were two hand-written layers deep — the list baked into the build, and `docs/model-catalog.json` fetched from main. Both drift, and neither can know anything about the machine it is running on.

Measured on 2026-09-07 against the CLIs installed on one machine:

| Provider | Curated list | Reality |
|---|---|---|
| `codex` | 4 models | 10 in the CLI's own `~/.codex/models_cache.json` — missing `gpt-reserve`, `gpt-daybreak-blue-latest`, `gpt-5.5`, `gpt-5.4-mini`, `gpt-5.3-codex-spark`, `codex-auto-review` |
| `antigravity` | no 3.8, no 3.6 | `agy models` reports the full Gemini 3.8 and 3.6 Flash families |

Two changes.

**1. The curated list is refreshed** from what those CLIs actually report, in both copies (`docs/model-catalog.json` and the baked `src/shared/modelCatalog.json`), plus `claude-opus-4-7` and `claude-opus-4-6`, which were reachable but unlisted. `test/model-catalog.test.cjs` and `test/provider-config.test.cjs` pin picker contents deliberately, so their expected lists move with it.

**2. `src/main/modelDetect.ts` adds a third layer** — what the CLI on *this* machine says about itself — opt-in behind a new `autoDetectModels` setting (Settings › General, default off). It is the only layer that can be right about a private preview, an account tier, or a CLI pinned to an old version. Layering is per provider, and detection can only add or correct rows, never empty a picker: a provider that reports nothing is absent from the result and keeps its curated list.

Two deliberate limits, both load-bearing:

- **Only CLIs that can actually be asked are detectable.** `codex` maintains its models cache itself (so detection is a file read, not a subprocess) and `agy models` is a documented subcommand. `claude --model` takes an alias or a full name and offers no list; the gemini CLI has no enumeration either. Both are absent by design — inventing a list for them would be strictly worse than the curated one.
- **For antigravity the detected id is the DISPLAY NAME, not the slug.** This is the trap the tests exist for. `agy models` prints `slug<TAB>display name`, but the flag rejects the slug:

  ```
  $ agy --model gemini-3.8-flash-high --print hi
  Error: invalid model selection (--model "gemini-3.8-flash-high"):
    model gemini-3.8-flash-high is not recognized as a known model
  Available models:
    Gemini 3.8 Flash (High)
    ...
  ```

  Detecting the obvious-looking first column would produce a picker in which every row fails at spawn. (This also confirms the existing catalog is right to use display names for this provider — I checked before changing them, and did not.)

Probes are bounded (6 s timeout, 512 KiB output cap, 120-char id cap) and every failure mode — missing binary, non-zero exit, timeout, corrupt cache, oversized id — collapses to "we learned nothing" and leaves the curated list rendering. Cross-platform: no platform-specific code; `execFile` with `windowsHide`, `path.join` throughout, and a CLI that is not installed is simply not detected.

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before

The shipped catalog against what the CLIs on this machine actually report:

```
$ python3 -c "import json; d=json.load(open('src/shared/modelCatalog.json'));
              print([m.get('id') for m in d['providers']['codex']])"
[None, 'gpt-6-astra', 'gpt-5.6-sol', 'gpt-5.6-terra', 'gpt-5.6-luna']

$ python3 -c "import json; d=json.load(open('$HOME/.codex/models_cache.json'));
              print([m['slug'] for m in d['models']])"
['gpt-6-astra', 'gpt-reserve', 'gpt-5.6-sol', 'gpt-5.6-terra', 'gpt-5.6-luna',
 'gpt-daybreak-blue-latest', 'gpt-5.5', 'gpt-5.4-mini', 'gpt-5.3-codex-spark',
 'codex-auto-review']
                                              # 6 models the picker cannot offer

$ agy models
gemini-3.8-flash-high     Gemini 3.8 Flash (High)
gemini-3.8-flash-medium   Gemini 3.8 Flash (Medium)
gemini-3.8-flash-low      Gemini 3.8 Flash (Low)
gemini-3.6-flash-high     Gemini 3.6 Flash (High)
...                                # none of these are in the antigravity list
```

There is no setting to change that, and no code path that reads either source:

```
$ grep -rn "models_cache\|agy models" src/
$                                                      # nothing — no detection
```

### After

The curated list matches the CLIs, and the new layer reads them live:

```
$ python3 -c "import json; d=json.load(open('src/shared/modelCatalog.json'));
              print(len(d['providers']['codex']), len(d['providers']['antigravity']))"
11 18                                                     # was 5 and 12

$ node --test test/model-detect.test.cjs
✔ agy: the id is the DISPLAY NAME, because that is what --model accepts
✔ agy: a line with no tab is skipped, not guessed at
✔ codex: reads the cache the CLI maintains itself, keeping slugs
✔ codex: a missing or corrupt cache detects nothing rather than something wrong
✔ an oversized id is dropped before it can reach a --model flag
✔ only CLIs that can actually be asked are detectable
ℹ pass 6  ℹ fail 0
```

Settings › General gains an **Auto-detect models** switch (default off) that applies immediately — turning it on re-reads the CLIs, turning it off restores the curated list, with no reload either way.

## Checks

- `npm run build` — clean
- `npm run test:focused` — 836 pass, 1 fail (`update-download-asset`, which fails identically on a pristine `main` on this machine: it needs a downloaded Electron binary that did not fetch here)
- Model lists verified against `codex` 0.153.4 and the `agy` CLI on Linux; the `agy` id format verified through the CLI's own rejection message rather than assumed.

## Agent review

- The slug-vs-display-name question was resolved by running the CLI rather than reading its help text; the first draft would have used the slug. That is now the single most-commented line in the module and has its own test.
- It wanted to "fix" the existing antigravity entries to slugs on the grounds that they looked wrong. Checked first, and they were right — no change made.
- Left deliberately unchanged: `cursor-agent models` is mentioned in the catalog's own notes as another enumerable CLI, but it is not installed here, so no detector was written for it rather than one written blind. The extension point is a two-line addition next to the other two.


chaitanyagiri:
Detecting models from the installed CLIs is useful. It would be much easier to take as its own opt in PR, without the catalog refresh. Specifics:

* **Ids that should not ship.** The catalog changes in `docs/model-catalog.json` and `src/shared/modelCatalog.json` add Codex ids that look internal or specific to one account, such as `gpt-reserve`, `gpt-daybreak-blue-latest` and `codex-auto-review`. Everyone would see those in the picker. Please drop the catalog refresh from this PR.
* **Detection replaces the curated list.** The description says detection only adds rows, but the merge in `src/renderer/src/store/config.ts` is `{ ...BAKED.providers, ...(REMOTE?.providers ?? {}), ...DETECTED }`, so a detected provider's list replaces the curated one outright. Merging detected ids into the curated list would match the description.
* **`agy` is not found when the app starts from Finder.** `detectAntigravityModels` calls `probe('agy', ['models'])` (`src/main/modelDetect.ts:139`) with the app's own `PATH`, which is minimal for a Finder launch. The rest of main resolves CLIs through `resolveCommand` in `src/main/shellEnv.ts`, and using it here would fix that.

PR #468, opened 2026-09-07
Sweep agents a previous run left behind

chaoukibarkia:
## What & why

Agent PTYs are children of the browser process, and the normal quit path tears them down: `killAll()` kills each one, and on POSIX closing the pty HUPs the foreground group. Neither runs if the app never gets to quit — a main process that is SIGKILLed, OOM-killed, or lost to a power cut leaves its agents running, reparented to init, with no window and nobody to stop them.

They are not merely idle. Observed live on 2026-09-07: an orphaned `codex` from a killed run still held its session rollout file open, so the **next** launch could not resume that agent at all, and it was dead on arrival on every relaunch until the orphan was killed by hand. A survivor also keeps burning tokens and CPU against an account nobody is watching.

`src/main/staleAgents.ts` records the PTYs a run starts and sweeps the survivors at the next startup, before that run spawns anything of its own. `PtyManager` owns the bookkeeping (add on spawn, remove on exit and kill, clear on `killAll` — a clean teardown leaves nothing to sweep), and every call is guarded so a ledger failure can never break a spawn.

**The whole risk is PID reuse.** A recorded pid may belong to a stranger by the next launch, and killing a stranger's process tree is far worse than leaving an orphan. So every kill is gated on *identity*, not liveness: the process must still have the start time **and** the command recorded when it was definitely ours — and that identity is read at spawn time, not at sweep time, when it would only describe whoever holds that pid then. Whenever identity cannot be established (process gone, `ps` did not answer, a ledger predating this field, or the platform is Windows) nothing is killed. Windows therefore sweeps nothing rather than something wrong; its quit path already runs a synchronous `taskkill /T /F` in `killAll()`, which is the case this covers elsewhere.

One subtlety worth a reviewer's eye: `ps -o lstart=` renders in the machine's **locale**, so it is neither fixed-width nor a known format — `Mon Sep  7 09:46:24 2026` in C, `lun. sept.  7 09:46:24 2026` in fr_FR. The first draft parsed it by column width and silently identified nothing. It is now parsed from the end. The string being locale-dependent is fine: it is only ever compared against a value this same machine recorded earlier, and a locale change between runs makes the comparison fail, which spares the process.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before

A run that does not get to quit leaves its agents behind, and nothing ever collects them:

```
$ ls -l /proc/*/exe | grep -c appimage_extracted      # after the app is gone
0
$ ps -eo pid,etime,comm | grep -E ' (agy|codex)$'     # its agents are not
   9133    01:00:27 codex
   9779    01:00:26 agy
  10003    01:00:25 codex
```

pid 10003 was still holding the rollout file of the agent it used to be:

```
$ for d in /proc/[0-9]*/fd; do ls -l $d 2>/dev/null \
    | grep -q rollout-...-01a075e4-....jsonl && echo "held by ${d}"; done
held by /proc/10003/fd
```

so the next launch could not bring that agent up at all:

```
  >_ OpenAI Codex (v0.153.4)
  Resuming session…
› Error: Failed to resume session from
  .../jim-mtpju5cs/.codex/sessions/2026/09/06/rollout-...-01a075e4-....jsonl:
  thread/resume failed during TUI bootstrap: thread/resume failed:
  thread 01a075e4-b741-7811-b7a0-8178b8ef26bc already has an active writer (code -32600)
— process exited (code 1) —
```

There is no code path that would have cleaned this up:

```
$ grep -rn "agent-pids\|sweepStale\|orphan" src/main/
$                                                        # nothing
```

### After

The same shape, exercised end to end against real processes — one detached child that gets swept, and one that must survive because its recorded identity does not match:

```
$ node --test test/stale-agents.test.cjs
✔ identity is start time AND command — either mismatch spares the process
✔ an unverifiable process is never killed
✔ a ledger row from an older build, with no identity, is never killed
✔ a missing or corrupt ledger sweeps nothing and does not throw
✔ a clean shutdown leaves nothing for the next launch to sweep
✔ an exited agent is forgotten, so its pid can never be swept later
✔ a real orphan is killed; the ledger is dropped afterwards
✔ a pid whose process is NOT the one we recorded is spared
ℹ pass 8  ℹ fail 0
```

The last two are the ones that matter: the first spawns a real detached process, records it, sweeps it, and asserts it is actually gone; the second spawns a real detached process, writes a ledger row claiming a *different* start time for that pid, and asserts the process is still running afterwards and was reported as `spared`.

## Checks

- `npm run build` — clean
- `npm run test:focused` — 838 pass, 1 fail (`update-download-asset`, which fails identically on a pristine `main` on this machine: it needs a downloaded Electron binary that did not fetch here)

## Agent review

- The first draft parsed `lstart` at a fixed 24 columns, taken from the C-locale format. On an fr_FR machine that matched nothing, so *every* process was unverifiable and the sweep silently became a no-op. Caught by running the tests against real processes rather than by reading the code; now parsed from the end, with the locale dependence documented and argued.
- It initially read process identity at sweep time, which proves nothing — by then the pid may be someone else's. The lookup moved to spawn time.
- Left deliberately: no Windows identity check. `wmic`/CIM creation-date lookups would work, but writing one blind on a platform not available here is exactly how a wrong kill ships. The Windows path is a documented no-op, and `killAll()` already covers the clean-quit case there.

## Related

Best read after #465, which is what makes a SIGKILL of the main process unnecessary in the first place. This one covers the crashes and power cuts that #465 cannot.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -10,6 +10,7 @@ import { join, resolve, sep, basename, dirname, isAbsolute } from 'node:path';
 import { homedir } from 'node:os';
 import { request as httpsRequest } from 'node:https';
 import { PtyManager, type SpawnOptions } from './pty';
+import { AgentLedger, sweepStaleAgents } from './staleAgents';
 import { resolveCommand as resolveCliCommand, isSafeCommandName } from './shellEnv';
 import { initAutoUpdater, abortPendingRestart } from './updater';
 import { RealtimeFloorWatcher } from './realtimeFloorWatcher';
@@ -5269,6 +5270,17 @@ function onSystemResume(reason: string): void {
 }
 
 app.whenReady().then(() => {
+  // Reap agents a previous run left behind, BEFORE this run spawns any of its
+  // own — so the ledger being read is unambiguously the previous run's. An
+  // orphaned agent is not just idle: it keeps its provider session file open,
+  // and the next launch cannot resume that agent at all while it does.
+  // Every kill is gated on process identity; see staleAgents.ts.
+  const agentLedger = new AgentLedger(join(app.getPath('userData'), 'agent-pids.json'));
+  try {
+    sweepStaleAgents(join(app.getPath('userData'), 'agent-pids.json'), (m, d) => console.log(m, d));
+  } catch (e) { console.error('[stale-agents] sweep failed:', e); }
+  ptyManager.setLedger(agentLedger);
+
   // Realtime Michael mic-gate hygiene (rt-8 / Pam rt-10 nit): the voice session
   // opens the mic permission gate by persisting realtimeVoiceEnabled=true and
   // closes it on disconnect — but a hard crash/reload mid-session skips that
--- a/src/main/pty.ts
+++ b/src/main/pty.ts
@@ -4,6 +4,7 @@ import { existsSync, readFileSync, statSync } from 'node:fs';
 import { delimiter, join, win32 } from 'node:path';
 import { spawnSync } from 'node:child_process';
 import { ensureKilled, hardKillTree } from './procKill';
+import type { AgentLedger } from './staleAgents';
 import { expandTilde } from './fs';
 import { buildPtyEnv } from './ptyEnv';
 import { captureFromLoginShell, isSafeCommandName, userShellPath } from './shellEnv';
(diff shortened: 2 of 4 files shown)


chaitanyagiri:
Agents left running after a hard kill are a real cost, and a sweep is the right idea. Two problems make it unsafe as it stands:

* **A second instance kills the first instance's agents.** `app.whenReady()` (`src/main/index.ts:5272`) sits outside the `else` of `requestSingleInstanceLock()` (`:2226`). A process that loses the lock calls `app.quit()`, but `whenReady` still resolves, and the sweep is the first thing it runs. It reads the live instance's `agent-pids.json` and kills its agents. On Linux a second process starts for every `munderdifflin://` link, and on macOS `open -n` does the same. Please sweep only while holding the lock, and store the owning process id and its start time in the ledger so a live owner's agents are never swept.
* **On macOS the identity is read too early.** The command recorded for a new agent is captured before the exec completes, so in 13 of 16 runs it recorded the wrong command. The identity check then refuses to kill real orphans too.

PR #470, opened 2026-09-07
Add a floor switcher, and give floors distinct titles

chaoukibarkia:
## What & why

Floors already ran in parallel — each window owns its own PTYs (routed by `PtySession.owner`), and `backgroundThrottling: false` keeps an occluded floor's heartbeat loops running even behind a lock screen. What was missing was any way to *reach* a particular one.

Two things stood in the way, and only the first is obvious.

**1. Nothing listed the open floors.** `installAppMenu()` is called once at startup, so even a static list would have gone stale on the first `New Floor`.

**2. Every floor was called "Munder Difflin".** `createWindow` passes `title: 'Munder Difflin — Floor'`, but `src/renderer/index.html` carries `<title>Munder Difflin</title>`, and in Electron the **page** title wins as soon as it loads — silently, with no error. Setting it in the constructor was never enough. Verified against a running build on 2026-09-07: every open window reported the same title to the compositor, so the window manager's own switcher could not tell them apart either.

`src/main/floors.ts` owns the naming and the cycling as pure functions; `index.ts` wires them:

- `page-title-updated` is refused and the title re-applied after load, so the page can no longer stomp it.
- `refreshFloors()` renumbers every window and rebuilds the menu on open **and** on close — numbering is positional, so closing floor 2 renumbers floor 3, and a stale menu would point at a window that is gone.
- A `Floors` menu lists the open windows, checkmarks the focused one, and binds each to `CmdOrCtrl+Alt+<n>`, plus `CmdOrCtrl+Alt+Left/Right` to cycle.

Two choices worth a reviewer's eye:

- **The accelerator is `CmdOrCtrl+Alt+<n>`, not the browser-style `CmdOrCtrl+<n>`.** A menu accelerator is claimed before the focused element sees the key, and these windows are mostly full of agent TUIs. `Ctrl+1` is plausible TUI input; `Ctrl+Alt+1` is not. This is the same concern that made the Edit menu's clipboard items `registerAccelerator: false`, arrived at from the other direction — there the fix was to leave the key alone, here it is to pick a key nothing else wants.
- **The rows are `type: 'checkbox'`, not `'radio'`.** Radio items in an Electron submenu group take the first item's state when none is checked, which would mark floor 1 as current no matter which window had focus.

Numbering is positional rather than by a stable id, so the numbers stay dense and always match what the menu shows — the same behaviour browser tabs have. The primary keeps the bare product name; renaming it to "Floor 1" would be a downgrade.

No change to how floors run. Cross-platform: no platform-specific code, `CmdOrCtrl` throughout, and the menu itself is already gated on the existing `multiWindow` flag.

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before

Two windows open (primary + one `New Floor`), as the compositor sees them:

```
$ hyprctl clients -j | jq -r '.[] | select(.class|test("munder";"i")) | .title'
Munder Difflin
Munder Difflin
```

Identical. Nothing in the app lists them either — the menu is `File / Edit / View / Window`, with no floor entry anywhere:

```
$ grep -n "role: 'viewMenu'" -A2 src/main/index.ts
    { role: 'viewMenu' },
    { role: 'windowMenu' }
  ];
```

So the only way to reach a specific floor is to guess from the window manager's list of identically-named windows.

### After

Titles are distinct and numbered, and the naming/cycling is pinned by tests:

```
$ node --test test/floor-switcher.test.cjs
✔ the primary keeps the bare product name; floors are numbered from 2
✔ every floor gets a DISTINCT title — the whole point of the change
✔ the menu row reads the same way as the key that reaches it
✔ accelerators stop at 9 rather than binding nonsense
✔ the accelerator avoids keys an agent TUI might want
✔ cycling wraps in both directions
✔ cycling with nothing focused enters at the near end
✔ cycling with no windows reports nothing to focus
ℹ pass 8  ℹ fail 0
```

The menu gains a `Floors` submenu that tracks the open windows:

```
Floors
  ✓ 1  Munder Difflin                Ctrl+Alt+1
    2  Munder Difflin — Floor 2      Ctrl+Alt+2
    3  Munder Difflin — Floor 3      Ctrl+Alt+3
    ─────────────────────────────
    Next Floor                       Ctrl+Alt+→
    Previous Floor                   Ctrl+Alt+←
```

Closing floor 2 renumbers floor 3 to 2 and rebuilds the list, so the numbers, the titles and the accelerators never disagree.

## Checks

- `npm run build` — clean
- `npm run test:focused` — 838 pass, 1 fail (`update-download-asset`, which fails identically on a pristine `main` on this machine: it needs a downloaded Electron binary that did not fetch here)

## Agent review

- The first draft only changed the `title:` option in `createWindow`, which does nothing — the page title overrides it on load. Caught by checking the live window titles rather than by reading the constructor.
- It initially reached for browser-style `Ctrl+<n>`, which would have stolen keystrokes from the agent TUIs that fill these windows. There is a test for the key shape now, precisely because the wrong choice is invisible until someone loses a keystroke mid-run.
- Radio menu items looked like the natural fit for "which floor is current" and were wrong for it; checkbox items are used instead.
- Left deliberately: no in-window floor switcher UI, and no per-floor agent scoping. The on-disk hive is process-global and floors share it (as documented on the `multiWindow` flag) — changing that is a much larger design question than a switcher.

### DIFF
--- a/src/main/floors.ts
+++ b/src/main/floors.ts
@@ -1,0 +1,68 @@
+/**
+ * Telling floors apart, and moving between them.
+ *
+ * Floors already run in parallel — each window owns its own PTYs (routed by
+ * `PtySession.owner`), and `backgroundThrottling: false` keeps an occluded
+ * floor's heartbeat loops running. What was missing was any way to *reach* a
+ * particular one.
+ *
+ * Two things stood in the way, and only the first is obvious:
+ *
+ *  1. Nothing listed the open floors. The app menu is built once at startup, so
+ *     even a static list would have gone stale on the first `New Floor`.
+ *
+ *  2. Every floor was called "Munder Difflin". `createWindow` passes
+ *     `title: 'Munder Difflin — Floor'`, but `src/renderer/index.html` carries
+ *     `<title>Munder Difflin</title>`, and in Electron the PAGE title wins as
+ *     soon as it loads — silently, with no error. Verified on 2026-09-07: every
+ *     open window reported the same title to the compositor, so the window
+ *     manager's own switcher could not distinguish them either. Setting the
+ *     title at construction is therefore not enough; it has to be re-applied
+ *     after load and defended against `page-title-updated`.
+ *
+ * Numbering is by POSITION, not by a stable id, so the numbers stay dense as
+ * floors come and go and always match what the switcher menu shows. A floor
+ * being renumbered when an earlier one closes is the same behaviour browser tabs
+ * have, and it keeps `Ctrl+Alt+2` meaning "the second one in the list".
+ */
+
+/** The primary window keeps the bare product name — it is the one users think
+ *  of as "the app", and renaming it to "Floor 1" would be a downgrade. */
+export const PRIMARY_TITLE = 'Munder Difflin';
+
+/** Title for the window at `index` in floor order (0 = primary). */
+export function floorTitle(index: number): string {
+  return index === 0 ? PRIMARY_TITLE : `${PRIMARY_TITLE} — Floor ${index + 1}`;
+}
+
+/** Menu label for the window at `index`. Numbered so the row reads the same way
+ *  as the accelerator that reaches it. */
+export function floorMenuLabel(index: number): string {
+  return `${index + 1}  ${floorTitle(index)}`;
+}
+
+/** Accelerator for the nth floor, or undefined past what we can bind.
+ *
+ *  `CmdOrCtrl+Alt+<n>` rather than the browser-style `CmdOrCtrl+<n>`: a menu
+ *  accelerator is claimed before the focused element sees the key, and these
+ *  windows are mostly full of agent TUIs. Ctrl+1..9 is plausible input for a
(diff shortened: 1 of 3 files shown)


chaitanyagiri:
The floor switcher is handy with several floor windows open. Two things to fix:

* **The checkmark goes stale.** `checked: w.isFocused()` is evaluated when the menu is built, and the menu is rebuilt by `refreshFloors()` (`src/main/index.ts:373`) but not when a window gains focus. Clicking into another floor with the mouse leaves the old one checked. Calling `refreshFloors()` from the `focus` handler (`:2371`) would fix it.
* **AltGr on Windows.** Windows reports AltGr as Ctrl+Alt, so `CmdOrCtrl+Alt+<n>` may fire when someone types a character that needs AltGr and a digit, for example `~ # { [` on a French layout, switching floors instead of typing. I have not tested this on Windows, so please check it before changing anything.

PR #472, opened 2026-09-07
fix(worker): a worker that signaled done leaves a settled inbox

gpechieu:
## What & why

An ephemeral worker's inbox is left exactly as the worker left it when it is released, and workers seldom file their own work order under `inbox/.done` before signaling done. The worker id is `worker-<request name>`, reused on every re-hire of the same name, so each new incarnation boots into its predecessors' finished orders: the boot nudge tells it to "work everything still pending", it spends its first turns re-triaging tasks its memory says are done, and the inbox-wake watchdog reads the oldest order as mail that has gone unanswered for days (`holding <worker>: boot-grace (mail pending 77359s …)` — 21 hours, on a worker spawned 30 seconds earlier).

Seen live on 2026-09-07 with god hiring by character name all day: 13 of 30 worker inboxes on one floor carried finished orders, one of them three deep (yesterday's order, today's, and an `inform` a peer sent after the worker was archived).

**Fix — a worker that signaled done leaves a settled inbox.**

- `hive.settleInbox(id)`: moves every unread inbox message to `inbox/.done`, logs `{kind:'inbox-settled', agentId, count}` and commits. Returns 0 when the inbox is clean or the agent has no mailbox; never throws.
- `ephemeralWorkerTick`: the **done** branch settles before the kill and logs `[worker] <id>: filed N unread inbox message(s) under inbox/.done`. Only the done path — an idle or token-cap reap never signaled completion, so its unread mail stays pending for god to see.
- Messages are filed, never deleted: `inbox/.done` is what the voice read-layer, `voiceMessages` and the memory miner already read.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before

before

On upstream `main` the new tests are red: a released worker's unread work orders stay in its inbox and greet the next hire under the same id as pending mail.

### After

after

With the fix the same tests are green: the done release files the mail that predates the done signal under `inbox/.done`, leaves later mail pending, and reports unanswered requests.

## How I tested it

- Steps:
  1. `npm run typecheck` — clean. `node --test --test-reporter=tap test/*.test.cjs` — 838/838 on this branch; `test/worker-inbox-settle.test.cjs` fails 4/4 against `main` (`settleInbox` missing; the tick never settles).
  2. Root cause established on the live app before writing code: `dev.log` (`[worker] spawned <worker>` followed by `holding … mail pending 77359s`), the worker's session transcript (boot nudge listing yesterday's message id; "already completed per my memory"), `ls` of the inbox and `.done`, and a sweep of every `agents/worker-*/inbox`.
  3. Live validation on the next app restart (a dev-mode restart kills the running agents, so it is scheduled).


chaitanyagiri:
Right problem. Worker ids repeat per name, so the next hire boots into the previous worker's finished orders and the watchdog reads them as ignored mail. One fix first.

**Only settle what arrived before the done signal.** `settleInbox` (`src/main/hive.ts:1030`) moves every unread file, and it runs on the done path (`src/main/index.ts:4894`). A message that lands after the worker signalled done, such as a follow up question from god, gets filed as read and nobody ever sees it. Please compare each message's timestamp to the done signal and leave newer ones where they are. It would also help to tell god when a `request` or `query` is filed unread, so nothing disappears silently.

gpechieu:
Done — the settle now stops at the done signal, and god hears what was filed.

- `workerDoneAt()` replaces `workerSignaledDone()`: same scan and stale-done guard, but it returns the (newest) done signal's timestamp, so the release path has a cut-off.
- `settleInbox(id, before)` leaves any message created after `before` in place, still pending (`created_at`, falling back to the file's mtime for a half-written file). It returns `{ moved, kept, unanswered }`, where `unanswered` lists the filed `request` / `query` messages (id, act, from, subject).
- The done path logs what it kept and, when a request or query was filed unread, sends god an `inform` with the list (through `informGod`, so a worker with a Slack thread gets the close-the-loop line too): "if one of them is not the order it just completed, nobody is working on it — resend it or re-hire".
- Tests cover the cross-over (a query sent after the done stays pending; the earlier request is filed and reported), the mtime fallback, and the index.ts pin now requires both the cut-off and the report to god before the kill.

Verified: all 6 tests fail on the previous commit, pass now; typecheck clean; 840/840 in the local suite.

gpechieu:
**Agent review of this PR** (CONTRIBUTING: "run your coding agent over your own PR, and paste what it found"). Claude Code went over `3e4f9f62..6a4cf360` for cross-platform behaviour, paths with spaces, provider neutrality, hot paths and security, then hygiene.

- **Cross-platform / paths with spaces:** clean — `path.join` everywhere, the rename stays inside the same directory tree, the cut-off compares epoch ms (`toISOString` stamps vs `mtimeMs`) so it is timezone-neutral; no shell interpolation.
- **Provider neutrality:** clean — pure file protocol (`act:"done"` JSON in the outbox), nothing Claude-specific.
- **Hot paths:** clean — `settleInbox` runs once per release; a live worker without a done already scanned outbox + `.sent` fully every tick before this PR.
- **Security:** `f` is a readdir basename (no traversal); `from` is router-authoritative. It flagged that an agent-written subject containing a newline could forge extra list lines in the message to god — fixed in 57019cd8 (ids, senders and subjects are folded onto one line).
- **Design — the one should-fix, fixed in 57019cd8:** the worker's own work order is dispatched by `processSpawnRequest` in conversation `worker-<reqId>` as a `request` from god, and per this PR's own data workers rarely file it, so nearly every release would have told god its own completed order was "filed unread" — one extra god turn and hive commit per release for nothing. `settleInbox` now reports each filed message's `conversation`, and the done path leaves that one out of the report; only *other* requests/queries reach god.
- **Hygiene — fixed:** the "a later settle retries" comment on a failed rename promised a retry that does not exist (the message simply stays pending, the pre-settle state).
- **Flagged and deliberately not changed:** `normalize()` honours an agent-supplied `created_at`, so a bare timestamp without an offset would parse as local time and skew the cut-off — the protocol has the harness stamp `created_at`, and agents that write their own get the fallback they wrote; `.sent` growing across re-hires is pre-existing.

Verified after the fix: typecheck clean, the settle test file 6/6, local suite 840/840.

gpechieu:

gpechieu:

PR #473, opened 2026-09-07
Slack media: outbound file upload + inbound download fix + thread reply-ledger

Monilnarang:
## What & why

Adds Slack **media** support in both directions and closes two silent data-loss bugs.

- **Outbound:** a tokenless helper (`resources/md-slack-upload.cjs`) plus a loopback `/upload`
  route on `SlackReplyServer` let the app share a local file into a Slack thread using Slack's
  external-upload flow. The caller only ever hands over a *local file path* — the bot token
  stays in the main process and is never exposed to the helper or the caller.
- **Inbound:** the file downloader now prefers `url_private_download`, follows Slack-host
  redirects (re-sending auth **only** to `*.slack.com`), and refuses the 2-line HTML
  auth-redirect *stub* it used to save to disk instead of the real file.
- **Reliability:** thread replies now persist a bot-thread ledger so follow-up replies in a
  thread aren't dropped, and `postSlackReply` returns the posted message `ts`.

## Type of change

- [x] Bug fix
- [x] New feature

## Evidence

There is no visible UI for this change (it is Slack transport/reliability in the main process),
so evidence is a terminal capture of the failing behaviour reproduced and then guarded, plus the
focused test suite that goes red→green.

### Before

A file shared into Slack was saved as a 2-line HTML auth-redirect **stub** rather than the real
bytes — the downloader followed the `url_private` redirect to the login page and wrote the HTML
response verbatim:

```
$ cat slack-files/foo.md
<a href="https://<workspace>.slack.com/?redir=%2Ffiles-pri%2F...%2Ffoo.md">Found</a>.
```

And there was **no** outbound path at all — the app could reply with text but could not deliver
a file into a thread.

### After

The guard rejects the HTML stub and only accepts real file bodies; the outbound `/upload` route
reads the real bytes and supplies the token from the main process. Focused suite is green:

```
$ node --test test/slack-upload.test.cjs test/slack-file-download.test.cjs
✔ buildUploadBody: requires channel/thread/file; resolves abs path; carries title+comment
✔ /upload: reads the local file and hands its bytes + target to uploadFn (200)
✔ /upload: records the thread as bot-participated (onReplied) on success
✔ /upload: missing file path → 400, uploadFn never called
✔ /upload: nonexistent file → 400 (never a stub upload)
✔ /upload: wrong reply token → 401
✔ /upload: no bot token configured → 503
✔ /reply still works after adding the /upload route (no regression)
✔ looksLikeHtmlStub catches the EXACT auth-redirect stub from the bug report
ℹ tests 9
ℹ pass 9
ℹ fail 0
```

Manually verified outbound end-to-end: a batch of image files was delivered into a live Slack
thread via the upload helper (helper exits non-zero unless Slack returns HTTP 200 && `ok`, so a
clean exit is genuine delivery, not a stub).

## How I tested it

- OS: macOS
- Steps:
  - `npm run typecheck` — pass (node + web projects).
  - `npm run build` — pass (exit 0).
  - `node --test test/slack-upload.test.cjs test/slack-file-download.test.cjs` — 9/9 pass
    (upload route reads real bytes, token supplied by main not the caller; download guard
    passes real PDF/PNG/JPEG/GIF bodies while catching HTML stubs).
  - Manually: delivered files into a live Slack thread via the upload helper; confirmed the
    thread received the real files.

## Checklist

- [x] Before and after evidence is included above (terminal/test capture — no UI to screenshot).
- [x] `npm run test:focused` passes (focused suite: `slack-upload` + `slack-file-download`).
- [x] This PR is one change (Slack media/reliability).
- [x] I read the diff myself; no debug output, no secrets, no unrelated churn.

### DIFF
--- a/resources/md-slack-reply.cjs
+++ b/resources/md-slack-reply.cjs
@@ -10,16 +10,25 @@
  * Slack's chat.postMessage. The token never appears here, in the prompt, or in
  * any transcript.
  *
+ * On a SUCCESSFUL post it also UPSERTS the bot-thread follow-ledger
+ * (`slack-bot-threads.json`, alongside the app config in userData) that
+ * `md-slack-poller.cjs` reads to decide which threads to pull replies for. This
+ * is what makes a user's later THREAD REPLY reach the bot with no @-mention:
+ * conversations.history only returns top-level messages, so a thread is only
+ * followed if it's in this ledger. Every agent/god reply goes through THIS CLI,
+ * so writing the ledger here (rather than relying on the MD main process) is what
+ * guarantees the ledger actually exists. The ledger holds channel + thread
+ * timestamps ONLY — never a token. A ledger write failure never breaks the reply.
+ *
  * Usage:
  *   node md-slack-reply.cjs --channel C123 --thread 1700000000.000100 --text "..."
- *   (optional) --config /abs/path/to/slack-reply.json
- *
- * The discovery file is located via, in order: --config, then the
- * MD_SLACK_REPLY_CONFIG env var (injected into every agent by main).
+ *   (optional) --config /abs/path/to/slack-reply.json   (else MD_SLACK_REPLY_CONFIG)
+ *   (optional) --ledger /abs/path/to/slack-bot-threads.json  (else alongside config)
  */
 'use strict';
 
 const fs = require('node:fs');
+const path = require('node:path');
 const http = require('node:http');
 
 /** Parse `--key value` and `--key=value` pairs from argv. */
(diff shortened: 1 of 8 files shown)


chaitanyagiri:
Thanks. The inbound download fix is a real bug: Slack's file URL redirects, the current code does not follow it, and Slack's login page gets saved as the file. Your redirect handling looks right (exact host match, auth sent only to Slack hosts, https on every hop, 5 hops at most).

We'd like to take that part now. Could you open it as its own PR? The rest has problems that would hold it up:

* **`/upload` reads any file on disk.** `handleUpload` (`src/main/slack.ts:674`) takes an absolute path, follows symlinks through `statSync`, and posts anything up to 100 MB. An agent reacting to a message in the bridged channel could be talked into posting the app's config or an SSH key into the thread. It needs a root it cannot leave, such as the agent's working folder, checked after resolving symlinks.
* **It would not work in a built app.** `electron-builder.yml:43` ships `md-slack-reply.cjs` as an extra resource but not `md-slack-upload.cjs`, and the setup docs never ask for the `files:write` scope that Slack's upload API needs.
* **The thread reply ledger and `slackPollingOnly`** are better reviewed on their own, after the two points above.

PR #476, opened 2026-09-08
Feat/agent chat

perkays:
## What & why

An agent's only visible surface was its raw TUI: spinners, tool-call lines, box drawing, and the harness's own injected text scrolling past. There was no way to just read what an agent said, or to reply without driving a terminal. This adds a Chat tab beside Terminal for every agent — god and worker alike — showing the plain user/assistant turns and nothing else, with the same message queue underneath so a reply typed here reaches the agent exactly as a Terminal-tab message does.

Where the conversation lives turns out to be per-provider, and that is most of the work. Claude Code hands every hook the path of a JSONL transcript, so that one is tailed incrementally (offset-cached, like the existing usage reconciler). OpenCode writes no transcript file at all — it keeps every turn in one SQLite database — so its bridge plugin now stamps the session id on each payload it posts, and a second reader queries that database read-only. Codex writes a JSONL rollout per session under its own `CODEX_HOME`, named after the session id, and a third reader tails that. All three strip every tool call / tool result block before anything crosses the IPC boundary; only Claude's and Codex's readers share code (`tailChatJsonl` in `chatTranscript.ts` does the incremental, torn-line-safe tailing, each provider supplies its own record parser), OpenCode's SQL query needs none of that.

The remaining piece is deciding what counts as the human talking. Several things the app or the CLI itself types into an agent arrive shaped exactly like a real prompt:
- the hive's identity/protocol brief injected at spawn (for every provider that cannot take it as a system prompt, it goes in through `--prompt` and lands as the session's first user message — a wall of harness text sitting in the chat window),
- the inbox-wake nudge,
- the `/compact` from the context cap,
- Codex's own injections on top of those — the repo's `AGENTS.md` block and its `<environment_context>` block, and its `developer`-role seat (skills/tool guidance), which is never a conversation turn at all.

Each is recognised by a predicate that lives with the code that writes it, so a reword cannot leave a stale copy behind in the reader.

A provider with no reader at all (codex was this, briefly, mid-series) gets an honest "Chat cannot read this CLI's history yet" rather than a permanent, misleading "No conversation yet." — the code and the string both say "not read yet", never "keeps no history", because that would be false for a CLI like Codex that always kept one.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### After

OpenCode reader against a live agent's database, before and after the filters:

```
resolved session: ses_f7e3acaafffeNWXQ2d2eNGGrBX
turns: 18
  assistant | Floor rotation handled. Summary: ...
  user      | bist du bereit ? sind alle agenten am laufen ?
  assistant | Ich bin bereit — aber alle drei Agenten sind noch nicht am Laufen: ...
  user      | wie weit bist du ?
  assistant | Sprint "KI-Gegner" — Stand: ...
```

Codex reader against a live reviewer's `CODEX_HOME`, resolved the way `hive:agentChat` resolves it:

```
registry sessionId: 01a08233-72ea-7481-98c9-ef954a56aadb
resolved rollout  : rollout-2026-09-08T20-06-44-01a08233-72ea-7481-98c9-ef954a56aadb.jsonl
turns: 21
  …
  20:55:57 user      | alles in ordnung ?
  20:56:03 assistant | Ja, der Review-Ablauf funktioniert wieder. Alle Urteile sind verbucht, …
  21:05:07 user      | alles in ordnung ?
  21:05:12 assistant | Ja, bei mir ist alles in Ordnung. Die Inbox war zuletzt leer, …
  21:32:49 user      | alle in ordnung ?
  21:32:54 assistant | Ja, bei mir läuft alles. Die Inbox ist gerade geprüft und leer. …
```

That rollout's user seat also held the `# AGENTS.md instructions` block, the hive brief on every resume (five times), and the inbox nudges; 17 `developer` records are Codex's skills/tool guidance. None of it came through — only the human's three prompts and the replies. The Claude reader was checked the same way against a live transcript: spawn brief and tool traffic stripped, both human messages present.

## How I tested it

- OS: Windows 11 (10.0.26200)
- Steps:
  1. `npm run typecheck` — both programs clean.
  2. `npm run test:focused` — 847 tests, 811 pass. The 28 failures are pre-existing on `main` (Windows symlink/EPERM suites); same set, same count, verified by running the suite on `main` for comparison.
  3. `npm run build` — succeeds.
  4. Ran the OpenCode reader's exact SQL against the live `opencode.db` of a running agent: session resolved, 18 turns returned, filters confirmed. Confirmed separately that `better-sqlite3` opens that live WAL database read-only under Electron's own runtime (`npx electron`), since the module is built for Electron's ABI and cannot be exercised from plain node.
  5. Ran the Claude reader against a live Claude Code transcript: 19 turns, tool traffic stripped, human messages present.
  6. Read a live Codex reviewer's rollout record by record to learn the real shape before writing the parser (`response_item/message` roles and content block types), then ran the finished reader against that same `CODEX_HOME` — output above.
  7. New tests in `test/chat-transcript.test.cjs` (16) cover the shared filters, the tool-stripping, a torn trailing line during a tail, the OpenCode/Codex seams the feature hangs from, Codex's own injected-prompt patterns, and rollout resolution (session id first, newest-write fallback second).

Not covered: the IPC-to-UI path is exercised only by typechecks, not by a click-through in the packaged app.

## Credit (optional)

Discord:

X:

## Checklist

      spacing, or fonts.
      `ATTRIBUTION.md`.

### DIFF
--- a/src/main/chatTranscript.ts
+++ b/src/main/chatTranscript.ts
@@ -1,0 +1,124 @@
+import { closeSync, existsSync, openSync, readSync, statSync } from 'node:fs';
+import { isInboxNudge } from '../shared/hiveNudge';
+import { isCompactionCommand } from '../shared/providerAutomation';
+
+/** One plain-text turn extracted from a Claude Code transcript — a user prompt
+ * or an assistant reply, with every tool_use/tool_result block stripped out.
+ * This is the "desktop-app view": just the conversation, none of the raw TUI
+ * noise (spinners, tool logs, box drawing) the terminal panel shows. */
+export interface ChatMessage {
+  role: 'user' | 'assistant';
+  text: string;
+  ts: number;
+}
+
+/** Bounds how many turns a single tab keeps in memory/render — a long-running
+ * agent's transcript can hold thousands; the chat tab only needs recent ones. */
+const MAX_MESSAGES = 300;
+
+interface TailEntry {
+  size: number;
+  mtimeMs: number;
+  /** Bytes parsed so far — always ends on a newline boundary, mirroring the
+   * usage tailer in transcript.ts so a torn trailing line is simply re-read
+   * once the writer completes it. */
+  offset: number;
+  messages: ChatMessage[];
+}
+
+const tailCache = new Map<string, TailEntry>();
+/** Soft bound; when crossed, the oldest-inserted half is dropped (entries
+ * rebuild on demand). Mirrors USAGE_CACHE_MAX in transcript.ts. */
+const TAIL_CACHE_MAX = 512;
+
+function extractText(content: unknown): string {
+  if (typeof content === 'string') return content.trim();
+  if (!Array.isArray(content)) return '';
+  const parts: string[] = [];
+  for (const block of content) {
+    if (block && typeof block === 'object' && (block as { type?: unknown }).type === 'text') {
+      const text = (block as { text?: unknown }).text;
+      if (typeof text === 'string' && text.trim()) parts.push(text.trim());
+    }
+  }
+  return parts.join('\n\n');
+}
+
+/** Parse complete JSONL lines into chat turns, appending to `out`. A record
+ * whose content is ONLY tool_use/tool_result blocks yields no text and is
+ * skipped — that is exactly the raw tool traffic the chat view exists to hide. */
+function parseChatLines(text: string, out: ChatMessage[]): void {
+  for (const line of text.split('\n')) {
+    const trimmed = line.trim();
+    if (!trimmed) continue;
+    let rec: { type?: unknown; message?: { content?: unknown }; timestamp?: unknown };
+    try {
+      rec = JSON.parse(trimmed);
+    } catch {
+      continue;
+    }
+    if (rec.type !== 'user' && rec.type !== 'assistant') continue;
+    const body = extractText(rec.message?.content);
+    if (!body) continue;
+    // The inbox-wake nudge and a /compact command are typed into the pty by the
+    // app itself (mail poll / context-cap trigger), not by the human — same text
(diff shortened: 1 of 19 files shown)


chaitanyagiri:
A chat view per agent is something people will use, and nothing like it exists today. It needs a fix round before it can go in:

* **Claude Code's own records show up as things the user typed.** `parseClaudeRecord` (`src/main/chatTranscript.ts:81`) keeps every `type: 'user'` record that `isAppGeneratedPrompt` does not catch. On real hive transcripts, 27% to 72% of the "you" bubbles were compaction summaries, slash command output and interrupt notices. Skipping records marked `isMeta` or `isCompactSummary`, and bodies wrapped in `<command-name>` or `<local-command-stdout>`, would cover most of it.
* **The first read blocks the main process.** On a cold cache, `tailChatJsonl` (`:117`) allocates and reads the whole file synchronously. A 276 MB transcript took 3.1 s and 664 MB of memory, and that repeats after every restart because the cache lives in memory. Reading only the tail of the file, or reading off the main thread, fixes it.
* **`transcript_path` is trusted as sent.** `src/main/hooks.ts:163` stores whatever path a hook payload carries, under whatever agent id it names, and the chat handlers read it (`src/main/index.ts:3866`, `:3898`). Please check that the path belongs to that agent before reading it.
* **Every poll resends the whole list.** The view polls every 1.5 s and gets the full message array each time, 300 to 400 KB on a long session. Sending only messages newer than the last one the view has would cut that.
* **OpenCode session choice.** It can pick up the user's own OpenCode session when one runs in the same folder, or the previous run's session after a restart.

PR #477, opened 2026-09-08
Fix/windows hook 83 lookup

perkays:
## What & why

On Windows, with a harness home whose path contains a space, hook commands do not run. The default home is `Documents\Munder Difflin`, so this is the out-of-the-box state, and it hits Claude, Codex, agy, pi, gemini and grok agents alike. Hooks are non-blocking, so nothing fails loudly: the agent keeps working and looks healthy, while live status, cost, the `Stop`-driven inbox drain and the transcript path the Chat tab reads all quietly stop arriving. The only place it shows is the agent's own TUI.

Three shells host hooks on Windows, and the quoted `"…\hive-node.cmd" "…\cth-hook.cjs"` form loses in every one of them, each for its own reason:

- **cmd.exe** (agy, pi, gemini) strips the outer pair of quotes off `cmd /c "A" "B"` and re-reads the line from the middle of the launcher's own path → `'C:\Users\…\Documents\Munder' is not recognized as an internal or external command`.
- **PowerShell** (Codex) parses a line that *starts* with a quoted string as an expression, not a command → `Unexpected token '"C:\…\cth-hook.cjs"' … At line:1 char:66`, char 66 being exactly where the second quoted path begins.
- **Git Bash** (Claude Code — on Windows too) runs a `.cmd` by handing the line to cmd.exe, so it inherits the cmd.exe failure above verbatim. And the obvious alternative, an unquoted 8.3 short path, fails differently there: bash treats a backslash outside quotes as an escape, so `C:\Users\…\HIVE-N~1.CMD` arrives as `C:UsersfardiDOCUME~1MUNDER~1hivebinHIVE-N~1.CMD: command not found`.

So there are two fixes, one per shell family.

For **cmd.exe and PowerShell** the answer is a command with no space and therefore no quotes: the 8.3 short path. `3cf1c999` reduces hook paths to it. `119eab6c` makes that lookup actually work — it asked cmd for the short name with the path quoted inside the spawn argument, and a `"` does not survive that trip: Node escapes it as `\"` (the MSVCRT convention) and cmd.exe, which has no backslash escape, reads the line as garbage. Since the function is only ever called for a path that HAS a space, it failed in 100% of its calls and silently returned the long path. The quotes now come from an environment variable, so nothing Node escapes ever reaches cmd.

For **Git Bash** the answer is a bash command. `68852a57` tried the short path for Claude and moved its failure from exit 1 to exit 127; `e2cdb51d` corrects that. Windows now writes the POSIX `hive-node` sh launcher alongside `hive-node.cmd`, and Claude's hook commands take that one, double-quoted with forward slashes. No cmd.exe hop, no dependence on 8.3, and the launcher invariant the existing hook tests pin still holds. Grok, which the intermediate commit had also moved to the short path without evidence of its hook shell, is back on the form it had.

Where 8.3 is disabled on the volume the cmd/PowerShell path still has to quote — but it now logs once instead of leaving a wall of `Hook failed` as the only evidence.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before

The generated commands, and what each host shell does with them:

```
Claude  settings.json : "C:\…\hive-node.cmd" "C:\…\cth-hook.cjs"
  Git Bash → cmd.exe  exit 1   'C:\Users\fardi\Documents\Munder' is not recognized as an
                               internal or external command

Codex   config.toml   : "\"C:\\…\\hive-node.cmd\" \"C:\\…\\cth-hook.cjs\""
  PowerShell          exit 1   At line:1 char:66
                               Unexpected token '"C:\…\cth-hook.cjs"' in expression or statement.
```

The 8.3 lookup that was meant to prevent both, measured as shipped:

```
AS-SHIPPED  -> "C:\\\"C:Users\fardiDocumentsMunder Difflinhive\binhive-node.cmd\\\""
accepted?   -> false
```

And the short path handed to Claude anyway, as Claude Code itself recorded it in the transcript (`hook_non_blocking_error`, 19:06:09Z):

```
/usr/bin/bash: line 1: C:UsersfardiDOCUME~1MUNDER~1hivebinHIVE-N~1.CMD: command not found
```

### After

Codex / cmd.exe hosts:

```
hook command : C:\Users\fardi\DOCUME~1\MUNDER~1\hive\bin\HIVE-N~1.CMD C:\Users\fardi\DOCUME~1\MUNDER~1\hive\bin\cth-hook.cjs
cmd.exe      exit=0
PowerShell   parse OK, exit=0
```

Claude / Git Bash host, same directory with a space, every shape side by side:

```
quoted hive-node.cmd, forward slashes    exit=1    '…\Munder' is not recognized
short 8.3, backslashes                   exit=127  command not found
short 8.3, forward slashes               exit=0
quoted POSIX hive-node, forward slashes  exit=0    env and arguments intact   ← what Claude now gets
```

## How I tested it

- OS: Windows 11 (10.0.26200); Windows PowerShell 5.1, cmd.exe, Git Bash 2.50; default harness home `C:\Users\<me>\Documents\Munder Difflin` (contains a space)
- Steps:
  1. Read the `settings.json` and `config.toml` the running app had just generated for a live Claude agent and a live codex agent. Confirmed the codex parser error's `char:66` matches the generated line byte for byte, and reproduced the Claude error verbatim by running its exact command through cmd.exe.
  2. Reproduced the 8.3 lookup as shipped from node — returned the garbage above, so the guard rejects it and the long path survives.
  3. After the short-path change, read Claude Code's OWN record of the result in the agent's transcript: `hook_non_blocking_error` with the `/usr/bin/bash: … command not found` stderr, and a `stop_hook_summary` naming the short-path command. That is what established Git Bash as Claude's hook host on Windows.
  4. Ran every candidate shape through `bash.exe -c` in a directory with a space, with no `AGENT_ID`/`HIVE_SOCK` so the shim no-ops rather than posting into a live hive — table above.
  5. Confirmed Git Bash executes a `#!/bin/sh` launcher off NTFS with no mode bits set (chmod is a no-op there), so the Windows-side POSIX launcher needs no chmod.
  6. Reverted the Claude call site by hand and re-ran the guard test to confirm it fails on regression, then restored it.
  7. **End to end, against the live app.** Started the app on this branch, let it write both launchers, spawned a Claude worker through a hive spawn request so the app generated its `settings.json` with the new command, then ran real Claude Code headless with exactly that hooks block against the app's live named pipe:
     ```
     claude -p "Reply with exactly: OK hooks" --settings <app-generated hooks> → exit 0, "OK hooks"
     transcript 399124f3-….jsonl: hook_non_blocking_error records: 0
     hive/log.jsonl: {"kind":"session","agentId":"worker-hookprobe-…","sessionId":"399124f3-…"}
     ```
     The last line is the app's `recordSession`, which only runs on a received hook payload, carrying the same session id Claude named its transcript after — so bash → POSIX launcher → shim → pipe → HookServer round-tripped, which is the link that had been dead.
  8. `npm run typecheck` clean, `npm run build` succeeds.
  9. `npm run test:focused` — 844 tests, 808 pass. Baseline on `main` is 834/798/28; this branch adds 10 tests and 10 passes with the same 28 pre-existing Windows symlink/EPERM failures.

New tests in `test/windows-hook-shortpath.test.cjs` (8). Five are behavioural, run
against the real cmd.exe, PowerShell and bash.exe, because every one of these bugs
lived at a shell boundary that reading the source did not reveal. Two of them
assert the OLD shapes still fail under their shells — so if a shell ever changes
its quoting rules, simplifying this becomes a deliberate decision rather than an
accident. `test/hive-hook-node.test.cjs` learned that on Windows "the launcher"
is either of the two.

Not covered: a volume with 8.3 disabled. The cmd/PowerShell path still quotes
there, and is now reported rather than silent, but I had no such volume to test on.

## Credit (optional)

Discord:

X:

## Checklist

      spacing, or fonts.
      `ATTRIBUTION.md`.

### DIFF
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -1204,7 +1204,19 @@ export class HiveManager {
   private hookSettings(shim: string, cwd: string, cfg: McpDefaultsMap, theme?: 'light' | 'dark', writableDirs: string[] = []): unknown {
     // Bundled node, NOT bare `node` — see nodeLauncherPath(). Claude runs each of
     // these through `sh -c` with a stripped PATH, where `node` is often absent.
-    const cmd = this.nodeRun(shim);
+    //
+    // On Windows that shell is cmd.exe, and the quoted form does NOT survive it:
+    // `cmd /c "A" "B"` strips the outer pair of quotes off the whole line and then
+    // re-reads it from the middle of the launcher's own path, so every hook of
+    // every Claude agent whose hive home contains a space — the default one does —
+    // died with
+    //   'C:\Users\…\Documents\Munder' is not recognized as an internal or external command
+    // Hooks are non-blocking, so nothing failed loudly: the agent kept working
+    // while live status, cost, the Stop-driven inbox drain, and the transcript
+    // path behind the Chat tab all quietly stopped arriving. Take the same short
+    // path codex/agy/pi/gemini already take — no space, so no quotes, so nothing
+    // for a shell to re-split.
+    const cmd = process.platform === 'win32' ? this.nodeRunUnquoted(shim) : this.nodeRun(shim);
     const entry = (matcher?: string) => ({
       ...(matcher ? { matcher } : {}),
       hooks: [{ type: 'command', command: cmd }]
@@ -2448,8 +2460,14 @@ export class HiveManager {
         ...(matcher ? { matcher } : {}),
         // Let Grok apply its event-aware defaults (5s normally, 600s for Stop).
         // Grok is a HOOK bridge (not a proxy sidecar), so it is hit by the same
-        // `node: command not found` 127 — bundled node here too.
-        hooks: [{ type: 'command', command: this.nodeRun(shim) }]
+        // `node: command not found` 127 — bundled node here too. And by the same
+        // Windows space problem as Claude above: nodeRunUnquoted's own comment
+        // already lists grok among the providers it covers, but this call site
+        // never took it.
+        hooks: [{
+          type: 'command',
+          command: process.platform === 'win32' ? this.nodeRunUnquoted(shim) : this.nodeRun(shim)
+        }]
       });
       const hooks = {
         PreToolUse: [tool('.*')],
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -26,6 +26,7 @@ import {
 import { join, dirname, basename, isAbsolute, relative } from 'node:path';
 import { homedir } from 'node:os';
 import { spawnSync, spawn, type ChildProcess } from 'node:child_process';
+import { joinCommandLine } from '../shared/commandLine';
 import { randomBytes, createHash } from 'node:crypto';
 import type { AgentUsageSample } from './usage';
 import { COMMAND_GROUPS } from '../shared/claudeCommands';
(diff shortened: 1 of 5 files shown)


chaitanyagiri:
Thanks. Hooks going silent on Windows when the path has a space is a real bug, and the POSIX launcher for Claude plus the 8.3 lookup is the right approach. Three things before we can take it:

* **Volumes with 8.3 names turned off.** When no short name comes back (`src/main/hive.ts:658`), the path stays long and gets quoted, and your own log line says PowerShell hosted hooks (Codex) refuse that. So on those volumes Codex hooks are still dead, and the only trace is a console line. Please either find a form PowerShell accepts there or tell the user somewhere they will actually see it.
* **Profile paths with non ASCII characters.** The `spawnSync('cmd', ...)` call at `:651` decodes stdout as `utf8`, while cmd writes in the console code page by default. A profile like `C:\Users\José` likely comes back garbled, fails `existsSync`, and falls back to the long path. `cmd /d /u /c` with `encoding: 'utf16le'` avoids that. I have not run this on a Windows machine, so please check it.
* **The call runs on the main process with no `windowsHide` and no `timeout`.** A slow or stuck cmd blocks the app, and a console window can flash. Please add both.

The 4 real shell tests skip on macOS, so they have not run anywhere yet. A run on Windows would help.

PR #479, opened 2026-09-09
Stop the hive router's synchronous git commit from blocking the main thread

mblumkin:
## Summary

The hive message router runs `git add -A` + `git commit` synchronously (`spawnSync`) on every routed message, in `HiveManager.commit()` / `.git()` (`src/main/hive.ts`). With `gc.autoDetach=false` that also runs `gc --auto` inline. This blocks the Electron main thread for the whole duration of the commit — measured live against a production hive repo at several seconds per routed message, freezing the UI.

`git()` now uses `execFile` instead of `spawnSync`, so the git child process no longer blocks the main thread. `commit()` is serialized through an explicit promise queue (`this._commitQueue`) instead of relying on `spawnSync`'s incidental single-threaded serialization — with `gc.autoDetach=false`, two real git processes racing the same `.git/objects/` is exactly the hazard that flag exists to prevent (see the comment above `git()`), so losing that free serialization without replacing it would have been worse than the freeze it fixes.

Cadence and what gets committed are unchanged — still `add -A`, still one commit per routed batch. This is off-thread, not off-cadence.

A `flushCommits()` method is added so a caller that needs the working tree to be quiescent (most concretely: a test deleting the ephemeral hive home it created) can wait for the queue to drain. Every existing `HiveManager` test's teardown now awaits it before removing its temp dir — with `spawnSync`, a commit always finished before the function that queued it returned, so no test needed to ask for that guarantee explicitly; async `git()` removes that free ordering. One instance of the same bug was in production code, not just tests: `ensureHive()`'s own `git init` call was fire-and-forget and not on the queue, so a caller could observe `ensureHive()` return before `git init` had actually finished on disk — it's queued now too.

Not changed in this PR (flagged as a separate, larger follow-up): the router's `git add -A` could instead pass only the paths each of the 9 `commit()` call sites actually touched. That's independent of the threading fix and touches all 9 sites individually, so I've kept it out of this diff.

### Before

Measured with a standalone Node script (not the full app — the freeze isn't easily screen-recordable) against a real-sized clone of a production hive repo (896 MB, cloned locally so history/tree size matches). It calls `HiveManager.commit()` once (with one new file staged, matching a router commit's shape) and times how long the *synchronous call itself* takes to return — that duration is exactly how long it blocks the calling thread.

```
$ node bench-commit-block.cjs <home> <repo-root-on-main> block
call-returns-in: 1103.6ms
actual-commit-completes-in: (same as above -- synchronous)
```
Two more trials (fresh clones each time): 277.9ms, 263.6ms. All in the hundreds-of-ms-to-second range that matches the reported freeze in kind (the live production repo is larger and dirtier than this clone, which is consistent with the live report being ~6s).

### After

Same script, same-sized fresh clone, against this branch:

```
$ node bench-commit-block.cjs <home> <repo-root-on-this-branch> flush
call-returns-in: 0.0ms
actual-commit-completes-in: 182.8ms
```
Two more trials: 0.0ms / 304.8ms, 0.0ms / 151.4ms. The call returns to the caller immediately every time; the real git work (confirmed still happening, via `flushCommits()`) completes shortly after, off the main thread.

## Rebuild / verify note

This branch only changes `src/main/hive.ts` (+ test files). To verify against a running build:
1. `npm install && npm run build && npm run dist:mac` (or your platform's `dist:*`) from this branch.
2. Install/launch the built app pointed at a real (or copied) hive repo.
3. Route a message and watch for UI responsiveness during the commit — it should no longer visibly freeze. `console.warn('[hive] commit gave up after 5 attempts')` would appear in the main process log if a queued commit ever failed outright (it didn't in any of my local test runs).

No change to what gets committed, the commit message format, or commit cadence — only where the git subprocess's blocking time lands.

## Test plan
- [x] `npm run typecheck` (node + web) — clean
- [x] `npm run test:focused` — 831/834, matching `main`'s 831/834 exactly (the 3 remaining failures are pre-existing/unrelated: SIGILL/exit-signal-forwarding tests, reproduced identically on `main` before this change)
- [x] `npm run build` — clean
- [x] Standalone before/after timing harness against a real-sized (896M) hive repo clone, 3 trials each side (see above)

### DIFF
--- a/src/main/hive.ts
+++ b/src/main/hive.ts
@@ -25,7 +25,7 @@ import {
 } from 'node:fs';
 import { join, dirname, basename, isAbsolute, relative } from 'node:path';
 import { homedir } from 'node:os';
-import { spawnSync, spawn, type ChildProcess } from 'node:child_process';
+import { spawnSync, spawn, execFile, type ChildProcess } from 'node:child_process';
 import { randomBytes, createHash } from 'node:crypto';
 import type { AgentUsageSample } from './usage';
 import { COMMAND_GROUPS } from '../shared/claudeCommands';
@@ -191,9 +191,8 @@ export interface SpawnInjection {
 
 const HOP_CAP = 12;
 
-function sleepSync(ms: number): void {
-  const sab = new SharedArrayBuffer(4);
-  Atomics.wait(new Int32Array(sab), 0, 0, ms);
+function sleep(ms: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, ms));
 }
 
 /** Filesystem- and sort-safe timestamp, e.g. 2026-05-30T14-03-11-123Z. */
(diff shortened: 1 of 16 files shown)


mblumkin:
Closing per repo maintainer's own workflow preference -- this stays a local patch for now, re-applied after the next upstream release if not already fixed by then. Thanks for the fork access.

PR #512, opened 2026-09-14
fix(hive): mail to an archived agent is filed, and its sender is told nobody is there

gpechieu:
## What & why

An archived agent (its terminal is gone) keeps its inbox, so a message to it lands there, is logged as delivered, and stays unread. Filing it is right — a worker re-hired under the same id reads it on its first turn (#472 settles a released worker's inbox only up to its done signal) — but the sender learned nothing: god kept mailing dead agents for hours (seen live 2026-08-16), and a request to a worker nobody re-hired was simply lost.

Now a direct message to an archived recipient is still filed (and logged `archived-recipient`), and when it expects an answer (`request` / `query` / `propose`) and the sender is a live agent, the sender gets an immediate `inform` from `system`, threaded on the original (`in_reply_to`, `conversation`, `hops + 1`, `requires_reply: false` so it can never ping-pong): `[no one is there to answer — "<id>" is archived] <subject>`, with what to do about it (route to a live agent or hire one). Inform-only mail is filed quietly; broadcasts already skip archived agents; god (the bounce target) is never treated as archived.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before

before

On main the new test is red: the request is filed and reads as delivered, and the sender's inbox stays empty.

### After

after

Same test with the fix: filed verbatim, the sender holds the notice, the ledger has the `archived-recipient` entry and the message log still reports the delivery that happened.

## How I tested it

- OS: macOS 15 (Darwin 25.6), Node 22, branch cut from `main` @ 417d8dec
- Steps:
  - `node --test test/hive-archived-recipient.test.cjs` → 3 of 4 fail on `main` (src stashed), 4/4 pass with the fix.
  - `npm run typecheck` clean; `npm run test:focused` 838/838.
  - The four cases: request → filed + notice (act, from `system`, `in_reply_to`, `requires_reply` false, subject prefix, body) + log + delivered list; query/propose notify while inform/done are filed quietly; a live non-god sender is notified and a sender off the roster (`system`) is not; an unarchived agent gets ordinary mail again.

## Credit (optional)

Discord:

X:

## Checklist

- [x] Any new UI derives from `DESIGN.md` / `tokens.ts` — no ad-hoc colors, spacing, or fonts. (no UI)
- [x] If I added art, it's my own or compatibly licensed, and listed in `ATTRIBUTION.md`. (no art)


gpechieu:
**Agent review of this PR** (CONTRIBUTING: "run your coding agent over your own PR, and paste what it found"). Claude Code went over `417d8dec..ce221cea` for cross-platform behaviour, paths with spaces, provider neutrality, hot paths and security, then hygiene.

- **Cross-platform / paths with spaces:** clean — only `join()` / `existsSync` / `atomicWriteJson`; filenames are harness ids; no shell.
- **Provider neutrality — two things it flagged, both fixed in be60a476:**
  - The archived branch sat *after* the hookless and proxy-tier branches, so an archived target on those engines never reached it: the mail went to the terminal handoff (there is no terminal), nothing was filed, no log, no notice. The archived check now runs right after the assistant bounce — with no terminal the inbox is where the mail goes, whatever the engine.
  - The notice was `deliver()`ed straight into the sender's inbox, which a hookless / proxy-tier sender never drains. It now goes through `routeMessage`, so such a sender receives it the way it receives any mail (terminal handoff, or the god bounce when no renderer is there). `requires_reply:false` and the live-sender guard keep it from re-entering; hops are +1.
- **Hot paths:** clean — reuses `reg`, one O(1) lookup, extra I/O only on the archived path.
- **Security / loop risk:** clean — the notice cannot ping-pong (`requires_reply:false`, archived senders are never notified, a reply to `system` hits the existing no-inbox drop + bounce once); the subject echoes the sender's own text; the target id is a registry key.
- **Hygiene — fixed:** the redundant `msg.from !== t` guard (targets already exclude the sender).
- **Flagged and deliberately not changed:** if `deliver()` fails for an archived id the `continue` skips the base no-inbox bounce — theoretical, no flow deletes an agent dir while keeping its registry entry; harness senders that are not registry agents (`scheduler`, `breaker`, the literal `'god'` used by the worker dispatch) get no notice — they cannot be mailing an archived agent in practice (the dispatch targets a worker that was just spawned).

Verified after the fix: typecheck clean, the test file 5/5 (new case: an archived custom-engine agent is filed and its sender told; a custom-engine sender is told through routing), local suite 839/839.

gpechieu:

PR #513, opened 2026-09-14
fix(main): under npm run dev, change-home and reset exit cleanly instead of relaunching into a blank window

gpechieu:
## What & why

`changeHome` (Settings → change harness home, also the onboarding "create a hive" path) and `resetAll` both end in `app.relaunch()` + `app.exit(0)`. Packaged, that is the clean re-bind. Under `npm run dev` the renderer is served by electron-vite's dev server, which lives in the wrapper process that watches this Electron and exits with it — so the relaunched Electron started against a dead dev server and came up with a **blank window** that the developer had to hunt down and kill before running `npm run dev` again. It happens on every hive create/switch in dev.

`relaunch.ts` decides the plan: `'exit'` when `ELECTRON_RENDERER_URL` is set (electron-vite's dev marker, the one the window loader already keys on), `'relaunch'` otherwise. `index.ts` has one `relaunchOrExit(reason)` that both handlers go through; in dev it prints a console line saying why and what to run, and exits cleanly. Packaged behaviour is unchanged.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Build / CI

### Before

before

On main both handlers call `app.relaunch()` unconditionally; the sequence that ends in the blank window is spelled out in the capture, and the new test has nothing to load.

### After

after

With the fix: the plan is `exit` under `ELECTRON_RENDERER_URL`, the console line names the trigger and says to run `npm run dev` again, and the text pin confirms both handlers use the helper and `app.relaunch()` appears only inside it.

## How I tested it

- OS: macOS 15 (Darwin 25.6), Node 22, branch cut from `main` @ 417d8dec
- Steps:
  - `node --test test/dev-relaunch-exit.test.cjs` → fails on `main` (no `relaunch.ts`, no helper), 3/3 pass with the fix.
  - `ELECTRON_RENDERER_URL=http://localhost:5173 node -e '…relaunchPlan()…'` prints `exit` and the notice (in the After capture).
  - `npm run typecheck` clean; `npm run test:focused` 837/837.

## Credit (optional)

Discord:

X:

## Checklist

- [x] Any new UI derives from `DESIGN.md` / `tokens.ts` — no ad-hoc colors, spacing, or fonts. (no UI)
- [x] If I added art, it's my own or compatibly licensed, and listed in `ATTRIBUTION.md`. (no art)


gpechieu:
**Agent review of this PR** (CONTRIBUTING: "run your coding agent over your own PR, and paste what it found"). Claude Code went over `417d8dec..9307e75f` for cross-platform behaviour, paths with spaces, provider neutrality, hot paths and security, then hygiene.

- **Cross-platform:** clean — it verified in the installed electron-vite 5 that the wrapper sets `ELECTRON_RENDERER_URL` before spawning Electron with an inherited env (all OSes) and exits on the child's close, so nothing of it remains on Windows; the window loader already keys on the same variable, so dev would be broken anyway if it were unset. Packaged: the variable is undefined, so the plan is `relaunch`, identical to before this PR. **One thing it flagged, fixed in 90bf29f3:** `console.warn` followed by an immediate `app.exit(0)` — stderr is asynchronous on pipes and on Windows TTYs, so the notice could be lost. It is now written with a callback that exits once it is out.
- **Paths with spaces / provider neutrality / hot paths:** n/a — no paths handled (the URL is host:port), no provider logic, runs once just before exit.
- **Security:** clean — env read only; no renderer-controlled input reaches `relaunchPlan`; no new IPC or fs.
- **Hygiene:** clean — no leftover debug output, no commented code, no reformatting of untouched lines; the renderer callers (`HivePicker`, `SettingsModal`) already treat `ok` as "never resolves" and die with the process; the one-shot `skipHivePickerOnce` flag is consumed by the next `npm run dev`, which is the intended landing.
- **Flagged and deliberately not changed:** the renderer comments still say "relaunches" (true when packaged); `changeHome` / `reset` skipping `telemetry.stop` / `stopAllProxyBridges` etc. vs `teardownAndQuit` is pre-existing on both branches and out of this PR's scope.

Verified after the fix: typecheck clean, the test file 3/3, local suite 837/837.

gpechieu:

PR #584, opened 2026-09-21
fix(hive): repair an agent cwd in app

### LINKED ISSUE #445: Renaming an agent's project folder permanently breaks that agent, with no in-app way to fix it
**Version:** Munder Difflin v0.4.6 (packaged) · Windows 11 Home 26200

## What happened

I renamed a worker's project folder on disk (`Documents\員工新增測試` → `Documents\agent_test`,
to get rid of the non-ASCII characters). The app was closed at the time. On the next launch the
agent was effectively dead: it never started, and there is no UI affordance anywhere to point it
at the new path.

The agent is not deleted — its `registry.json` entry is intact, `archived: false`, memory and
`sessionId` all still there. The only thing that changed is `cwdValid: false`.

## Why it happens

`cwdValidity(cwd)` validates that the cwd is absolute **and currently exists as a directory**,
and the result is persisted onto the registry entry:

> Validate an agent's cwd the way a spawn does — it must be an ABSOLUTE path that exists as a
> directory. Surfaced as `cwdValid` on the registry entry so the roster reliably exposes whether
> a worker's working directory is usable.

That is correct behaviour on its own. The problem is that **there is no way to correct the path
afterwards.** The IPC surface exposes `hive:renameAgent(id, name)` — which renames the *display
name* only. There is no `setAgentCwd` / "change folder" / "relocate" handler. `setCwd` exists
only as React state inside the **Add Agent** dialog, so a folder can be chosen when an agent is
created and never again.

So a rename of a folder — an ordinary thing to do — puts an agent into a state that the app can
detect and report but not repair.

## Side effects observed

The failed startups armed the circuit breaker. The worker went `healthy → steering → constrained`
and accumulated `Circuit breaker: steer` / `Circuit breaker: constrain` messages in its inbox,
which is misleading: the agent was not looping or overspending, it simply could not launch.

## Current workaround (hand-editing app state)

Both files must be edited — changing only one has no effect:

1. `<harnessHome>\hive\registry.json` → `agents["<id>"].cwd`
2. `%APPDATA%\munder-difflin\config.json` → the matching entry in `registeredRepos`

After that `cwdValid` flips back to `true` and `fleet.json` picks the new path up without an app
restart. It works, but it means editing internal state by hand, and a user who does not know
`cwdValid` exists has no way to find out why their agent vanished.

## Suggested fixes (roughly in order of value)

1. **A "change folder" action on the agent card**, backed by a `hive:setAgentCwd(id, path)` IPC —
   the same folder picker the Add Agent dialog already has. This alone solves it.
2. **Surface the reason in the UI.** When `cwdValid === false`, show the agent as *needs
   attention* with the message the orchestrator already receives ("Working directory: X, which is
   not a valid directory — spawning there would fail") rather than letting it silently fail to
   start.
3. **Do not arm the circuit breaker on spawn failures** caused by an invalid cwd — that is a
   configuration problem, not runaway behaviour, and the breaker messages point the user at the
   wrong diagnosis.
4. Optional: on startup, if a registered repo path is missing but a sibling directory matches by
   inode/creation time, offer it as a suggested relocation.

## Note on non-ASCII paths (possibly related, possibly separate)

The reason for the rename in the first place: Claude Code encodes the cwd into the session
directory name under `~/.claude/projects/`, replacing each non-ASCII character with a single `-`.
Two different Chinese folder names of the same length under the same parent therefore collide
onto one session/memory directory (`C--Users-sjack-Documents-------`). Recommending ASCII project
folder names in the docs — or warning at Add Agent time — would help users avoid this.

## Unrelated minor issue found while investigating

A malformed message written into `agents/<id>/outbox/` is renamed to `bad-<filename>` and moved
into `outbox/.sent/`, then silently dropped. The sender gets no bounce and no error, so an agent
waiting on a reply waits forever with no way to tell that its message never left. A bounce back
into the sender’s own inbox (the way an unknown recipient is already handled) would make this
self-diagnosing.

TTAWDTT:
## What & why

A project folder renamed on disk used to leave an agent permanently unusable: the registry detected `cwdValid: false`, but the app had no way to repair the path. This adds a localized "needs attention" state plus a folder picker-backed `hive:setAgentCwd` repair action that updates the durable registry, fleet snapshot, and matching registered repo while preserving agent identity, memory, and session.

Closes #445

## Type of change

- [x] Bug fix

### Before

Issue #445 before: a renamed folder leaves cwdValid false with no in-app repair control

### After

Issue #445 after: agent card and detail panel expose a change-folder repair action

## How I tested it

- OS: Windows 11
- Steps: `npm run typecheck`; `node --test test/agent-cwd-repair.test.cjs`; `npm run build`.

Discord: ttawdtt

## Checklist

      spacing, or fonts.
      `ATTRIBUTION.md`.


TTAWDTT:
Superseded by the consolidated final PR for this issue (#586).

PR #583, opened 2026-09-21
feat(office): show persistent avatar nametags

TTAWDTT:
## What & why

Every floor avatar now carries a persistent nametag showing its display name. The tag is a child of the avatar sprite, so it moves, fades, and world-scales with it; the store sync updates a live rename without recreating the avatar.

Closes #340

## Type of change

- [x] New feature

### Before

Issue #340 before: the office floor shows two side-by-side workers with action bubbles but no identity

### After

**Update:** nametag dimensions reduced after review feedback — ontSize 8→5, TAG_HEIGHT 12→8, TAG_TEXT_Y 8→5 (closer to feet), background padding 8→4.

Issue #340 after: persistent nametags below each avatar, including Adaline, Bernard, Clara and Michael

## How I tested it

- OS: Windows 11
- Steps: `npm run typecheck`; `node --test test/office-avatar-nametag.test.cjs`; reviewed sprite parenting and the store sync path for add/rename/hide behaviour.

Discord: ttawdtt

## Checklist

- [x] **Before and after evidence is attached above**, under both headings.
- [x] `npm run typecheck` passes.
- [x] Focused tests pass.
- [x] `npm run build` succeeds.
- [x] This PR is **one change**. Unrelated fixes belong in their own PR.
- [x] I read the diff myself before opening this, and there is no debug output,
      commented-out code, or unrelated formatting churn in it.
- [x] Any new UI derives from `DESIGN.md` / `tokens.ts` — no ad-hoc colors,
      spacing, or fonts.
- [x] If I added art, it's my own or compatibly licensed, and listed in
      `ATTRIBUTION.md`.

PR #588, opened 2026-09-21
feat(office): show persistent avatar nametags

### LINKED ISSUE #340: Avatars carry a live action bubble but no name — telling agents apart on the floor means clicking each one

TTAWDTT:
## What & why

Every floor avatar now carries a persistent nametag with its display name. The tag is a child of the avatar sprite, so it moves, fades, and world-scales with the character, and the store sync updates a live rename without recreating the avatar.

Closes #340

## Type of change

- [x] New feature

### Before

Issue #340 before: two side-by-side workers show action bubbles but no identity

### After

Issue #340 after: persistent nametags below each avatar

## How I tested it

- OS: Windows 11
- Steps: `npm run typecheck`; `node --test test/office-avatar-nametag.test.cjs`; `npm run build`.

Discord: ttawdtt

## Checklist

      spacing, or fonts.
      `ATTRIBUTION.md`.


TTAWDTT:
Superseded by #583.

